PR #8, opened 2025-11-14
Update benchmark document

HuXin0817:
# Update benchmark document

* Upload reproducible documents and sports model configuration (same as experimental environment)
* Modify the dependency configuration to align with the pyproject.toml version requirements. 
* Also change absolute equal to greater than or equal to
Modify the concurrency parameter to 10 to avoid model platform current limitation.

TODO: 
* Verify whether the latest dependency version is compatible through ci
* Write model multi-ak proxy platform to deal with current limiting issues


benchmark/README.md
```diff
@@ -0,0 +1,418 @@
+# PowerMem Benchmark
```
wayyoungboy: move: benchmark/README.md  to docs/benchmark/overview.md

PR #11, opened 2025-11-14
Remove 'no fact falling back simple add' action 

Evenss:
When adding a new memory, if the LLM does not extract the facts, there's no need to fall back to performing a simple add operation.


Teingi (approved):
LGTM

PR #38, opened 2025-11-26
feat(config): expand provider validation using factory classes

Evenss:
# feat(config): expand provider validation using factory classes

EmbedderConfig, LLMConfig, and VectorStoreConfig now validate providers against both hardcoded lists and dynamically registered providers in their respective Factory classes. This allows for easier extension and support of custom providers without modifying the config files directly.


Teingi (approved):
LGTM

PR #79, opened 2025-12-05
[enhancement]:dbconfig optimize

### LINKED ISSUE #70: [Enhancement]: Easier to switch config in .env
### Description

Maybe it is complicated to switch config in .env by uncommenting which may sometimes be confusing.Is there some way easier to switch it？I also noticed there are some redundant configurations which have much space to optimize.

longbingljw:
## Summary

It is a draft to provide a way to switch config easier.
close #70


Teingi (approved):
LGTM

PR #101, opened 2025-12-11
fix(env): update GRAPH_STORE_HOST from localhost to 127.0.0.1

HuXin0817:
update GRAPH_STORE_HOST from localhost to 127.0.0.1 in .env.example


Teingi (approved):
LGTM

PR #110, opened 2025-12-11
fix graph store time zone

### LINKED ISSUE #109: [Enhancement]: Time zone management in graph databases should be consistent with that in vector databases.
### Description

Time zone management in graph databases should be consistent with that in vector databases.

Evenss:
Modify the time zone management in the graph database to align it with that of the vector database.

Close #109


Teingi (approved):
LGTM

PR #114, opened 2025-12-15
refactor: llm and embedding config simplify

### LINKED ISSUE #70: [Enhancement]: Easier to switch config in .env

### LINKED ISSUE #99: [Enhancement]: Add ollama and huggingface option
### Description

I notice that there is no option to select ollama,huggingface and other local models in .env although there are these logics.Have these logics been completely tested?There is only apikey option in .env such as openai and qwen.

longbingljw:
## Summary

1.llm and embedding config simplify
2.add option to huggingface and ollama 
close #70
close #99

### DIFF
--- a/docs/guides/0004-custom_prompts_usage.md
+++ b/docs/guides/0004-custom_prompts_usage.md
@@ -389,7 +389,7 @@ Only return relationships that need to be deleted.
 """,
         
         config={
-            "host": "localhost",
+            "host": "127.0.0.1",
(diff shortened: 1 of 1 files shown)


Teingi (approved):
LGTM

PR #115, opened 2025-12-15
Bump version to 0.2.0

Teingi:
- Update version number from 0.1.0 to 0.2.0 across all files
- Add bump-version make target for automated version updates
- Update version in pyproject.toml, version.py, telemetry, audit, and examples

### DIFF
--- a/src/powermem/version.py
+++ b/src/powermem/version.py
@@ -2,7 +2,7 @@
 Version information management
 """
 
-__version__ = "0.2.1"
+__version__ = "0.2.0"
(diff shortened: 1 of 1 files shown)


src/powermem/version.py
```diff
@@ -2,11 +2,12 @@
 Version information management
 """
 
-__version__ = "0.1.0"
+__version__ = "0.2.1"
```
Evenss: version string is not correct?

Evenss (approved):

PR #128, opened 2025-12-17
fix(config):fix embedding_model_dims string bug

Evenss:
## Summary
Fix the BUG in OceanBase database where passing embedding_model_dims as a string causes a subsequent creation to result in the string "1536" not being equal to the number 1536


Teingi (approved):
LGTM

PR #173, opened 2026-01-13
refactor(server/config): validate with the native pydantic-settings approach

PsiACE:
## Summary

as title.

## Solution Description

For env prefix and .env file parsing, we can directly rely on pydantic-settings. For specific requirements, such as logic that requires converting multiple values to booleans, corresponding function `_parse_boolish` have been implemented.

I have verified that some basic functions, such as auth and rate limit are working properly.


Teingi (approved):
LGTM

PR #178, opened 2026-01-14
refactor(memory): make config loader work with pydantic-settings

PsiACE:
## Summary

config loading now uses pydantic‑settings and PowermemSettings preserves existing env prefixes and to_config output.

## Solution Description

As part of our entry point governance, we are transitioning from manual environment variable parsing to structured settings models.


Teingi (approved):
LGTM

PR #190, opened 2026-01-20
Regression case modify

Ripcord55:
## Summary
Regression case modify

## Solution Description
API Server case collection logic modification; User_defile case logic fix.


Teingi (approved):
LGTM

PR #229, opened 2026-02-04
Fixed run failure caused by incorrect folder name

Evenss:
## Summary
Fixed run failure caused by incorrect folder name


Teingi (approved):
LGTM

PR #253, opened 2026-02-11
fix search bug

Evenss:
## Summary
- Fixed an issue where user isolation was not working: user2 incorrectly referenced user1's memory_id for an update, and the request was rejected due to a mismatch in user_id.
- Fixed a batch delete SQL error: a database layer exception was triggered when the delete_all method was executed.

---
## Summary
- Fixed an issue where user isolation was not working: user2 incorrectly referenced user1's memory_id for an update, and the request was rejected due to a mismatch in user_id.
- Fixed a batch delete SQL error: a database layer exception was triggered when the delete_all method was executed.


Teingi (approved):
LGTM

Issue #256, opened 2026-02-11
[Bug]: PGVectorConfig failed to load variable parameters


Ripcord55:
### Describe the bug

PGVectorConfig failed to load variable parameters
<img width="3420" height="804" alt="Image" src="https://github.com/user-attachments/assets/fc64772f-b559-44c1-84a0-a9e9a35c023c" />

### Environment

0.5.0_dev

### Fast reproduce steps

Using PG as the database, add memory

Ripcord55:
Verified

<img width="2262" height="832" alt="Image" src="https://github.com/user-attachments/assets/58cd5e5f-7494-43e0-a08e-0a790895b114" />

Issue #255, opened 2026-02-11
When adding memory, the parameter 'user_id' was not included[Bug]:


Ripcord55:
### Describe the bug

When adding memory, the parameter 'user_id' was not included, It causes the failure to add memory when the user_id is different but the content is similar
<img width="2056" height="1198" alt="Image" src="https://github.com/user-attachments/assets/599dfd04-365b-45f5-a040-c9baa929bc4d" />
The data in the actual database has not been updated:
<img width="2052" height="248" alt="Image" src="https://github.com/user-attachments/assets/ff62696f-8d57-4e3a-9db9-7ba72806188b" />

### Environment

0.5.0 dev

### Fast reproduce steps

Use different user_ids to add similar memory content

PR #261, opened 2026-02-21
fix: user_id isolation in _intelligent_add and PGVectorConfig defaults (#255, #256)

stakeswky:
## Summary

Fixes two related issues:

### 1. user_id isolation in _intelligent_add (#255)

**Problem:** When `_intelligent_add` searches for similar memories to deduplicate, it doesn't include `user_id`/`agent_id` in the search filters. This means User A's memory could be considered a duplicate of User B's memory if the content is similar enough, causing silent data loss.

**Fix:** Explicitly set `user_id` and `agent_id` in `search_filters` before calling `search_memories`, ensuring deduplication only happens within the same user's scope.

### 2. PGVectorConfig defaults (#256)

**Problem:** `PGVectorConfig` has `host`, `port`, and `user` defaulting to `None`, but the validator requires both `host+port` and `user+password` to be provided together. This means a minimal config like `{"password": "secret"}` fails validation even though sensible defaults (localhost:5432, postgres) exist.

**Fix:**
- Set defaults: `host="127.0.0.1"`, `port=5432`, `user="postgres"`
- Relax validator to only enforce pairing when non-default values are explicitly provided

## Testing

- Verified that `_intelligent_add` now correctly scopes similarity search per user
- Verified that `PGVectorConfig(password="secret")` works with defaults
- Verified that explicit non-default values still require their pair

### DIFF
--- a/src/powermem/core/memory.py
+++ b/src/powermem/core/memory.py
@@ -831,6 +831,13 @@ class Memory(MemoryBase):
                 }
                 search_filters.update(simple_metadata)
 
+            # Ensure user_id/agent_id are always in search_filters for correct isolation
+            # This prevents cross-user memory deduplication when content is similar
+            if user_id is not None:
+                search_filters['user_id'] = user_id
+            if agent_id is not None:
+                search_filters['agent_id'] = agent_id
+
             # Search for similar memories with reduced limit to reduce noise
             # Pass fact text to enable hybrid search for better results
             similar = self.storage.search_memories(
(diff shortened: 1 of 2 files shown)


src/powermem/core/memory.py
```diff
@@ -831,6 +831,13 @@ def _intelligent_add(
                 }
                 search_filters.update(simple_metadata)
 
+            # Ensure user_id/agent_id are always in search_filters for correct isolation
+            # This prevents cross-user memory deduplication when content is similar
+            if user_id is not None:
+                search_filters['user_id'] = user_id
+            if agent_id is not None:
+                search_filters['agent_id'] = agent_id
+
```
Evenss: This operation is already included in the `search_memories` section below, so there's no need to add it here.

PR #271, opened 2026-02-26
fix(oceanbase): pass where_clause in pyobvector-compatible shape

Teingi:
Ensure filter clauses are provided as a list of SQLAlchemy expressions for pyobvector search APIs, and fix FTS/LIKE fallback query construction to avoid SQLAlchemy type errors.


Evenss (approved):
LGTM

PR #276, opened 2026-03-01
regression modify

Ripcord55:
## Summary
Regression logic modify.

## Solution Description
Increase performance of seekdb, split multitasking execution regression, and add report overview function.


Teingi (approved):
LGTM

PR #305, opened 2026-03-09
refactor(cli): nest memory commands under memory

Teingi:
refactor(cli): nest memory commands under "memory", keep only shell, fix completion

- Move memory operations under "pmem memory": add, search, get, update,
  delete, list, delete-all (was pmem <cmd> at root)
- Remove redundant "interactive" command; keep "pmem shell" as the single
  REPL entry point
- Fix shell completion: use default for COMP_CWORD/CURRENT so [ -eq ] never
  gets an empty operand (fixes "unknown condition: -eq")
- Update completion command list and in-command examples to match


Evenss (approved):
LGTM

PR #361, opened 2026-03-18
pyobvector dependencies

Ripcord55:
## Summary
Resolve pyobvector dependencies

## Solution Description
Pyobvector requires sqlglot>=26.0.1, but there is no upper limit set. The problem is that the new version of sqlglot (possibly after v26.25.0) has changed the API and removed the direct export of expressions.
Pyproject.toml: Add sqlglot>=26.0.1,<26.25.0 version limit, ensure installation compatibility version


Teingi:
@Evenss Please review this PR.

Evenss (approved):
LGTM

PR #900, opened 2026-04-09
Action Repair

Ripcord55:
## Summary
Partial case fixes
## Solution Description
Modify the logic of CLI related cases to avoid instability


Teingi (approved):
LGTM

PR #901, opened 2026-04-12
fixed(core): OpenAI LLM

Teingi:
调用 response_format: {"type": "json_object"} 做事实抽取和记忆决策。
- 会出现 message.content 为空、None，或非字符串（如分段列表），旧逻辑直接 json.loads(response)，触发 Expecting value: line 1 column 1 (char 0)，或后续解析失败。

新增 llm_json_text_with_fallback：先带 json_object 请求，若正文为空再 不带 response_format 重试；新增 parse_fact_extraction_json / parse_memory_actions_json，用 extract_json、parse_json_from_text 等做 容错解析，避免对空串裸 json.loads。

---
Call response_format: {"type": "json_object"} to make fact extraction and memory decisions.
- It will appear that message.content is empty, None, or non-string (such as a segmented list), and the old logic directly json.loads(response) triggers Expecting value: line 1 column 1 (char 0), or subsequent parsing fails.

Added llm_json_text_with_fallback: first request with json_object, if the text is empty, try again without response_format; add parse_fact_extraction_json / parse_memory_actions_json, use extract_json, parse_json_from_text, etc. for fault-tolerant parsing to avoid naked json.loads of empty strings.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -14,7 +14,7 @@ TIMEZONE=Asia/Shanghai
 # 1. Database Configuration (Required)
 # =============================================================================
 # Choose your database provider: sqlite, oceanbase, postgres
-DATABASE_PROVIDER=sqlite
+DATABASE_PROVIDER=oceanbase
 
 # -----------------------------------------------------------------------------
 # SQLite Configuration (Default - Recommended for development)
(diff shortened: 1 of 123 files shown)


Evenss (approved):
LGTM

PR #905, opened 2026-04-17
Add ob441-reg and sparse case

Ripcord55:
## Summary
Version regression of ob441 and sparse vector correlation test case

## Solution Description
Add OB441-reg.yml as the regression test for the ob441 version. Only manual triggering is allowed. docker deploys ob441 to test local mixed search cases and sparse vector cases test_sparse_search.py as a sparse vector-related test case.
image


Teingi (approved):
LGTM

PR #906, opened 2026-04-21
feat(examples): Add MoonBit client example for PowerMem HTTP API

tonyfettes:
## Summary

This PR adds a simple, lightweight MoonBit client example demonstrating how to integrate PowerMem's intelligent memory capabilities into MoonBit applications.

## Solution Description

It mirrors the structure of `examples/go/` — a self-contained, runnable demo with an inline HTTP client, inline request/response types, and a single `main` that walks through the core API surface. No dependencies beyond `moonbitlang/core` and `moonbitlang/async` (MoonBit's stdlib-tier packages), so the directory is copy-out runnable.

Uses the same `POWERMEM_BASE_URL` / `POWERMEM_API_KEY` environment variables as the Go example.

Tested against a local server (`powermem-server --host 0.0.0.0 --port 8000`, SQLite + Ollama.

### DIFF
--- a/examples/moonbit/.gitignore
+++ b/examples/moonbit/.gitignore
@@ -1,0 +1,2 @@
+_build/
+.mooncakes/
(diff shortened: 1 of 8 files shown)


Teingi (approved):
LGTM

PR #910, opened 2026-04-24
fix(intelligence): store memory_type in metadata JSON column and enab…

wayyoungboy:
…le promotion

Root cause (three layers):
1. _create_memory_internal (infer=True path) never called on_add, so memory_type was never set after intelligent add.
2. _simple_add merged extra_fields into top-level payload which OceanBase silently drops — only the metadata JSON column is persisted.
3. on_get read memory.get("memory_type") which is always None since OceanBase storage.get_memory() only surfaces 8 fixed fields, not top-level payload keys.

Fix:
- _create_memory_internal: call on_add and merge result into enhanced_metadata
- _simple_add / _simple_add_async: merge extra_fields into enhanced_metadata instead of memory_data top-level
- on_get: normalize memory_type/access_count/importance_score from metadata fallback; write all updates back through the metadata JSON column

### DIFF
--- a/src/powermem/core/async_memory.py
+++ b/src/powermem/core/async_memory.py
@@ -539,11 +539,13 @@ class AsyncMemory(MemoryBase):
         # enhanced_metadata = await self.intelligence.process_metadata_async(content, metadata)
         enhanced_metadata = metadata  # Use original metadata without LLM evaluation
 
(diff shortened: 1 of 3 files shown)


Teingi (approved):
LGTM

PR #918, opened 2026-05-06
fix(i18n): add missing zh-CN translations for theme switch component

### LINKED ISSUE #915: [Enhancement]: Set the language to Simplified Chinese, but the theme switch will still display English
### Description

Set the language to Simplified Chinese, but the theme switch will still display English
Image
powermem 1.1.1-Dashboard

wayyoungboy:
ThemeToggle used hardcoded English strings instead of i18n keys. Add theme.light/dark/system keys to both en.json and zh.json, and update the component to use useTranslation.

Fixes #915


wayyoungboy:
image

image

Teingi (approved):
LGTM

Issue #913, opened 2026-04-24
[Bug]: 应用侧开启记忆图谱开关，存储失败 || [Bug]: When the memory map switch is turned on on the application side, the storage fails


haixinke:
### Describe the bug

源码中将整个 `MemoryConfig` 对象传递给 `GraphStoreFactory.create()`，而不是只传递 `graph_store` 字典，导致初始化失败，应该只传递`graph_store` 相关配置

### Environment

v1.0.2

### Fast reproduce steps

初始化：
from powermem import UserMemory
self.memory_client = UserMemory(config=powermem_config)

powermem_config这个字典配置中带有图谱记忆存储相关配置

保存操作：
 result = self.memory_client.add(
                    messages=messages,
                    user_id=self.role_id,
                    native_language="zh",  # Force profile extraction in Chinese
                    # profile_type="topics",  # Extract structured topics (JSON) instead of plain text content
                    profile_type="content",
                    include_roles=["user"]  # Only extract profile from user messages, not AI assistant responses
                )

保存是报错的，后台日志报错

### Expected behavior

正常保存用户记忆，包括图谱记忆

### Actual behavior

所以记忆保存失败

---
### Describe the bug

In the source code, the entire `MemoryConfig` object is passed to `GraphStoreFactory.create()` instead of just passing the `graph_store` dictionary, resulting in initialization failure. Only the `graph_store` related configuration should be passed

### Environment

v1.0.2

### Fast reproduction steps

Initialization:
from powermem import UserMemory
self.memory_client = UserMemory(config=powermem_config)

The powermem_config dictionary configuration contains configurations related to map memory storage.

Save operation:
 result = self.memory_client.add(
                    messages=messages,
                    user_id=self.role_id,
                    native_language="zh", # Force profile extraction in Chinese
                    # profile_type="topics", # Extract structured topics (JSON) instead of plain text content
                    profile_type="content",
                    include_roles=["user"] # Only extract profile from user messages, not AI assistant responses
                )

An error is reported when saving, and an error is reported in the background log.

### Expected behavior

Save user memory normally, including map memory

### Actual behavior

So memory saving failed

PR #919, opened 2026-05-06
Fixed the issue where a TypeError was thrown when using MemoryGraph due to GraphStoreFactory refusing to accept a complete configuration object.

wayyoungboy:
Fixed the issue where a TypeError was thrown when using MemoryGraph due to GraphStoreFactory refusing to accept a complete configuration object.
Fixs #913


Teingi (approved):
LGTM

Issue #941, opened 2026-05-27
Support a built-in default embedding model for true zero-config startup (especially in powermem-ts)


happyhust:
### Problem

Today, when a user doesn't configure an `embedder`, `MemoryConfig` falls back to `OpenAIEmbeddingConfig` with model `text-embedding-3-small`. While this is a "default" in code, it still **requires the user to provide `OPENAI_API_KEY`** before the system can actually run. This means there is no real zero-config path — first-time users always have to obtain an API key from an external provider before they can even try the project.

This friction is especially painful for **powermem-ts**, where we want users to be able to clone, install, and run with literally no configuration at all.

### Proposal

Support a true **built-in default embedding model** that works without any external API key or network call. When the user does not configure `embedder`, the system should automatically fall back to this default and start successfully.

Possible implementations:

- A small, locally-runnable model (e.g. a lightweight sentence-transformers / MiniLM-class model bundled or auto-downloaded on first use).
- Or a deterministic, dependency-free fallback embedder for demo / quick-start purposes, clearly labeled as "default — replace for production use."

The exact model choice can be discussed, but the key requirements are:

- **No API key required.**
- **No external service required at startup** (auto-download on first use is acceptable; runtime calls to external APIs are not).
- **Reasonable quality** so the out-of-the-box experience actually demonstrates the system's value.
- **Easy to override** via config when users want to switch to OpenAI / Qwen / SiliconFlow / etc.

### Why this matters

- Enables **true zero-config startup** — `clone → install → run` with no setup steps.
- Particularly important for **powermem-ts**, where the goal is a frictionless first-run experience.
- Complements the broader config-simplification effort (see #940).
- Lets users evaluate the system before committing to a paid API provider.

### Acceptance criteria

- [ ] A built-in default embedder is available and used automatically when `embedder` is not configured.
- [ ] The default embedder works with no API key and no manual setup.
- [ ] powermem-ts can start and serve a basic memory add / search flow with zero configuration.
- [ ] Documentation clearly states what the default is, its quality/limitations, and how to switch to a production-grade embedder.

PR #945, opened 2026-05-28
feat(embedder): zero-config startup with built-in default embedder; split .env.example (closes #940 #941 #949 #948 #)

### LINKED ISSUE #940: Too many config options — simplify toward zero-config startup, plus split .env.example into minimal vs. full
### Problem

The current configuration surface is overwhelming. New users face a wall of options before they can even get the system running, which raises the barrier to entry significantly. Most options have reasonable defaults and shouldn't be required for a first-time launch.

### Proposal

**1. Aim for zero-config startup**

Make the system runnable out of the box with minimal (ideally zero) required configuration. Only force users to set things that genuinely have no safe default (e.g. API keys for external services). Everything else should fall back to sensible defaults and be opt-in.

**2. Split `.env.example` into two files**

- **`.env.example`** — A minimal file that lists **only the configuration items that are strictly required** to start the system. Each item should have clear inline comments explaining what it is, why it's required, and how to obtain a value (e.g. where to get an API key). At the top of the file, add a note like:

  > "These are the only variables required to run the system. If you want to optimize for peak performance and quality, see `.env.example.full` for the complete set of tunable options."

- **`.env.example.full`** — A comprehensive file that lists **every available configuration option**, grouped by component (LLM, embedder, vector store, graph store, reranker, etc.), with comments describing each option's effect, defaults, and tuning guidance. This serves as both documentation and a reference for advanced users.

### Why this matters

- Lowers the barrier to first-run success — users can clone and start in minutes, not hours.
- Makes the "happy path" obvious and the "power user path" still fully discoverable.
- Reduces support burden from confusion over which knobs are actually required.

### Acceptance criteria

- [ ] System starts successfully with a minimal `.env` derived from `.env.example` (only required keys set).
- [ ] `.env.example` contains only required variables, with explanatory comments.
- [ ] `.env.example.full` contains all variables, grouped and commented.
- [ ] README points to both files and explains when to use each.

happyhust:
## Summary

This PR resolves **#940** (config simplification) and **#941** (built-in default embedder so zero-config startup actually works) in a single, narrowly-scoped change.

### What changes

**1. Built-in default embedder (#941)**

`MemoryConfig().embedder` previously fell back to `OpenAIEmbeddingConfig` + `text-embedding-3-small`, which made the "default" path require `OPENAI_API_KEY` before the system could run — there was no real zero-config story.

This PR adds a built-in default embedder that mirrors **the exact model used by pyseekdb's [`DefaultEmbeddingFunction`](https://github.com/oceanbase/pyseekdb/blob/develop/src/pyseekdb/client/embedding_function.py)**: `sentence-transformers/all-MiniLM-L6-v2` via ONNX runtime, 384 dimensions. The model auto-downloads to `~/.cache/pyseekdb/onnx_models/` on first use and runs entirely locally afterwards.

- New `PyseekdbDefaultEmbedding` wrapper class
- New `PyseekdbDefaultEmbeddingConfig` registered under provider name `"default"`
- `MemoryConfig.embedder.default_factory` switched from `OpenAIEmbeddingConfig` to `PyseekdbDefaultEmbeddingConfig`
- pyseekdb is **already a runtime dependency** (`pyseekdb>=0.1.0` in `pyproject.toml`), so this adds no new transitive dependencies

After this change, `MemoryConfig()` constructs successfully with no env vars and no API key. Acceptance criteria from #941:
- [x] A built-in default embedder is available and used automatically when `embedder` is not configured
- [x] The default embedder works with no API key and no manual setup
- [x] Documentation clearly states what the default is, its quality/limitations, and how to switch to a production-grade embedder

**2. `.env.example` split (#940)**

The 384-line `.env.example` was the "wall of options" #940 calls out. This PR:

- Moves the full reference to **`.env.example.full`** (every existing knob, grouped by component, unchanged content) with a new header explaining when to copy it vs the minimal version.
- Trims **`.env.example`** down to the strictly-required keys: just the LLM provider block (`LLM_PROVIDER`, `LLM_API_KEY`, `LLM_MODEL`). Everything else now defaults sanely: SQLite at `./powermem.db`, local default embedder, no graph store, no reranker, etc.
- Updates `README.md`, `README_CN.md`, and `README_JP.md` to point at both files and to call out the new zero-config defaults.

Acceptance criteria from #940:
- [x] System starts successfully with a minimal `.env` derived from `.env.example` (only required keys set)
- [x] `.env.example` contains only required variables, with explanatory comments
- [x] `.env.example.full` contains all variables, grouped and commented
- [x] README points to both files and explains when to use each

### What does **not** change

- No existing provider behaviour is altered; OpenAI / Qwen / SiliconFlow / etc. still work exactly as before.
- Anyone with an existing `.env` that sets `EMBEDDING_PROVIDER=openai` (or qwen / etc.) is unaffected — the new default only kicks in when nothing else is configured.
- No new dependencies; pyseekdb is already in `pyproject.toml`.

### Test plan

- [x] `pytest tests/unit/test_pyseekdb_default_embeddings.py` — 8 new tests covering: embed round-trip (mocked, no model download in CI), embed_batch, empty input, None rejection, default model/dim match pyseekdb, provider registry, factory resolution, and the headline `MemoryConfig()` zero-config invariant.
- [x] `pytest tests/unit` — 186 tests pass (full unit suite, no regressions from this branch).
- [x] Clean-env smoke check: `env -i python -c "from powermem.configs import MemoryConfig; cfg = MemoryConfig(); assert cfg.embedder.provider == 'default' and cfg.embedder.api_key is None"` succeeds with no `.env`, no env vars.
- [ ] Optional: maintainer to confirm first-run ONNX model download path in `~/.cache/pyseekdb/...` matches expectations on Windows / macOS / Linux.

### Notes for reviewers

- The new `.env.example` is intentionally minimal — only the LLM API key remains, because the LLM is the one component without a credential-free default. Everything else has a working default.
- The new embedder class is a thin wrapper around pyseekdb's existing `DefaultEmbeddingFunction`; we deliberately did **not** re-implement the ONNX download/inference path so the two systems stay in lock-step on the default model.


happyhust:
**Update — added seekdb as the default database (sharing the OceanBase backend).**

Following the same zero-config principle as the embedder, this branch now also makes **embedded SeekDB** the default vector store. SeekDB and OceanBase are the same engine (same SQL surface, same `OceanBaseVectorStore` Python class) — only configuration differs (embedded vs. remote cluster). Surfacing that explicitly removes the "why does my default config write to `./seekdb_data` under the `oceanbase` provider name?" confusion.

### What changed in this commit

- New `SeekDBConfig` / `SeekDBGraphConfig` subclasses (in `storage/config/oceanbase.py`), registered under provider name `"seekdb"`, both with `_class_path` pointing at the existing `OceanBaseVectorStore` / `MemoryGraph` — i.e. **no new backend code**.
- `SeekDBConfig` only overrides what differs from `OceanBaseConfig`: provider name, embedded-mode defaults (empty `host`, on-disk `./seekdb_data`), and `SEEKDB_*` env aliases. Each `SEEKDB_*` alias falls back to the corresponding `OCEANBASE_*` alias, so flipping `DATABASE_PROVIDER` between `seekdb` and `oceanbase` doesn't require rewriting `.env`.
- Defaults switched everywhere it matters:
  - `MemoryConfig.vector_store.default_factory`: `SQLiteConfig` → `SeekDBConfig`
  - `DatabaseSettings.provider` env default: `"sqlite"` → `"seekdb"`
  - `Memory()` storage_type fallback: `"oceanbase"` → `"seekdb"`
  - Deprecated `create_config()` / `CreateConfigOptions` default also updated
- `.env.example.full` reorders the database section to lead with embedded SeekDB; `.env.example` (minimal) advertises SeekDB + the local embedder as the zero-config defaults. README / README_CN / README_JP follow suit.

### Test plan

- `pytest tests/unit/test_seekdb_default_storage.py` — 7 new tests pin the contract: provider registration, shared class path between `seekdb` and `oceanbase`, embedded-mode defaults, `SEEKDB_*` env aliases, `MemoryConfig()` default, `DatabaseSettings()` default.
- `pytest tests/unit` — **193 passed** (up from 186 in the previous commit), no regressions.
- Clean-env smoke check:
  ```
  env -i python -c "from powermem.configs import MemoryConfig; cfg = MemoryConfig(); \
    assert type(cfg.vector_store).__name__ == 'SeekDBConfig' and cfg.vector_store.host == ''"
  ```
  succeeds with no `.env` and no env vars.

Now `MemoryConfig()` truly starts end-to-end with **no API key, no separate database service, and no configuration**: embedded SeekDB on disk + local `all-MiniLM-L6-v2` embedder.

PR #950, opened 2026-05-29
fix(logging): SDK file logging, trace context, and gzip rotation

wayyoungboy:
## Summary

- Connect `LOGGING_*` environment variables to the `powermem.*` SDK logger tree so SDK/storage logs are written to the configured file (default `./logs/powermem.log`) with rotation and optional gzip backups.
- Initialize logging from `Memory.__init__` and from server `setup_logging()` so both standalone SDK and HTTP server paths get file handlers.
- Propagate `request_id`, `user_id`, and `agent_id` into SDK logs via `contextvars` and `LoggingMiddleware` (parsed from query params or `/users/{id}` / `/agents/{id}` paths).
- Fix compressed log rotation: backups are stored as `base.N.gz` so `backupCount` semantics remain correct (avoids breaking the standard `RotatingFileHandler` chain).

## Motivation

`LOGGING_FILE` and related settings existed in `.env.example` but were not applied to the `powermem` logger namespace, so SDK logs often never reached `powermem.log`. This change makes configuration effective and aligns log format with trace fields for request correlation.

## Changes

| Area | What changed |
|------|----------------|
| `src/powermem/logging_config.py` | New module: `setup_powermem_logging()`, `CompressingRotatingFileHandler`, size parsing |
| `src/powermem/log_context.py` | New: async-safe trace context + `TraceContextFilter` |
| `src/powermem/core/memory.py` | Call `setup_powermem_logging()` on init |
| `src/server/middleware/logging.py` | Call SDK logging setup; propagate trace context per request |
| `src/powermem/config_loader.py`, `.env.example` | Default `LOGGING_FORMAT` includes trace fields |
| `tests/unit/test_logging_config.py` | Unit tests for parsing, setup, rotation, trace filter |

**Note:** HTTP access logs still use server config (e.g. `server.log`); this PR only configures the `powermem.*` SDK tree.

## Test plan

- [x] `uv run pytest tests/unit/test_logging_config.py`
- [ ] Start server with `LOGGING_FILE=./logs/powermem.log`, send a request with `user_id` / `agent_id` (query or path), confirm trace fields appear in `powermem.log`
- [ ] Instantiate `Memory()` in a script without server; confirm logs land in the configured file
- [ ] With `LOGGING_COMPRESS_BACKUPS=true`, trigger rotation and confirm backups are `*.N.gz` and count ≤ `LOGGING_BACKUP_COUNT`

## Security / privacy

- No secrets or credentials in this PR (only `.env.example` placeholder changes).
- Default log format includes `user_id` and `agent_id`; adjust `LOGGING_FORMAT` in production if those identifiers should not be persisted in log files.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -223,7 +223,7 @@ AUDIT_LOG_ROTATION_SIZE=100MB
 # =============================================================================
 # General logging settings
 LOGGING_LEVEL=DEBUG
-LOGGING_FORMAT=%(asctime)s - %(name)s - %(levelname)s - %(message)s
+LOGGING_FORMAT=%(asctime)s - %(name)s - %(levelname)s - [%(request_id)s] [%(user_id)s] [%(agent_id)s] - %(message)s
 LOGGING_FILE=./logs/powermem.log
 LOGGING_MAX_SIZE=100MB
 LOGGING_BACKUP_COUNT=5
--- a/src/powermem/config_loader.py
+++ b/src/powermem/config_loader.py
@@ -107,7 +107,7 @@ class LoggingSettings(_BasePowermemSettings):
 
     level: str = Field(default="DEBUG")
     format: str = Field(
-        default="%(asctime)s - %(name)s - %(levelname)s - %(message)s"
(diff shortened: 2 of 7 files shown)


lightzt99:
LGTM

PR #952, opened 2026-06-01
fix(logging): complete logging system gaps across SDK, server, CLI, a…

wayyoungboy:
…nd MCP

- Use module-level logger instead of root logger in 6 LLM integration modules
- Add exc_info=True to ~100 logger.error() calls in agent layer except blocks
- Attach TraceContextFilter to server and uvicorn loggers for request tracing
- Replace plain FileHandler with CompressingRotatingFileHandler in audit logger
- Make server log rotation configurable via POWERMEM_SERVER_LOG_MAX_SIZE env var
- Add JSON logging mode (LOGGING_FORMAT=json) to SDK via JsonLogFormatter
- Share single file handler across uvicorn loggers to prevent rotation races
- Replace traceback.print_exc() with logger.exception() in CLI commands
- Export all fields from LoggingSettings.to_config() and AuditSettings.to_config()
- Add module logger to MCP server

### DIFF
--- a/src/powermem/agent/agent.py
+++ b/src/powermem/agent/agent.py
@@ -153,7 +153,7 @@ class AgentMemory:
             logger.info(f"AgentMemory initialized in {self.mode} mode")
             
         except Exception as e:
-            logger.error(f"Failed to initialize AgentMemory: {e}")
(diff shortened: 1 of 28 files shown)


lightzt99:
LGTM

PR #961, opened 2026-06-03
fix(claude-code-plugin): improve SETUP.md + fix Anthropic temperature/top_p conflict

lightzt99:
## Summary

- Auto-detect LLM config from `~/.claude/settings.json` in Step 2
- Document Anthropic model name normalization rule (dots → dashes)
- Fix `plugin uninstall` idempotency (`2>/dev/null || true`)
- Parallelize embedding model download with hook build / plugin staging
- Add error entry [E010] for Anthropic `temperature`+`top_p` conflict
- **Bug fix**: `AnthropicLLM.generate_response` now drops `top_p` before calling the API

## Motivation

Six failure modes were observed in the original PR. This update addresses three more
discovered during actual setup on macOS:

| # | Problem | Impact |
|---|---------|--------|
| 7 | `settings.json` model names use dots (`4.6`) but Anthropic API requires dashes (`4-6`) | 404 invalid_model |
| 8 | `plugin uninstall` fails with exit code 1 when plugin not yet installed | Setup script aborts mid-run |
| 9 | `AnthropicLLM` sends both `temperature` and `top_p` to Anthropic API | 400 invalid_params — every memory write silently drops all extracted facts |

Issue 9 was verified by direct API call: sending both parameters always returns
`400 - temperature and top_p cannot both be specified for this model`.

## Changes

| File | What changed |
|------|-------------|
| `apps/claude-code-plugin/SETUP.md` | Auto-detect flow, model name normalization note, parallel model download, uninstall `\|\| true`, [E010] error entry |
| `src/powermem/integrations/llm/anthropic.py` | `params.pop("top_p", None)` before `client.messages.create()` |

## Test plan

- [x] Confirm no API key echoes in the setup flow
- [x] Confirm ModelScope download + bridge script produces a cache that `try_to_load_from_cache` recognizes
- [x] Confirm polling health check waits correctly on a slow first launch
- [x] Direct API verification: `temperature`+`top_p` → 400; `temperature`-only → 200; `top_p`-only → 200
- [x] Full setup round-trip: write probe memory, search returns `dragonfruit-zx9`, cleanup, global hook calls confirmed in server.log

## Security / privacy

- The DATA SAFETY masking rules apply to the AI assistant executing SETUP.md, not to any code path. No secrets are logged or transmitted by this change.
- The bridge script copies model files between local cache directories only; no network access after the initial ModelScope download.

### DIFF
--- a/apps/claude-code-plugin/SETUP.md
+++ b/apps/claude-code-plugin/SETUP.md
@@ -14,10 +14,18 @@ source tree or not, ask you for the few required secrets, and wire PowerMem up a
 Set up PowerMem memory for Claude Code on this machine **globally**. Do the whole
 integration autonomously and ask me for any secret you need — never invent credentials.
 
+**🔒 DATA SAFETY — API Key Masking (MANDATORY):**
+When displaying ANY `.env` content — current values, proposed changes, confirmation
+summaries, or any other output — you MUST mask `LLM_API_KEY` and any other secret
+values (passwords, tokens, keys):
+- **Key ≥ 10 chars:** show only first 4 + last 4 characters (e.g. `sk-a…b12x`)
+- **Key < 10 chars:** show `***`
+Non-secret values (provider, model, base URLs, storage type, etc.) may be shown in full.
+
 **⚠️ `.env` changes always require user approval.** Before modifying `.env` for any
 reason (LLM config, embedder settings, storage switches, ...), show the user the
-current values, propose the exact change, and WAIT for confirmation. Never patch
-`.env` silently.
+**masked** current values (using the rules above), propose the exact change, and WAIT
+for confirmation. Never patch `.env` silently.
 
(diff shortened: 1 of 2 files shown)


wayyoungboy (changes requested):
Thanks for improving the setup guide. One command should be corrected before merge: this PR currently recommends:\n\n```bash\npip install -e . 'powermem[server]' 'powermem[mcp]'\n```\n\nFor a local editable source install, please use the extras on the editable requirement itself so pip installs this checkout with both extras enabled:\n\n```bash\npip install -e '.[server,mcp]'\n```\n\nThat keeps the source-path setup unambiguous and avoids resolving extras through a separate package requirement.

lightzt99:
The pip install command has been updated to `pip install -e '.[server,mcp]'` (and similar extras-on-editable form throughout the file). Could you take another look when you have a chance? Thanks!

wayyoungboy:
LGTM

PR #966, opened 2026-06-05
feat(search): wire dashboard to semantic search, add time_range + sort

LINxiansheng:
## Solution Description

1. fix bug when wire dashboard to semantic search
2. add time_range + sort between frontend and backend


wayyoungboy:
Review note: this needs fixes before merge.

1. `time_range` filtering can throw on common naive datetime strings. `parse_time_range_cutoff()` returns an aware UTC datetime, but stored `created_at` values may be naive strings such as `2026-06-01 12:00:00`. The comparisons in `src/server/api/v1/memories.py:239` and `src/server/api/v1/search.py:107` can then raise `TypeError: can't compare offset-naive and offset-aware datetimes`.

2. The list endpoint applies `time_range` after fetching only the first 1000 memories (`src/server/api/v1/memories.py:220`). For stores with more than 1000 rows, deep pages and `total` can be wrong because filtering happens on a truncated in-memory set.

3. Search time filtering only filters and sorts the recalled candidate window (`min(limit * 5, 200)` in `src/server/api/v1/search.py:100`). That does not guarantee global top-K/newest results within the time range.

Also consider validating POST search `time_range`, `sort_by`, and `order` the same way the GET list endpoint does, so unsupported values fail predictably.

wayyoungboy:
Updated review for the latest head (`5086924`): this still needs fixes before merge.

1. `time_range` filtering can still raise on naive datetimes. `parse_time_range_cutoff()` returns an aware UTC datetime, while `_coerce_datetime()` in `src/server/api/v1/search.py:25` and `_created_at()` in `src/server/api/v1/memories.py:230` can return naive datetimes for common stored values like `2026-06-01 12:00:00`. The comparisons at `search.py:44` and `memories.py:241` will then raise `TypeError: can't compare offset-naive and offset-aware datetimes`. `_apply_sort()` also uses `datetime.min` as a naive fallback at `search.py:60`, so sorting can fail when aware timestamps and missing/invalid timestamps are mixed. Please normalize all parsed timestamps and fallbacks to one timezone policy before comparing/sorting.

2. List `time_range` pagination/total are still based on a truncated first page. `src/server/api/v1/memories.py:220` fetches only the first 1000 rows, filters in memory, then slices. For stores with more than 1000 memories, deep pages and `total` will be wrong, and valid memories outside the first 1000 can never appear. This should either be pushed into the storage query/count layer or clearly not exposed as paginated/full-total semantics.

3. Search `time_range`/sort still operate only on the recalled candidate window (`min(limit * 5, 200)` at `src/server/api/v1/search.py:95`). That may be acceptable as an approximate semantic-search behavior, but it should be documented as candidate-window post-filtering rather than treated as global newest/top-K within a time range.

Nonblocking: POST search fields `time_range`, `sort_by`, and `order` in `src/server/models/request.py:75` are still unvalidated; invalid values are silently ignored or coerced to desc-like behavior. Consider using Literal/pattern validation to match the GET endpoint contract.

lightzt99:
LGTM

PR #967, opened 2026-06-05
fix(ci): isolate regression embedding dimensions

wayyoungboy:
## Summary

  Fix regression CI embedding dimension conflicts after the default embedder changed to the local 384-dim `all-MiniLM-L6-v2`.

  ## Changes

  - Pin the main regression workflow to Qwen `text-embedding-v4` with 1536 dimensions.
  - Move the main regression OceanBase collection to `memories_1536`.
  - Make the multimodal scenario use `OCEANBASE_COLLECTION` instead of hardcoding `memories`.
  - Add a dedicated 384-dim regression scenario using the built-in default embedder on `memories_384`.

  ## Why

  Regression tests previously shared the same OceanBase collection name, `memories`.

  After the zero-config default embedder change, regular scenarios created a 384-dim `memories` table, while multimodal tests still expected 1536 dimensions.
  OceanBase correctly rejected the mismatch.

  This separates the 1536-dim and 384-dim paths so both are covered without table schema conflicts.

  ## Testing

  - `python -m py_compile tests/regression/test_scenario_7_multimodal.py tests/regression/test_scenario_9_default_embedding_384.py`
  - `pytest tests/unit/test_pyseekdb_default_embeddings.py -q`
  - `pytest tests/regression/test_scenario_9_default_embedding_384.py -q`
  - `pytest tests/regression/test_scenario_7_multimodal.py --collect-only -q`

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -145,13 +145,18 @@ jobs:
           set_env_var OCEANBASE_USER root
           set_env_var OCEANBASE_PASSWORD ""
           set_env_var OCEANBASE_DATABASE powermem
-          set_env_var OCEANBASE_COLLECTION memories
+          set_env_var OCEANBASE_COLLECTION memories_1536
           set_env_var DATABASE_PROVIDER oceanbase
           set_env_var LLM_PROVIDER siliconflow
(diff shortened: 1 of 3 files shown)


lightzt99:
LGTM

PR #971, opened 2026-06-05
Fix quoted OceanBase password handling

### LINKED ISSUE #921: [Bug]: When starting the powermem-server using docker, if the password of the OceanBase database is wrapped in single quotes, the connection to the database fails
### Describe the bug

When starting the powermem-server using docker, if the password of the OceanBase database is wrapped in single quotes, the connection to the database fails
Image
Image

The actual password value in the Python process:
Image

### Environment

powermem 1.1.1

### Fast reproduce steps

1. Download the docker package and docker load
2. set OCEANBASE_PASSWORD='powermem'
3. Command line execution:
```python
python - <<'PY'
import os
print("OCEANBASE_PASSWORD =", repr(os.getenv("OCEANBASE_PASSWORD")))
PY
``` 

### Expected behavior

Consistent with non-Docker environments, automatically remove the single quotes enclosing passwords

### Actual behavior

N/A

### Additional context

N/A

wayyoungboy:
## Summary
- strip matching single or double wrapping quotes from OceanBase and graph-store passwords loaded from environment settings
- apply the same normalization when OceanBaseVectorStore is built directly from connection_args
- consolidate quote-stripping into a shared helper and document the compatibility assumption
- add regression coverage for docker-style quoted OCEANBASE_PASSWORD values

Fixes #921.

## Verification
- `python3.11 -m pytest tests/unit/test_config_loader.py tests/unit/test_oceanbase.py tests/unit/test_issue_921_oceanbase_password.py -q`
- `python3.11 -m compileall -q src/powermem/storage/config/oceanbase.py src/powermem/storage/config/base.py src/powermem/storage/oceanbase/oceanbase.py src/powermem/utils/strings.py tests/unit/test_issue_921_oceanbase_password.py`
- `git diff --check`
- sensitive-pattern scan over changed storage/test files: no matches

### DIFF
--- a/src/powermem/storage/config/base.py
+++ b/src/powermem/storage/config/base.py
@@ -2,6 +2,7 @@ from typing import Any, ClassVar, Dict, Optional, Union
 from pydantic import AliasChoices, Field, field_validator
 from pydantic_settings import BaseSettings
 from powermem.settings import settings_config
+from powermem.utils.strings import strip_wrapping_quotes
 
 
 class BaseVectorStoreConfig(BaseSettings):
@@ -161,11 +162,7 @@ class BaseGraphStoreConfig(BaseVectorStoreConfig):
     @field_validator("password", mode="before")
     @classmethod
     def _strip_wrapping_quotes_from_password(cls, value: Any) -> Any:
-        if isinstance(value, str) and len(value) >= 2:
-            stripped = value.strip()
-            if stripped[0] == stripped[-1] and stripped[0] in {"'", '"'}:
-                return stripped[1:-1]
-        return value
+        return strip_wrapping_quotes(value)
     
     db_name: str = Field(
         default="test",
--- a/src/powermem/storage/config/base.py
+++ b/src/powermem/storage/config/base.py
@@ -157,6 +157,15 @@ class BaseGraphStoreConfig(BaseVectorStoreConfig):
         ),
         description="Database password"
     )
+
+    @field_validator("password", mode="before")
+    @classmethod
+    def _strip_wrapping_quotes_from_password(cls, value: Any) -> Any:
+        if isinstance(value, str) and len(value) >= 2:
+            stripped = value.strip()
+            if stripped[0] == stripped[-1] and stripped[0] in {"'", '"'}:
+                return stripped[1:-1]
+        return value
     
     db_name: str = Field(
         default="test",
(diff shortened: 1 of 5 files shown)


lightzt99:
@wayyoungboy  A few notes on the implementation:

1. Duplicated logic: `_strip_wrapping_quotes` appears in three places (`base.py` validator, `oceanbase.py` validator, and the standalone function in `oceanbase/oceanbase.py`). Worth extracting into a shared utility and importing it everywhere.
2. `len` check inconsistency: `len(value) >= 2` checks the original string, but the quote detection runs on stripped. 
3. Legitimate quoted passwords: A password that genuinely begins and ends with the same quote character would be silently mangled. Worth adding a comment stating the assumption.

**Broader question**: isn't the standard convention to write env vars without wrapping quotes? The shell consumes quotes, and `.env` files are generally expected to hold bare values. By silently stripping quotes here, we risk surprising behavior if a user's actual password happens to start and end with the same quote character.

Is it worth fixing the symptom here, or would it be better to document the correct syntax and surface a clear error when quoted values are detected?

wayyoungboy:
Thanks for the review. I pushed a follow-up commit addressing the implementation notes:

- consolidated the repeated quote handling into `powermem.utils.strings.strip_wrapping_quotes`
- fixed the length check to run against the trimmed string before indexing
- documented the compatibility assumption in the helper docstring
- added focused tests for the trimmed-length edge case and the literal quote tradeoff

On the broader question: I agree bare env values are the standard syntax. I kept the compatibility behavior here because the issue is specifically about `.env` / Docker-style quoted values being passed through literally to OceanBase auth. The helper strips only one matching wrapper pair, so the behavior is scoped, but the docstring now calls out that users who intentionally need quote characters at both ends should configure the password without wrapper syntax.

Verification:

```bash
python3.11 -m pytest tests/unit/test_config_loader.py tests/unit/test_oceanbase.py tests/unit/test_issue_921_oceanbase_password.py -q
python3.11 -m compileall -q src/powermem/storage/config/oceanbase.py src/powermem/storage/config/base.py src/powermem/storage/oceanbase/oceanbase.py src/powermem/utils/strings.py tests/unit/test_issue_921_oceanbase_password.py
git diff --check
```

lightzt99:
After careful consideration, we believe the standard `.env` convention treats everything to the right of = as a literal value — quotes included. A value written as `OCEANBASE_PASSWORD='mypassword'` should be interpreted as the string `'mypassword'` with the quotes, not `mypassword`. Under that convention, the quoted-password scenario is not a code defect.

PR #973, opened 2026-06-05
Fix SDK add when embedded seekdb is owned by server

### LINKED ISSUE #926: [Bug]: Add memory using the SDK mode under the premise of enabling the API Server
### Describe the bug

Add memory using the SDK mode under the premise of enabling the API Server
Image
set

### Environment

1.1.1
pyseekdb

### Fast reproduce steps

1. make server-start(The configuration file uses the pyseekdb mode)
2. memory.add()

### Expected behavior

Memories can be added normally

### Actual behavior

An error "[seekdb] db  /data/1/powermem/powermem/seekdb_data the opened by other process"

### Additional context

N/A

wayyoungboy:
## Summary
- detect embedded OceanBase/seekdb initialization errors caused by the DB already being opened by another process
- when the local PowerMem API health endpoint is healthy, fall back to a small HTTP memory client for common SDK operations
- keep unrelated initialization/configuration errors unchanged
- preserve the public SDK bool return contract for fallback delete APIs
- add regression coverage for the seekdb lock fallback and non-lock error behavior

Fixes #926.

## Verification
- `python3.11 -m pytest tests/unit/test_memory.py tests/unit/test_issue_926_memory_http_fallback.py -q` (12 passed)
- `python3.11 -m compileall -q src/powermem/core/memory.py tests/unit/test_issue_926_memory_http_fallback.py`
- `git diff --check`
- sensitive-pattern scan over changed core/test files: no matches

### DIFF
--- a/src/powermem/core/memory.py
+++ b/src/powermem/core/memory.py
@@ -59,6 +59,393 @@ def _forget_marker_updates() -> Dict[str, Any]:
     }
 
 
+def _normalize_api_base_url(raw_url: Optional[str]) -> str:
+    base_url = (raw_url or "http://localhost:8848").rstrip("/")
+    if base_url.endswith("/api/v1"):
+        return base_url
+    return f"{base_url}/api/v1"
+
+
+def _looks_like_embedded_seekdb_lock_error(exc: Exception) -> bool:
+    text = str(exc).lower()
+    return (
+        "opened by other process" in text
+        or "already opened" in text
+        or ("seekdb" in text and "lock" in text)
+        or ("open seekdb failed" in text and "ob_error(4000)" in text)
+    )
+
+
+def _is_embedded_oceanbase_config(
+    storage_type: str,
+    vector_store_config: Dict[str, Any],
+) -> bool:
+    if storage_type.lower() != "oceanbase":
+        return False
+    connection_args = vector_store_config.get("connection_args") or {}
+    host = vector_store_config.get("host") or connection_args.get("host")
+    return not str(host or "").strip()
+
+
+def _messages_to_content(messages: Any) -> str:
+    if isinstance(messages, str):
+        return messages
+    if isinstance(messages, dict):
+        return str(messages.get("content", ""))
+    if isinstance(messages, list):
+        parts: List[str] = []
+        for item in messages:
+            if isinstance(item, dict):
+                content = item.get("content")
+                if content is not None:
+                    parts.append(str(content))
+            elif item is not None:
+                parts.append(str(item))
+        return "\n".join(parts)
+    return str(messages)
+
+
+class _HTTPMemoryClient:
+    """SDK fallback for embedded seekdb when the API server owns the DB."""
+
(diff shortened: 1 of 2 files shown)


wayyoungboy:
I removed the stale self-review note and pushed a follow-up fix for the real issue it pointed out.

The HTTP fallback now preserves the public SDK contract for delete APIs:

- `Memory.delete(...)` calls the fallback client and returns `True` when the HTTP operation succeeds
- `Memory.delete_all(...)` does the same, instead of leaking the fallback client's response dict
- the fallback client can still keep its detailed response internally for filtering/batch behavior

Added regression coverage on the `Memory` wrapper path so this contract stays explicit.

Verification:

```bash
python3.11 -m pytest tests/unit/test_memory.py tests/unit/test_issue_926_memory_http_fallback.py -q
python3.11 -m compileall -q src/powermem/core/memory.py tests/unit/test_issue_926_memory_http_fallback.py
git diff --check
```

lightzt99:
@wayyoungboy Thanks for the fix. Please rebase onto main to resolve the conflict. 

Additionally, found a small contract inconsistency in the reset() fallback:
```
if self._http_client:
    return self._http_client.reset()  # returns {"deleted_count": ..., "memory_ids": [...]}
```
`Memory.reset()` has no return annotation and the normal path returns `None` implicitly. The fallback leaks `_HTTPMemoryClient.reset()`'s dict — same pattern the earlier self-review fixed for `delete`/`delete_all`.

wayyoungboy:
Rebased this PR onto the latest `main` and pushed a single updated commit (`170f673`).

Fixes included:

- resolved the merge conflict by replaying the HTTP fallback change on top of current `main`
- fixed the `Memory.reset()` fallback contract so it now calls the HTTP fallback reset but returns `None`, matching the normal SDK path
- made fallback checks compatible with `Memory` objects constructed without `__init__` by using a safe `_http_client` lookup
- added coverage for the real embedded seekdb lock error shape: `open seekdb failed OB_ERROR(4000): Common error`
- added regression coverage for the `Memory.reset()` fallback contract

Local verification with the project virtualenv:

```bash
source .venv/bin/activate
python -m compileall -q src/powermem/core/memory.py tests/unit/test_issue_926_memory_http_fallback.py
git diff --check
PYTHONPATH=src python -m pytest tests/unit/test_memory.py tests/unit/test_issue_926_memory_http_fallback.py tests/unit/test_intelligence_forget_soft_mark.py -q
# 17 passed
PYTHONPATH=src python -m pytest tests/unit -q
# 361 passed, 1 warning
PYTHONPATH=src python -m pytest tests/integration -q
# 46 passed
PYTHONPATH=src python -m pytest tests/regression/test_api.py -q
# 46 passed
```

I also tried broader local runs:

- `tests/e2e --maxfail=1` stops on missing `DASHSCOPE_API_KEY`, unrelated to this change.
- `tests/regression --maxfail=1` gets through API regression and dashboard skips, then stops around embedded seekdb/native hybrid local state. The fallback now recognizes the generic seekdb lock exception, but that regression target still needs its own local environment isolation.

PR #985, opened 2026-06-08
Fix PyPI publish trigger for tag releases

wayyoungboy:
## Summary

- dispatch publish.yml from the tag release workflow after GitHub Release creation
- keep PyPI trusted publishing bound to publish.yml
- add workflow_dispatch support to publish.yml for manual backfills
- use skip-existing for idempotent PyPI backfill runs

## Validation

- parsed build.yml and publish.yml with PyYAML
- git diff --check


lightzt99:
LGTM

PR #987, opened 2026-06-08
Make soft-forgotten memories rank lower

wayyoungboy:
## Summary
- apply a configurable demotion multiplier to search results marked with `should_forget`
- read soft-forget markers from top-level, metadata, and memory_management metadata
- make relevance calculation work with storage results that expose content as `memory`

## Scope
This keeps the existing search/get lifecycle behavior intact. It only makes the soft-forget marker introduced by #984 affect search ranking.

## Tests
- `python -m pytest tests/unit/intelligence tests/unit/test_intelligence_forget_soft_mark.py`


lightzt99:
LGTM

PR #994, opened 2026-06-09
doc(claude-code-plugin): improve setup auto-detect and dashboard

lightzt99:
## Summary
- Add LLM config auto-detect priority chain: OS env → `~/.claude/settings.json` → manual input, matching Claude Code's own priority
- Add CN region detection via `ipinfo.io` for ModelScope/HuggingFace model download routing
- Add dashboard async background build with `curl` auto-detect after server startup
- Add shell command key leak prevention rules: inline `${VAR:0:4}...${VAR: -4}` masking, `sed -E` redaction, `read -s` for secrets
- Add pip `timeout 60` + 3-retry loop with CN PyPI mirror (`pypi.tuna.tsinghua.edu.cn`) auto-detection
- Update E006 error section and installed-plugin model preload to branch on CN/non-CN region

### DIFF
--- a/apps/claude-code-plugin/SETUP.md
+++ b/apps/claude-code-plugin/SETUP.md
@@ -85,17 +85,19 @@ Follow these steps:
 7. The hook launcher reads `runtime.env`, so once init writes a base URL, prompt
    recall and session-save hooks use that backend automatically.
 
-Model preload uses the same robust path as the source setup below: download from
-ModelScope first, then bridge the files into the HuggingFace hub cache layout that
-the default embedder checks. Do **not** pre-warm with
-`sentence_transformers.SentenceTransformer(...)` or raw `huggingface_hub`; those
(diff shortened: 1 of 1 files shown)


wayyoungboy:
Reviewed the SETUP.md changes. I think two parts should be fixed before merging:

1. In the non-CN model preload path, the snippet imports `huggingface_hub` directly but does not install it first. On a fresh source install this can fail with `ModuleNotFoundError: No module named 'huggingface_hub'`. The non-CN branch should install it with the same pip timeout/retry conventions before running `snapshot_download`.

2. The dashboard section says the build should be asked/launched before starting the server so it can run in parallel with model download and server startup, but the current ordering places it after waiting for model download and after the server health loop. Please move the dashboard prompt/background launch earlier, right after starting the model download background job and before waiting on that job / starting the server.

These are documentation-flow issues, but they can lead users to a failed or slower setup if followed literally.

PR #1001, opened 2026-06-10
fix(skill-store): fix broken embed calls and dedup threshold

jfeng18:
## Summary

The skill distillation storage pipeline (`add_skill()`, `search_skills()`) was non-functional due to two compounding bugs:

- `self._embed()` was never defined on Memory — all calls raised `AttributeError` at runtime. Replaced with `self.embedding.embed()` at 5 call sites.
- Dedup `similarity_threshold` defaulted to 0.75, but the skill store's RRF fusion score maxes out at `4*(1/60) = 0.0667`. The dedup condition could never trigger, making the merge path dead code. Changed default to 0.03.

## Details

**Bug 1: `_embed` undefined**
```python
# Before (crashes):
title_emb = self._embed(title)        # AttributeError

# After (works):
title_emb = self.embedding.embed(title)  # uses the initialized embedder
```

**Bug 2: threshold/score-space mismatch**
```
Max RRF score: 4 * 1/(60+0) = 0.0667
Old threshold: 0.75  →  0.0667 > 0.75 = False (always)
New threshold: 0.03  →  0.0667 > 0.03 = True  (when 2+ lanes match)
```

## Test plan

- [x] Verified `self.embedding` is unconditionally set in `__init__` (line 268)
- [x] Verified `embed()` signature is compatible across all 16 embedder implementations
- [x] Verified no remaining `self._embed(` calls in codebase
- [x] Verified no test mocks `_embed` (no breakage)
- [x] No existing tests assert on old threshold value 0.75
- [ ] CI tests pass
- Note: `add_skill()` and `search_skills()` have no dedicated test coverage (pre-existing gap)

### DIFF
--- a/src/powermem/configs.py
+++ b/src/powermem/configs.py
@@ -230,7 +230,7 @@ class SkillStoreConfig(BaseModel):
     """Configuration for skill sub-store."""
     enabled: bool = Field(default=False, description="Enable skill storage")
     collection_name: Optional[str] = Field(default=None, description="Table name; auto-generated if None")
-    similarity_threshold: float = Field(default=0.75, description="Dedup similarity threshold")
+    similarity_threshold: float = Field(default=0.03, description="Dedup RRF score threshold")
     index_type: Optional[str] = Field(default=None, description="Vector index type (hnsw, ivf, etc.). Falls back to vector_store.config.index_type if not set")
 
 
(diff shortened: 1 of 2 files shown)


lightzt99 (approved):
LGTM.

PR #1006, opened 2026-06-10
fix: apply access reinforcement to memory decay

### LINKED ISSUE #998: reinforcement_factor is configured but never applied to decay
## Observation

`reinforcement_factor` (default 0.3) is:
- Defined in `IntelligentMemoryConfig`
- Stored in memory metadata via `process_memory_metadata()`
- Never consumed by `calculate_decay()` or any other code path

The decay formula `R = exp(-t / (24 * S))` always uses `created_at` as the time anchor. Accessing a memory increments `access_count` and may trigger promotion, but does NOT reset or slow the decay clock. A frequently-accessed `long_term` memory will still be forgotten at ~108 days regardless of engagement.

## Impact

The system lacks true spaced-repetition reinforcement. Memory lifetime is determined entirely by creation time + tier, not by usage patterns. This contradicts the Ebbinghaus model where recall strengthens retention.

## Possible approaches

1. **Reset anchor on access**: On each access, update a `last_reinforced_at` timestamp. Use `max(created_at, last_reinforced_at)` as the decay anchor.
2. **Scale rate on access**: Multiply `effective_rate` by `(1 + reinforcement_factor * access_count)`, making frequently-accessed memories decay slower.
3. **Weighted average anchor**: `effective_time = lerp(created_at, last_accessed_at, reinforcement_factor)` — each access pulls the clock forward.

Each approach has different implications for storage bloat (memories never dying) vs. true reinforcement. Happy to discuss or prototype.

wayyoungboy:
## Summary

Fixes #998 by applying `reinforcement_factor` and `access_count` when resolving the effective Ebbinghaus decay strength. Accessed memories now decay more slowly in both search ranking and forget checks, while unaccessed memories keep the previous behavior.

The effective decay strength now uses diminishing returns for access frequency:

```text
effective_decay_rate = base_decay_rate * (1 + reinforcement_factor * log1p(access_count))
```

This keeps the change scoped to existing fields and does not add schema changes, reset timestamps, or change lazy forgetting semantics.

## Tests

```bash
python -m pytest tests/unit/intelligence/test_ebbinghaus_decay_rate.py -q
# 23 passed

python -m pytest tests/unit/intelligence -q
# 46 passed

python -m compileall -q src/powermem/intelligence/ebbinghaus_algorithm.py tests/unit/intelligence/test_ebbinghaus_decay_rate.py

git diff --check
```

### DIFF
--- a/src/powermem/intelligence/ebbinghaus_algorithm.py
+++ b/src/powermem/intelligence/ebbinghaus_algorithm.py
@@ -6,7 +6,7 @@ This module implements the Ebbinghaus forgetting curve for memory management.
 
 import logging
 import math
-from typing import Any, Dict, Optional, List
+from typing import Any, Dict, Optional, List, Tuple
 from datetime import datetime, timedelta
 from powermem.utils.utils import get_current_datetime
 
@@ -371,8 +371,7 @@ class EbbinghausAlgorithm:
 
     def _resolve_decay_rate(self, memory: Dict[str, Any]) -> float:
         """Resolve the effective decay strength for a memory dict."""
-        meta = memory.get("metadata") or {}
-        intelligence = meta.get("intelligence") or memory.get("intelligence") or {}
+        meta, intelligence = self._resolve_metadata_sections(memory)
 
         memory_type = (
             memory.get("memory_type")
@@ -380,12 +379,13 @@ class EbbinghausAlgorithm:
             or intelligence.get("memory_type")
         )
         if memory_type:
-            return self._get_decay_rate_for_type(memory_type)
+            base_rate = self._get_decay_rate_for_type(memory_type)
+            return self._apply_reinforcement(memory, base_rate)
 
-        stored = (
-            memory.get("decay_rate")
-            or meta.get("decay_rate")
-            or intelligence.get("decay_rate")
+        stored = self._first_present(
+            memory.get("decay_rate"),
+            meta.get("decay_rate"),
+            intelligence.get("decay_rate"),
         )
         if stored is not None:
             try:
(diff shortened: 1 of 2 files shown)


lightzt99:
Thanks for the clean implementation — no schema changes and all three `_resolve_decay_rate` paths covered is the right call.

One concern worth discussing before merging: the linear formula `1 + r × n` has no natural bound. With `reinforcement_factor=0.3`, a `working` memory accessed 100 times gets a 31× multiplier, pushing its half-life to ~32 days — well past `long_term` tier. At 1000 accesses the multiplier is 301×. In practice this means heavily-used memories become effectively immortal, which may or may not be intentional.

**Theoretical case for logarithmic scaling**

The diminishing-returns shape is well-supported in the memory science literature:

- **Power Law of Practice** (Newell & Rosenbloom, 1981): each additional repetition yields less benefit than the previous one — the classical result in cognitive psychology that learning curves follow a power/log shape, not linear.
- **ACT-R memory activation** (Anderson & Schooler, 1991): the activation formula `m = ln(Σ t_j^(-d))` converges to `ln(n)` for uniform access intervals, giving a direct theoretical derivation for log-scaled reinforcement.
- **SuperMemo SM-2**: interval growth is geometric per review, which over cumulative reviews produces a log relationship between review count and retention strength — the same diminishing-returns shape.

The consistent finding across all three frameworks: the marginal benefit of the *k*-th access should decrease as *k* grows.

**Proposed change**

```python
import math

def _apply_reinforcement(self, memory: Dict[str, Any], base_rate: float) -> float:
    access_count = self._resolve_access_count(memory)
    reinforcement_factor = self._resolve_reinforcement_factor(memory)
    return base_rate * (1 + reinforcement_factor * math.log1p(access_count))
```

`math.log1p(n)` computes `ln(1 + n)`, which is exactly 0 when `access_count=0` (preserving the current behavior for unaccessed memories) and is numerically stable near zero.

**Comparison at `reinforcement_factor=0.3`, `working` tier (base S=1.5):**

| access_count | Linear multiplier | Log multiplier | Log half-life |
|---|---|---|---|
| 0 | 1.00× | 1.00× | ~1 day |
| 1 | 1.30× | 1.21× | ~1.2 days |
| 10 | 4.00× | 1.72× | ~1.7 days |
| 100 | 31.0× | 2.38× | ~2.4 days |
| 1000 | 301× | 3.07× | ~3.1 days |

The log version saturates naturally — no cap needed, and the asymptotic behavior is physically reasonable (no memory lives forever purely from access count). Happy to update the PR if this direction looks good.

jfeng18 (commented):
Nice work — this is exactly what issue #998 needed. The formula direction is correct and the code is clean.

One concern: the multiplier `(1 + rf * ac)` grows without bound. With the default `rf=0.3`:

| access_count | multiplier | working forget time |
|-------------|-----------|--------------------|
| 0 | 1.0x | 1.8 days |
| 10 | 4.0x | 7.2 days |
| 100 | 31.0x | 56 days |

Since `on_search` increments `access_count` for every returned result, a frequently-surfaced memory can accumulate hundreds of accesses quickly (e.g., 10 searches/day × 30 days × appearing in results = 3000 ac → 4.5 year forget time).

Two possible mitigations:
- Cap the multiplier: `min(1 + rf * ac, 10)` — simple, configurable
- Diminishing returns: `1 + rf * log(1 + ac)` — ac=1000 gives only 3x instead of 301x

Either would keep the reinforcement benefit for typical usage (ac=0–10) while preventing runaway lifetime at high access counts. Happy to hear your thoughts on which fits better.

wayyoungboy:
Thanks, this concern is valid. I updated the PR to use diminishing returns instead of the unbounded linear multiplier:

```text
effective_decay_rate = base_decay_rate * (1 + reinforcement_factor * log1p(access_count))
```

This preserves the existing behavior when `access_count=0`, keeps reinforcement meaningful for accessed memories, and avoids runaway lifetimes for frequently surfaced memories without adding a new cap/config field.

Re-ran the relevant checks:

```bash
python -m pytest tests/unit/intelligence/test_ebbinghaus_decay_rate.py -q
# 23 passed

python -m pytest tests/unit/intelligence -q
# 46 passed

python -m compileall -q src/powermem/intelligence/ebbinghaus_algorithm.py tests/unit/intelligence/test_ebbinghaus_decay_rate.py

git diff --check
```

lightzt99:
LGTM

PR #1010, opened 2026-06-10
fix: scope graph entities by user

### LINKED ISSUE #1002: graph_entities table lacks user_id — cross-user data leakage in graph search
## Problem

The `graph_entities` table has no `user_id` column, and `_search_node()` ignores the `filters` parameter entirely. This means:

1. **Entity dedup is global**: when user A adds an entity "Alice", user B adding "Alice" will reuse user A's entity row (via vector similarity dedup in `_search_source_node`).
2. **Search seed discovery is global**: when user B searches, `_search_node()` returns entities created by any user as seed nodes for multi-hop traversal.

Relationships are correctly scoped by `user_id`, so the traversal itself respects user boundaries. But the seed node selection does not, which can expose entity names and types across users.

## Reproduction

```python
memory.add("Alice works at Acme", user_id="user_a")
# user_b's graph search can discover entity 'alice' as a seed node
results = memory.search("Alice", user_id="user_b")
# results["relations"] may contain 'alice' entity from user_a's data
```

## Suggested fix

1. Add `user_id` column to `graph_entities` table
2. Include `user_id` in `_search_node()` WHERE clause
3. Schema migration for existing deployments (add column, backfill from relationships table)

This is a data isolation issue rather than a functional bug, so it only matters in multi-user deployments. Happy to discuss the migration strategy or submit a PR.

wayyoungboy:
## Summary

Closes #1002.

This PR scopes OceanBase graph entities by `user_id` so graph entity deduplication and seed discovery no longer reuse nodes across users.

Changes:
- adds `user_id` to `graph_entities` and indexes `(user_id, name)`
- writes `user_id` when creating graph entities
- filters graph entity vector search by `user_id`
- filters entity name lookups used by relationship deletion by `user_id`
- recreates legacy unscoped graph tables when an existing `graph_entities` table lacks `user_id`
- avoids dropping tables if schema inspection fails; startup raises instead
- adds unit coverage for schema creation, legacy migration, entity creation, search scoping, and deletion lookup scoping

## Validation

- `python -m flake8 src/powermem/storage/oceanbase/oceanbase_graph.py tests/unit/test_oceanbase_graph.py --select=F601,F821,E999`
- `git diff --check`

### DIFF
--- a/src/powermem/storage/oceanbase/oceanbase_graph.py
+++ b/src/powermem/storage/oceanbase/oceanbase_graph.py
@@ -294,11 +294,35 @@ class MemoryGraph(GraphStoreBase):
             - graph_relationships: Stores relationships between entities
         """
 
-        if not self.client.check_table_exists(constants.TABLE_ENTITIES):
+        entities_exists = self.client.check_table_exists(constants.TABLE_ENTITIES)
+        if entities_exists:
+            has_user_id = self._entities_table_has_user_id()
+            if has_user_id is None:
+                raise RuntimeError(
+                    f"Unable to inspect {constants.TABLE_ENTITIES} schema before graph migration"
+                )
+            needs_entity_migration = not has_user_id
+        else:
+            needs_entity_migration = False
+
+        if needs_entity_migration:
+            logger.warning(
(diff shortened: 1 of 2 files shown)


lightzt99:
LGTM

Issue #1003, opened 2026-06-10
Graph add() pipeline is non-atomic with check-then-create race on entities


jfeng18:
## Problem 1: Non-atomic write pipeline

`MemoryGraph.add()` runs delete then add as separate uncommitted operations:

```python
deleted_entities = self._delete_entities(to_be_deleted, filters)
added_entities = self._add_entities(to_be_added, filters, entity_type_map)
```

A crash between delete and add loses data — the old relationships are removed but new ones never created.

## Problem 2: Check-then-create race on entities

`_add_entities()` searches for an existing entity by vector similarity, then creates a new one if not found:

```python
source_node = self._search_source_node(source_embedding, filters, ...)
if source_node:
    source_id = source_node["id"]
else:
    source_id = self._create_entity(source, ...)
```

Two concurrent `add()` calls for the same entity can both find no match and both create new rows with different Snowflake IDs, producing duplicate entities.

## Suggested fixes

**For atomicity**: wrap the delete + add sequence in a single transaction.

**For the race condition**: either
- Add a UNIQUE index on `(name)` (or `(name, user_id)` if user scoping is added per #1002) and use INSERT ON DUPLICATE KEY UPDATE, or
- Use SELECT FOR UPDATE in `_search_source_node` to serialize concurrent creates for the same entity.

Both issues are low-impact in single-user / low-concurrency deployments. Happy to discuss tradeoffs.

lightzt99:
@jfeng18 Thanks for the find.

Both issues confirmed in the code.

For **atomicity**: `add()` (line 459–460) calls `_delete_entities` and `_add_entities` with no transaction wrapper. The multi-hop search method already uses `with self.engine.begin() as conn` (line 941), so the pattern is established — wrapping the delete + add sequence the same way should be straightforward.

For the **check-then-create race**: `_add_entities` does the classic search-then-insert without any lock (lines 1185–1200). Of the two options you suggest, a UNIQUE index + `INSERT ON DUPLICATE KEY UPDATE` is generally preferable over `SELECT FOR UPDATE` for OceanBase since it avoids holding a row lock during the embedding search. One coordination point: if #1002 lands first and adds `user_id` to `graph_entities`, the unique index should be `(user_id, name)` rather than just `(name)`. Worth aligning the two PRs on ordering or doing them together.

Happy to help review a PR for either fix. Are you planning to tackle these separately or in one shot alongside #1002?

jfeng18:
Thanks @lightzt99.

[...]

PR #1011, opened 2026-06-10
fix: make graph add relationship writes transactional

wayyoungboy:
## Summary

Partially addresses #1003.

This PR fixes the independent atomicity part of the graph `add()` issue by putting relationship deletes and relationship inserts in the same SQLAlchemy transaction.

Changes:
- wraps the `add()` delete+add relationship write stage in `engine.begin()`
- passes the active transaction connection into `_delete_entities()` and `_add_entities()`
- uses the transaction connection for relationship deletion, relationship existence checks, relationship insertion, and final entity-name reads
- re-raises delete failures inside the transaction path so failures roll back instead of being silently treated as zero deletes
- preserves the existing non-transaction client path for direct helper calls
- adds unit coverage for transaction connection propagation, relationship delete, and relationship insert paths

The check-then-create duplicate entity race should be handled separately after #1002 lands, because the correct unique key depends on `graph_entities.user_id` and should be `(user_id, name)`.

## Validation

- `python -m flake8 src/powermem/storage/oceanbase/oceanbase_graph.py tests/unit/test_oceanbase_graph.py --select=F601,F821,E999`
- `git diff --check`


lightzt99 (approved):
LGTM.

One conflict in `tests/unit/test_oceanbase_graph.py`. Both the old test and the new test should be kept, because they cover different paths through the same method. Please check.

PR #1012, opened 2026-06-10
fix: clean up graph search tools and results

### LINKED ISSUE #1004: Graph search: tool name mismatch, duplicate results, and missing noop tool
A few smaller issues found while reviewing the graph search system. Grouping them here since each is straightforward.

### 1. Tool name mismatch between RELATIONS_TOOL and RELATIONS_STRUCT_TOOL

`graph_tools_prompts.py` line 131 defines `"name": "establish_relationships"` but line 316 defines `"name": "establish_relations"`. When the `openai_structured` LLM provider is used, it picks the struct variant with the wrong name, causing the tool call to fail silently.

### 2. Duplicate triples in search results

`search()` calls `result_relations.extend()` for each seed entity without deduplication. Multiple seed entities traversing overlapping subgraphs produce duplicate `(source, relationship, destination)` triples in the output.

### 3. Noop tool defined but never provided to LLM

Entity extraction, relation extraction, and delete detection each provide only a single tool to the LLM — no noop option. The LLM is forced to extract something even when the text contains no entities, increasing hallucination risk.

### 4. BM25 scores discarded

Graph search computes BM25 scores for re-ranking but discards them before returning results. The caller has no way to rank graph results by relevance.

Happy to submit fixes for any of these.

wayyoungboy:
## Summary

Closes #1004.

This PR fixes the graph search/tool issues reported in #1004.

Changes:
- aligns the structured relations tool name with the regular `establish_relationships` tool name
- passes `noop` alongside entity extraction and relation extraction tools so the LLM can opt out
- deduplicates graph search triples across overlapping seed-node traversals
- includes the BM25 rerank score in returned graph search results
- adds unit coverage for the tool name, noop tools, duplicate removal, and score exposure

## Validation

- `python -m flake8 src/powermem/storage/oceanbase/oceanbase_graph.py src/powermem/prompts/graph/graph_tools_prompts.py tests/unit/test_oceanbase_graph.py --select=F601,F821,E999`
- `git diff --check`

### DIFF
--- a/src/powermem/prompts/graph/graph_tools_prompts.py
+++ b/src/powermem/prompts/graph/graph_tools_prompts.py
@@ -313,7 +313,7 @@ NOOP_STRUCT_TOOL = {
 RELATIONS_STRUCT_TOOL = {
     "type": "function",
     "function": {
-        "name": "establish_relations",
+        "name": "establish_relationships",
         "description": "Establish relationships among the entities based on the provided text.",
         "strict": True,
         "parameters": {
--- a/src/powermem/storage/oceanbase/oceanbase_graph.py
+++ b/src/powermem/storage/oceanbase/oceanbase_graph.py
@@ -503,9 +503,10 @@ class MemoryGraph(GraphStoreBase):
             if idx < len(search_output):
                 item = search_output[idx]
                 search_results.append({
-                    "source": item["source"], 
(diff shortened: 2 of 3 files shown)


lightzt99 (approved):
LGTM.

PR #1015, opened 2026-06-11
feat(api): filter listed memories by scope

### LINKED ISSUE #1014: feat(api): support metadata filters on list memories
## Problem

The Dashboard needs to show personal memories and group memories in separate views. Each view needs independent pagination over `GET /api/v1/memories`.

Today the list endpoint supports `user_id`, `agent_id`, pagination, sorting, and time range, but it does not expose metadata filters such as `scope`. Using the search endpoint as a workaround is not suitable because search is semantic retrieval, does not support offset-based pagination, and cannot reliably return all memories with `query=*`.

## Desired behavior

Allow list memories to filter by metadata fields, especially `scope`, while preserving correct `total`, `limit`, and `offset` semantics.

Example:

```bash
curl -X GET \
  "http://ip:8001/api/v1/memories?user_id=u01&agent_id=a01&scope=个人记忆&limit=100&offset=0&order=desc" \
  -H "accept: application/json"
```

The frontend should be able to request personal and group memories separately:

- `scope=个人记忆`
- `scope=群记忆`

## Acceptance criteria

- `GET /api/v1/memories` supports metadata filtering for `scope`.
- List results and `total` are calculated against the same filters.
- Pagination remains stable: filtering must happen before limit/offset.
- Existing list behavior without metadata filters remains unchanged.
- Tests cover separate pagination for personal and group memories.

wayyoungboy:
## Summary

Closes #1014.

- Add `scope` query parameter to `GET /api/v1/memories` for metadata scope filtering.
- Pass metadata filters through API, service, core memory, and storage count/list paths.
- Keep pagination stable by applying metadata filters before `limit`/`offset`.
- Document the new REST query parameter and add regression tests for personal/group-style scope pagination.

## Validation

- `pytest tests/unit/test_list_memory_filters.py tests/unit/test_memory.py -q`
- `PYTHONPATH=src pytest tests/unit -q`
- `python3 -m flake8 src/server/api/v1/memories.py src/server/services/memory_service.py src/powermem/core/memory.py src/powermem/storage/adapter.py tests/unit/test_list_memory_filters.py --select=F601,F821,E999`
- `git diff --check`

### DIFF
--- a/docs/api/0005-api_server.md
+++ b/docs/api/0005-api_server.md
@@ -540,6 +540,7 @@ curl -X POST "http://localhost:8848/api/v1/memories/batch" \
 **Query Parameters**:
 - `user_id` (optional): Filter by user ID
 - `agent_id` (optional): Filter by agent ID
+- `scope` (optional): Filter by metadata scope, such as personal or group memory
 - `limit` (optional, default: 100): Maximum number of results (1-1000)
 - `offset` (optional, default: 0): Number of results to skip
 - `sort_by` (optional): Field to sort by. Options: `created_at`, `updated_at`, `id`. If not specified, results are returned in their original order
@@ -558,6 +559,10 @@ curl -X GET "http://localhost:8848/api/v1/memories?user_id=user-123&limit=20&off
 
 # Filter by agent
 curl -X GET "http://localhost:8848/api/v1/memories?agent_id=agent-456&limit=50&offset=0" \
+  -H "X-API-Key: test-api-key-123"
+
+# Filter by metadata scope
+curl -X GET "http://localhost:8848/api/v1/memories?user_id=user-123&agent_id=agent-456&scope=personal&limit=20&offset=0" \
   -H "X-API-Key: test-api-key-123"
 
 # Sort by updated_at (descending - most recent first)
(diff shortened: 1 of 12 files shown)


lightzt99:
A small suggestion: 

it might be worth pushing the scope filter down to the DB layer rather than relying on in-memory filtering. The SQLite backend already supports `json_extract`-based filtering on nested payload paths via `_json_path_for_key` (`sqlite_vector_store.py:20`) — it just needs the adapter to translate the logical key to the backend-specific path:

- SQLite: `scope` → `metadata.scope` → `json_extract(payload, '$.metadata.scope') = 'personal'`
- OceanBase: `scope` → filter on the `metadata` JSON column directly

Both `list()` and `count()` on the vector stores already accept arbitrary filter keys with dotted nesting. If the adapter does this translation, the `has_extra_filters` / fetch-all / in-memory paginate branch could be dropped, keeping pagination and counting at the DB level and avoiding the large-fetch risk on bigger datasets.

wayyoungboy:
Thanks for the suggestion. I pushed an update that translates logical metadata filters before calling the vector store, so list/count no longer fetch all rows for scope filtering. SQLite and PGVector now use nested payload paths such as `metadata.scope`, while OceanBase keeps `scope` so its metadata JSON column filter is used directly.\n\nValidation done:\n- targeted scope/list tests passed\n- full unit suite passed: 421 passed\n- started the API server locally and verified the real `GET /api/v1/memories` flow: created 4 memories with personal/group scopes, confirmed `scope=personal` returned total 3 with correct paginated results, confirmed `scope=group` returned total 1, then cleaned up the 4 test memories

lightzt99:
LGTM

PR #1016, opened 2026-06-11
fix(init): include server,seekdb extras and improve LLM auto-detection

lightzt99:
## Summary

- **`init.sh`**: Default `POWERMEM_INIT_PACKAGE` to `powermem[server,seekdb]` so `pyseekdb` is installed on fresh plugin init, fixing `ModuleNotFoundError: No module named 'pyseekdb'` at server startup
- **`init.sh`**: Detect LLM provider from available API key rather than model name prefix — prevents provider/key mismatch when `settings.json` has e.g. `model: "qwen/..."` but the only available key is `ANTHROPIC_API_KEY`; key priority order: `ANTHROPIC_AUTH_TOKEN` → `ANTHROPIC_API_KEY` → `OPENAI_API_KEY` → `DEEPSEEK_API_KEY` → `DASHSCOPE_API_KEY`
- **`init.sh`**: Add `validate_llm_config` step after `.env` creation — makes a minimal test API call and fails early with a clear error if provider/model/key are incompatible, instead of silently starting a broken server
- **`SETUP.md`**: Fix non-CN model preload — add `pip install huggingface_hub` before `snapshot_download` call (mirrors the CN path which installs `modelscope` first)
- **`SETUP.md`**: Fix dashboard build ordering — move the background build prompt to immediately after model download is started in background, so dashboard build runs in parallel with model download and server startup

### DIFF
--- a/apps/claude-code-plugin/SETUP.md
+++ b/apps/claude-code-plugin/SETUP.md
@@ -327,7 +327,7 @@ writing. Never silently patch `.env`.**
         " >> /tmp/powermem-model-download.log 2>&1 &
 
       **Non-CN region (HuggingFace path):**
-        $POWERMEM_PYTHON -c "
+        $POWERMEM_PYTHON -m pip install -q huggingface_hub && $POWERMEM_PYTHON -c "
         from huggingface_hub import snapshot_download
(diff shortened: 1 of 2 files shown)


wayyoungboy:
LGTM

PR #1019, opened 2026-06-12
feat(init): change default data dir to ~/.powermem

lightzt99:
## Summary
- Remove `CLAUDE_PLUGIN_DATA` dependency from `common.sh`, `run-hook.sh`
- Hardcode data directory to `$HOME/.powermem`
- Add `~/.powermem/.env` to `settings.py` auto-discovery candidates
- Fix model prefix stripping bug in `init.sh` (prefix not stripped when provider detected from key)
- Update `SETUP.md` documented default path
- Add CLAUDE_PLUGIN_ROOT discovery guide to `SETUP.md` to prevent bash variable scoping errors when running plugin scripts

## Test plan
- [x] `common.sh` sources correctly with `DATA_DIR=$HOME/.powermem`
- [x] `run-hook.sh` reads `runtime.env` from `$HOME/.powermem/runtime.env`
- [x] `settings.py` discovers `.env` from `~/.powermem/.env`
- [x] `.env` created at `~/.powermem/.env` by init.sh
- [x] init.sh → LLM validation → pip install with extras → server startup healthy
- [x] Model prefix stripping: `anthropic/claude-sonnet-4-6` correctly becomes model=`claude-sonnet-4-6` with provider detected from key
- [x] CLAUDE_PLUGIN_ROOT discovery: two-step invocation avoids empty `$VAR` expansion on single-line `VAR=val sh "$VAR/..."`

### DIFF
--- a/apps/claude-code-plugin/SETUP.md
+++ b/apps/claude-code-plugin/SETUP.md
@@ -37,7 +37,7 @@ PyPI package.
 Installed-plugin init is idempotent and uses plugin-local state:
 
 ```text
-${CLAUDE_PLUGIN_DATA:-$HOME/.claude/plugins/data/memory-powermem-powermem}/
+$HOME/.powermem/
   .env
   runtime.env
   server.pid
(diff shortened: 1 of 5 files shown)


wayyoungboy:
LGTM

PR #1021, opened 2026-06-12
Fix regression auth restart CI

wayyoungboy:
## Summary
- run regression CI API server with a single worker to avoid multi-worker startup contention during auth-mode restarts
- dump server.log when regression tests fail so future startup failures expose the server-side error

## Verification
- source /data/code/wayyoung/powermem/.venv/bin/activate
- locally verified make server-start -> health OK with auth disabled, then auth enabled restart -> health OK
- git diff --check
- parsed .github/workflows/regression.yml with PyYAML

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -8,6 +8,7 @@ on:
       - 'tests/**'
       - 'pyproject.toml'
       - 'Makefile'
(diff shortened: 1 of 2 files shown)


wayyoungboy:
Closing this PR per maintainer request.

PR #1030, opened 2026-06-15
[uv] Use uvx for Claude plugin backend init

wayyoungboy:
## Summary

- Switch Claude Code plugin marketplace init from plugin-local venv installs to an uvx-style backend launcher using `uv tool run --from powermem[server,seekdb] powermem-server`.
- Keep the default storage path on embedded OceanBase/seekdb, add CN/non-CN uv installation guidance, and use TUNA as the CN default Python package index for uvx resolution.
- Update preload/status/reset/docs/SVG flowcharts and add unit coverage for uv install, CN mirror behavior, uvx invocation, and the no-venv init path.
- Synced the updated uvx section and SVG flowchart to Yuque: https://yuque.antfin.com/go/doc/528703254

## Validation

- `sh -n apps/claude-code-plugin/scripts/common.sh`
- `sh -n apps/claude-code-plugin/scripts/init.sh`
- `sh -n apps/claude-code-plugin/scripts/preload-model.sh`
- `sh -n apps/claude-code-plugin/scripts/status.sh`
- `pytest -q tests/unit/test_claude_plugin_uv_install.py`
- XML parse for `apps/claude-code-plugin/init-flow.svg` and `apps/claude-code-plugin/uv-init-flow.svg`
- `git diff --check -- apps/claude-code-plugin docs/integrations/claude_code.md tests/unit/test_claude_plugin_uv_install.py`
- `claude plugin validate apps/claude-code-plugin`
- `claude plugin validate .`

### DIFF
--- a/apps/claude-code-plugin/README.md
+++ b/apps/claude-code-plugin/README.md
@@ -30,25 +30,30 @@ installed skills such as `/memory-powermem:init`, `/memory-powermem:status`,
 `/memory-powermem:stop`, and `/memory-powermem:reset`.
 
 The marketplace install only installs the Claude Code plugin connector. The
-`/memory-powermem:init` step prepares the backend by creating a plugin-local venv
-and installing `powermem` from PyPI.
+`/memory-powermem:init` step prepares the backend by ensuring `uv`, then starts
+PowerMem with the uvx-style launcher
+`uvx --from 'powermem[server,seekdb]' powermem-server`. If `uv` is missing, init
+installs it automatically: non-CN networks use the official Astral installer,
+while CN networks use the USTC mirror at
+`https://mirrors.ustc.edu.cn/github-release/astral-sh/uv/LatestRelease/`.
+CN package resolution uses `--default-index https://pypi.tuna.tsinghua.edu.cn/simple`.
 
 The PyPI package used by init must include the backend features and dependencies
 required by the plugin, including the default local embedding path
 (`sentence-transformers` / `all-MiniLM-L6-v2`). If you are testing plugin changes
-that depend on unpublished backend code, run init with `POWERMEM_INIT_PACKAGE`
-instead of the skill command:
+that depend on unpublished backend code, set `POWERMEM_INIT_PACKAGE` to a Git URL;
+init passes it to `uvx --from`:
 
 ```bash
-POWERMEM_INIT_PACKAGE='powermem @ git+https://github.com/oceanbase/powermem.git@<branch-or-sha>' \
+POWERMEM_INIT_PACKAGE='powermem[server,seekdb] @ git+https://github.com/oceanbase/powermem.git@<branch-or-sha>' \
   sh "$CLAUDE_PLUGIN_ROOT/scripts/init.sh"
 ```
 
-For marketplace branch testing before merge, add the marketplace from the same
-branch:
+For marketplace branch testing before merge, you can also add the marketplace
+from the same branch:
 
 ```text
-/plugin marketplace add owner/powermem@<branch>
+/plugin marketplace add https://github.com/owner/powermem.git#<branch>
 /plugin install memory-powermem@powermem
 /reload-plugins
 ```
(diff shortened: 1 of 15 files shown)


lightzt99:
The PR switches from pip+venv to uvx, but init still requires system Python ≥ 3.11 before uv is even installed. This is mostly unnecessary given the actual requirements.

1. **`choose_python()` runs before `ensure_uv`**, checking `sys.version_info >= (3, 11)`. If no such Python exists, init exits hard — even though uvx could handle Python itself.

2. **The two embedded Python scripts** (`create_env_file`, `validate_llm_config`) only need **Python 3.6** — they use f-strings, `pathlib`, `urllib.request`, and `json`. No walrus operator, no match/case, no 3.11 features. The 3.11 check is a leftover from the venv era where `$BOOTSTRAP_PYTHON` was also the venv base interpreter.

3. **`ensure_uv` has no Python dependency.** It only needs `curl`/`wget` + `sh`. CN detection (`detect_public_ip_country`) has `|| true` fallback — if Python is unavailable it defaults to the official Astral installer.

The fix is a simple reorder:

```bash
ensure_uv                              # curl/wget only → zero Python dependency
uv python install 3.11                 # uv downloads standalone Python build
BOOTSTRAP_PYTHON=$(uv python find 3.11)
create_env_file / validate_llm_config  # run with uv-managed Python
uvx --python 3.11 --from "$PACKAGE" powermem-server
```

This eliminates the system Python 3.11 requirement entirely — only `curl` + `sh` are needed.

wayyoungboy:
Thanks, this is a good catch. I updated the init flow so the default path no longer requires a system Python >= 3.11.

The plugin now ensures `uv` first, installs/finds Python 3.11 through `uv`, and then uses that uv-managed interpreter for env generation, LLM validation, and the backend launch:

```sh
uv tool run --python "$BOOTSTRAP_PYTHON" --from "$PACKAGE" powermem-server ...
```

This is covered by the follow-up commits, especially `a7c1426` (`Include Claude plugin uv bootstrap helpers`). I also verified the updated flow in a fresh Ubuntu 20.04 container with no host mounts and no host networking: Claude Code installed via nvm/npm, the plugin installed from the marketplace branch, `/memory-powermem:init` succeeded, and health/add/search/hook recall all passed.

lightzt99:
Two additional issues in the updated code:

**1. `uv python install 3.11` has no CN mirror (`common.sh`)**

`ensure_uv` uses USTC for the uv binary; `configure_uv_index` uses TUNA for PyPI packages. But `ensure_bootstrap_python()` runs:

```sh
"$UV_BIN" python install 3.11
```

This downloads `python-build-standalone` tarballs from GitHub Releases — no mirror is configured, so CN users will hit timeouts or bandwidth caps. `UV_PYTHON_INSTALL_MIRROR` (uv ≥ 0.5.0) can redirect these downloads, but no stable public CN mirror for `python-build-standalone` exists yet.

Recommended workaround: document that CN users can set `POWERMEM_INIT_PYTHON=python3.11` (pointing to a pre-installed interpreter via pyenv/deadsnakes) to skip `uv python install` entirely.

**2. PID file holds the `uv` wrapper PID, not the server PID (`init.sh`)**

```sh
nohup "$UV_BIN" tool run ... powermem-server ... &
pid=$!
write_managed_pid "$pid"   # $! is the uv parent, not powermem-server
```

`uv tool run` uses fork, not exec. `pid_is_powermem_server()` still works correctly — the uv parent's args contain `powermem-server` so the grep matches ✓.

The edge case: if uv is killed with SIGKILL (e.g. OOM killer), the powermem-server child orphans on port 8848. The stale `.pid` points to the dead uv process, so `pid_alive()` returns false and the PID file is removed. The orphaned server is still healthy, so `is_healthy()` detects it and init exits 0 as "External PowerMem backend is healthy" — **init itself is safe**. However `stop.sh` loses its handle: the `.pid` file is gone, so `kill $(cat powermem.pid)` no longer works.

Low-probability in coding-agent use, but worth adding one line to SETUP.md: *"If `/memory-powermem:stop` cannot reach the server, run `kill $(lsof -ti :8848)` to stop it manually."*

wayyoungboy:
For the PID/orphan server edge case, I tightened `/memory-powermem:stop` in `6f43789`.

`stop.sh` still stops the recorded managed PID first, but it now also resolves the local runtime port from `runtime.env` and scans listening processes on that port via `lsof`/`ss`/`fuser`. It only stops candidates whose process args explicitly contain `powermem-server` and whose env matches the plugin env file when `/proc/<pid>/environ` is readable, so it should avoid killing unrelated services on the same port.

Covered by unit tests:
- orphaned `powermem-server` on the runtime port is stopped
- non-PowerMem listener on the runtime port is ignored

Ran:
- `pytest -q tests/unit/test_claude_plugin_uv_install.py`
- `sh -n` for the related Claude plugin scripts

This should address the uv-wrapper PID issue without requiring users to manually run `kill $(lsof -ti :8848)` in the normal stop flow.

PR #1031, opened 2026-06-15
feat: sqlite default to coding agent

### LINKED ISSUE #1029: Use SQLite as the default database for code agent integrations
## Requirement

Use SQLite as the default database for code agent integrations, especially:

- Claude Code
- Codex CLI

## Background

Code agent integrations are usually installed and run on a developer's local machine. For these scenarios, the default backend should be lightweight, local-first, and easy to bootstrap without requiring a separate database service or embedded OceanBase/seekdb setup.

SQLite is a better default for this path because it is simpler to initialize, easier to inspect, and more compatible across developer machines.

### Claude Code

When the `memory-powermem` Claude Code integration initializes PowerMem, it should default to SQLite storage unless the user explicitly configures another database provider.

Expected generated config shape:

```env
DATABASE_PROVIDER=sqlite
SQLITE_PATH=~/.powermem/sqlite_data/powermem.db
SQLITE_COLLECTION=memories
SQLITE_ENABLE_WAL=true
SQLITE_TIMEOUT=30
SQLITE_VECTOR_BACKEND=auto
```

### Codex CLI

When configuring `memory-powermem` for Codex CLI through MCP, the local PowerMem backend should also default to SQLite storage unless the user explicitly chooses another database provider.

The Codex CLI MCP config should only point to PowerMem, for example:

```json
{
  "mcpServers": {
    "powermem": {
      "url": "http://localhost:8848/mcp"
    }
  }
}
```

The underlying PowerMem `.env` should use SQLite by default.

## Expected behavior

- Code agent setup flows do not require OceanBase/seekdb by default.
- `powermem-server` and `powermem-mcp` can start with local SQLite storage.
- Existing user-provided database settings are preserved and not overwritten.
- Users can still opt into OceanBase or other storage providers explicitly.
- The default local embedding path remains compatible with SQLite.

## Documentation updates

Update relevant setup docs so they describe SQLite as the default for code agent scenarios:

- `apps/claude-code-plugin/SETUP.md`
- `apps/claude-code-plugin/README.md`
- `apps/mcp-client/SETUP.md`
- `docs/integrations/claude_code.md`
- `docs/integrations/codex.md`
- `docs/integrations/mcp_client.md`

## Acceptance criteria

- Claude Code init writes SQLite database defaults when no database provider is explicitly configured.
- Codex CLI MCP setup documentation uses SQLite as the default PowerMem backend.
- `powermem-mcp` install guidance includes the SQLite dependencies needed for the default local backend.
- Basic memory operations work with the SQLite default:
  - add
  - search
  - update
  - delete
  - list / get
- Re-running setup is idempotent and does not overwrite existing user-selected database providers.
- Tests cover SQLite config loading and SQLite vector backend behavior.

lightzt99:
## Summary

For coding agent scenarios (users running Claude Code locally on a single machine), OceanBase/seekdb is too heavy. The current `init.sh` unconditionally installs `powermem[server,seekdb]` and writes `DATABASE_PROVIDER=oceanbase`. This PR makes init default to SQLite and fixes known issues in the SQLite storage layer.

### Key changes

**`init.sh`**
- Reads `POWERMEM_INIT_DATABASE_PROVIDER` (defaults to `sqlite`); invalid values warn + fall back to sqlite
- PACKAGE is now dynamic: SQLite → `powermem[server,extras]` (sentence-transformers), OceanBase → `powermem[server,seekdb]`
- SQLite path writes `EMBEDDING_PROVIDER=huggingface` in `.env` — uses `sentence-transformers` locally, **no API key required**; default model `all-MiniLM-L6-v2` (same as `preload-model.sh` cache)
- `HF_HUB_OFFLINE=1` is written to `.env` for huggingface provider to prevent startup hang when `huggingface.co` is unreachable
- OceanBase path keeps `EMBEDDING_PROVIDER=default` unchanged
- Warning printed when `EMBEDDING_PROVIDER=huggingface` and `POWERMEM_INIT_PRELOAD_MODEL` is not set

**`sqlite_vector_store.py`**
- Fix `enable_wal` / `timeout` silently dropped via `**kwargs`
- Apply `PRAGMA journal_mode=WAL` (needed when hooks run concurrently)
- Add pysqlite3 fallback import
- Replace bare version check (≥ 3.9.0) with `_check_sqlite_features()` — probes FTS5 and JSON1 at connection time with a functional smoke test

### Idempotency / existing users

Once `.env` exists, `create_env_file()` is never called (existing guard). Existing OceanBase users who re-run init are completely unaffected.

### Embedding provider selection

| db_provider | Default EMBEDDING_PROVIDER | Deps installed | API key needed |
|---|---|---|---|
| sqlite | huggingface | powermem[server,extras] | No |
| oceanbase | default | powermem[server,seekdb] | No (local ONNX) |
| any | User-set via `POWERMEM_INIT_EMBEDDING_PROVIDER` | — | Depends on provider |

### Cross-machine behavior

| Scenario | Behavior |
|---|---|
| New user, first init | Writes `DATABASE_PROVIDER=sqlite`, `EMBEDDING_PROVIDER=huggingface`, installs `powermem[server,extras]` |
| Re-run init with existing `.env` | Preserves `.env`; reads `DATABASE_PROVIDER` from it for PACKAGE selection |
| Production / cluster users | Set `POWERMEM_INIT_DATABASE_PROVIDER=oceanbase` — original behavior unchanged |

### SQLite feature limitations (documented)

The following features are silently disabled under SQLite by existing code — this PR introduces no new restrictions: sub-stores, source store, skill store, graph store, sparse vectors, native hybrid search, data migration. All core coding agent operations (add, search, update, delete, Ebbinghaus decay) are fully supported.

## Testing

```
apps/claude-code-plugin/tests/test_env_generation.py  9/9 passed
  — db_provider branching, invalid fallback, case normalization, embedding override

tests/unit/storage/test_sqlite_init.py  10/10 passed
  — FTS5/JSON1 availability, WAL on disk / skipped for :memory: / disabled, timeout, CRUD
```

## Related

- Closes #1029


wayyoungboy:
Thanks for the update. I found two issues that should be addressed before merging:

1. **Default SQLite init installs too few dependencies.**
   In `apps/claude-code-plugin/scripts/init.sh`, the SQLite path now installs `powermem[server]`, but the generated `.env` still defaults to `EMBEDDING_PROVIDER=default`. That default embedder imports `pyseekdb` in `src/powermem/integrations/embeddings/pyseekdb_default.py`, while `pyseekdb` is only pulled by the `seekdb` extra, not by `server`. In a clean environment, the default init path can install successfully but then fail when `Memory` initializes or the first memory API is used. I think the SQLite default path should either keep installing the dependency needed by the default embedder, or switch the generated default embedder to one whose dependencies are included by `powermem[server]`.

2. **SQLite fallback should test required features, not only the version.**
   The init script and `SQLiteVectorStore` check `sqlite_version_info >= (3, 9, 0)`, but the store always creates an FTS5 virtual table and also relies on JSON1 behavior. FTS5/JSON1 availability depends on how SQLite was compiled, so a version check alone can pass while `CREATE VIRTUAL TABLE ... USING fts5` fails later. It would be safer to run a small smoke test for `fts5` and `json_extract` during init, and install/use `pysqlite3-binary` when the smoke test fails.

I also confirmed that `SQLITE_PATH`, `SQLITE_ENABLE_WAL`, and `SQLITE_TIMEOUT` are read by the config loader as expected. The PR description mentions `test_env_generation` / `test_sqlite_store`, but I did not see those tests submitted in this PR, so adding coverage for the default init dependency path would help prevent this regression.

wayyoungboy (commented):
Code review — read the full diff.

## Overall

The init.sh branching logic is clean, WAL fix is correct, and the pysqlite3 fallback import is standard. The idempotent .env guard works as expected.

## Issues

### 1. Default embedder will ImportError on clean SQLite installs

The SQLite init path installs `powermem[server]`, but the generated `.env` does not set `EMBEDDING_PROVIDER` — so it defaults to `"default"`, which imports `pyseekdb` (only available in the `[seekdb]` extra). On a clean SQLite install, the first `memory.add()` or server startup will fail with an ImportError.

Fix options:
- Add `EMBEDDING_PROVIDER=huggingface` (or another provider bundled in `[server]`) to the SQLite `.env` block, or
- Install the embedding dependency needed by the default provider in the SQLite pip path

### 2. Version check alone doesn't guarantee FTS5/JSON1

`sqlite_vector_store.py` checks `sqlite_version_info >= (3, 9, 0)`, but FTS5/JSON1 are compile-time options — some minimal/embedded builds (Alpine musl, certain docker images) may ship 3.9+ without FTS5. The init script has the same gap.

Suggestion: add a smoke test in `__init__` or init.sh:
```python
try:
    conn.execute("CREATE VIRTUAL TABLE _fts5_probe USING fts5(x)")
    conn.execute("DROP TABLE _fts5_probe")
except Exception:
    raise RuntimeError("SQLite FTS5 not available; install pysqlite3-binary")
```

### 3. Tests mentioned in PR body are not in the diff

PR description says `test_env_generation: 6/6 passed` and `test_sqlite_store: 11/11 passed`, but no test files appear in the changed files list. Were they meant to be included?

## Minor

- The `enable_wal` / `timeout` parameter promotion from `**kwargs` to explicit args is a good fix (they were silently dropped before).

lightzt99:
Thanks for the detailed review. Both issues have been addressed in the latest commits.

**Issue 1 — dependency mismatch on default embedder**

The SQLite path now generates `EMBEDDING_PROVIDER=huggingface` instead of `default`. `HuggingFaceEmbedding` uses `sentence-transformers` (no API key, local inference), which is included in the new `powermem[server,extras]` package installed on the SQLite path. The `default` embedder (`pyseekdb`) is only installed on the OceanBase path via `powermem[server,seekdb]` as before.

To prevent startup hangs when `huggingface.co` is unreachable, `HF_HUB_OFFLINE=1` is also written to `.env` for the huggingface provider. The default embedding model is `all-MiniLM-L6-v2`, matching what `preload-model.sh` downloads.

**Issue 2 — version check insufficient**

The `sqlite_version_info >= (3, 9, 0)` check has been replaced by `_check_sqlite_features(conn)` in `SQLiteVectorStore.__init__`. It runs an actual `CREATE VIRTUAL TABLE ... USING fts5` probe and a `json_extract('{}', '$')` call at connection time, and raises a descriptive `RuntimeError` if either fails. This catches compiled-out extensions regardless of version number.

**Tests**

Both test files are now included in the PR:
- `apps/claude-code-plugin/tests/test_env_generation.py` — 9 cases covering db_provider branching, invalid fallback, case normalization, and user embedding override
- `tests/unit/storage/test_sqlite_init.py` — 10 cases covering FTS5/JSON1 availability, WAL behavior (disk/memory/disabled), timeout parameter, and basic CRUD

wayyoungboy:
Thanks for addressing the earlier items. I think one blocker remains in the default SQLite path.

`apps/claude-code-plugin/scripts/init.sh` now writes `EMBEDDING_PROVIDER=huggingface` and `HF_HUB_OFFLINE=1` for the SQLite path, but when `POWERMEM_INIT_PRELOAD_MODEL` is not set it only prints a warning and still starts the server. On a clean machine/cache, `HuggingFaceEmbedding` constructs `SentenceTransformer(self.config.model)` during embedder initialization, so offline mode has no cached model to load. That leaves the default init path able to install/start but fail as soon as the embedder is initialized or used, which is the same clean-install class of issue.

Please either make the default SQLite path preload the model before starting when `HF_HUB_OFFLINE=1`, or only set offline mode when the model is known to be cached/preloaded.

Validation I ran:
- Using the project virtual environment, ran `python -m pytest apps/claude-code-plugin/tests/test_env_generation.py tests/unit/storage/test_sqlite_init.py tests/unit/test_claude_plugin_uv_install.py tests/unit/test_pyseekdb_default_embeddings.py`.

lightzt99:
@wayyoungboy The remaining clean-install blocker is addressed in three follow-up commits on `feat/sqlite-default-coding-agent` (`bb7f2fb`, `8a12486`, `3292757`).

### Problem

`init.sh` writes `HF_HUB_OFFLINE=1` for the huggingface path but, when `POWERMEM_INIT_PRELOAD_MODEL` was unset, only printed a warning. On a clean install with no cached model, `SentenceTransformer(...)` would fail at startup.

### Fix

`init.sh` now treats `POWERMEM_INIT_PRELOAD_MODEL` as a three-state switch:

- `0|false|no|off` → skip
- `1|true|yes|on` → force preload
- **unset** → infer from `EMBEDDING_PROVIDER`: preload when `huggingface`, skip otherwise

So the default SQLite path now auto-runs `preload-model.sh` (ModelScope → HF cache) before starting the server, putting `all-MiniLM-L6-v2` on disk before the embedder is constructed. Users who already have the model cached can opt out with `POWERMEM_INIT_PRELOAD_MODEL=0`.

For the remaining opt-out combination (`PRELOAD_MODEL=0` + `HF_HUB_OFFLINE` unset), `init.sh` detects country via the existing `detect_public_ip_country()` helper and, when CN + `huggingface` + user has not set `HF_ENDPOINT`, writes `HF_ENDPOINT=https://hf-mirror.com` into `.env`. `HF_ENDPOINT` is the `huggingface_hub` SDK's official mirror switch — zero code changes, no extra runtime dependency.

### Known limitation

Verified from a CN host: `hf-mirror.com` serves small files (config.json, tokenizer) correctly, but `model.safetensors` (90MB) redirects to `cas-bridge.xethub.hf.co` which is partially blocked in CN — downloads start but stall mid-file. So the mirror injection is best-effort, not a complete fix. The `init.sh` skip-preload warning and `README.md` troubleshooting section now make this explicit: CN users on the opt-out path may still fail at runtime, and the recommended path remains the default preload (or manual `preload-model.sh`).

The only remaining failure mode is `HF_HUB_OFFLINE=1` + `PRELOAD_MODEL=0` — an explicit user opt-out. There's no way to satisfy "offline + no cache + no preload" without downloading the model somewhere.

### Validation

- **init-preload** docker tests (5 scenarios): three-state switch + no-network failure — all pass
- **full-integration** docker test (8 checks): real `powermem-server` with SQLite + HuggingFace + `HF_HUB_OFFLINE=1` + pre-cached model, runs add → get → search → get_all → delete end-to-end — 8/8 pass
- **env generation unit tests** (14 cases): 5 new cases for `HF_ENDPOINT` injection — CN detection, non-CN, user override respected, non-huggingface provider, country-undetectable. 14/14 pass

wayyoungboy:
Thanks for the follow-up. I re-read the latest diff and I think there are still two things to fix before merging:

1. `POWERMEM_INIT_PRELOAD_MODEL=0` still cannot use the runtime-download / mirror path described in the comment and docs. `create_env_file()` unconditionally writes `HF_HUB_OFFLINE=1` whenever `EMBEDDING_PROVIDER=huggingface`, and `init.sh` exports the generated `.env` before starting `powermem-server`. So in the explicit opt-out case, `SentenceTransformer(...)` is still forced offline and `HF_ENDPOINT` cannot help. Please either only write `HF_HUB_OFFLINE=1` when preload is enabled or the model cache is known to exist, or make the opt-out contract explicit as "cache must already exist" and remove the runtime-download / mirror fallback messaging for that path.

2. The plugin README still describes the init launcher / branch-testing example as `powermem[server,seekdb]`, while the new SQLite default path uses `powermem[server,extras]` so that `sentence-transformers` is installed. Following the current docs can make testers miss the dependency needed by the new default `huggingface` embedder. Please update the default SQLite instructions/examples to use `[server,extras]`, and keep `[server,seekdb]` only for the OceanBase path.

Also, GitHub currently reports this PR as conflicting with `main`; the conflict is in `apps/claude-code-plugin/scripts/init.sh`, so the branch needs a rebase/merge resolution before it can be merged.

lightzt99:
@wayyoungboy Thanks for the re-review. All three items are addressed on `feat/sqlite-default-coding-agent` (force-pushed, now rebased on `main` with no conflicts):

**1. `HF_HUB_OFFLINE=1` unconditional write**

Removed entirely (`7a15e4f`). The line was incompatible with upstream #1057, which moved the embedding model download inside `PowerMem.HuggingFaceEmbedding`: cache hit → `SentenceTransformer(local_files_only=True)` (no network); cache miss + CN → ModelScope via `uvx` + bridge into HF cache; cache miss + non-CN → HF download with a 30s timeout. Writing `HF_HUB_OFFLINE=1` globally would block that internal download for non-CN users with an empty cache. `init.sh` no longer touches `HF_HUB_OFFLINE` or `HF_ENDPOINT` — PowerMem manages offline behaviour itself.

`POWERMEM_INIT_PRELOAD_MODEL` is now a no-op: `init.sh` no longer calls `preload-model.sh` (the download lives in PowerMem internals). If the variable is set, `init.sh` prints a deprecation message pointing to the new path (`7a15e4f`, docs cleaned up in `ec845d6`).

**2. README `[server,seekdb]` → `[server,extras]`**

Fixed (`a40c590`, `ec845d6`). README, SETUP, and `skills/init/SKILL.md` now use `powermem[server,extras]` for the SQLite default path (pulls `sentence-transformers` for the `huggingface` embedder) and keep `powermem[server,seekdb]` only for the OceanBase path. The troubleshooting example also got the dual-spec treatment.

**3. Rebase conflict on `init.sh`**

Resolved. Branch is now 8 commits on top of `main`, no conflicts.

**Extra fix discovered while re-testing**

`test_init_uses_uvx_launcher_instead_of_plugin_venv_install` was failing because my first commit (`3034c95`) carried an obsolete venv-install block that referenced `venv_powermem_server` — a function deleted by upstream PR #1030. After rebase, `$(venv_powermem_server)` expanded to empty, making the block always run but completely dead (uvx uses an ephemeral env, not that venv). The pysqlite3-binary auto-install from `eb27987` also landed in the dead venv. Fixed in `8a314fe`: the venv block is gone, and pysqlite3-binary is now injected via `uvx tool run --with pysqlite3-binary` when the bootstrap Python reports SQLite < 3.9.0.

**Regression guards**

Three new static-text assertions on `init.sh` (`9123d6f`):
- `test_init_does_not_write_hf_hub_offline_to_env` — any `HF_HUB_OFFLINE=1` reference must be inside a comment
- `test_init_package_spec_matches_db_provider` — SQLite → `[server,extras]`, OceanBase → `[server,seekdb]`
- `test_init_preload_model_is_deprecated_no_op` — `preload-model.sh` must not be called; only a deprecation message is allowed

**Tests**

- `apps/claude-code-plugin/tests/test_env_generation.py` — 9/9
- `tests/unit/storage/test_sqlite_init.py` — 10/10
- `tests/unit/test_claude_plugin_uv_install.py` — 19/19 (was 16/17 before the dead-venv fix)

Commit list on the branch:
```
8a314fe fix(init): remove dead venv-install block, rework pysqlite3 injection
9123d6f test(plugin): add regression guards for PR #1031 review points
ec845d6 docs(plugin): drop POWERMEM_INIT_PRELOAD_MODEL from init docs
7a15e4f fix(init): drop HF_HUB_OFFLINE=1 and obsolete preload switch
a40c590 docs(plugin): reflect [server,extras] default for SQLite path
5fff8b8 fix(sqlite): use huggingface embedder by default; add FTS5/JSON1 smoke test
eb27987 feat(init): auto-install pysqlite3-binary when system SQLite < 3.9.0
3034c95 feat(init): default to SQLite for coding agent, fix SQLiteVectorStore WAL bug
```

lightzt99:
@wayyoungboy Hi, I have verified all changes on this branch (`feat/sqlite-default-coding-agent`) using the Docker test environment at `~/test/powermem` with SQLite backend. Results: **162/162 passed, 0 failures**.

| Test layer | Count | Status |
|---|---|---|
| Docker integration (run.sh) | 9 | 9 PASS |
| Docker full-stack (HuggingFace + SQLite + ollama) | 8 | 8 PASS |
| SQLite init unit tests (`test_sqlite_init.py`) | 10 | 10 PASS |
| Claude Code plugin install (`test_claude_plugin_uv_install.py`) | 19 | 19 PASS |
| Embedding model download (`test_pyseekdb_default_download.py`) | 24 | 24 PASS |
| Go hook tests (scrubber + worker payload) | 47 | 47 PASS |
| Hook privacy E2E regression (`test_claude_hook_privacy_e2e.py`) | 6 | 6 PASS |
| OceanBase graph (`test_oceanbase_graph.py`) | 39 | 39 PASS |
| **Total** | **162** | **162 PASS** |

**Environment**: Docker compose with CentOS 7 host + Debian slim container, Python 3.11.13, SQLite with WAL enabled, HuggingFace `all-MiniLM-L6-v2` embedder, ollama for LLM fallback.

**Coverage**: SQLite CRUD + vector search + WAL mode, Claude Code plugin full lifecycle (uvx install → init config → hook event handling), hook privacy scrubbing (7 dimensions + E2E regression), embed model CN/non-CN download routing, OceanBase graph no regressions.

PR #1032, opened 2026-06-15
fix(graph): add UNIQUE index and upsert for entity dedup (#1003)

jfeng18:
## Summary

Addresses the race condition part of #1003.

The `graph_entities` table had no uniqueness constraint on `(user_id, name)`. Concurrent `add()` calls for the same entity could both pass the vector similarity check and both insert, producing duplicate entity rows.

## Changes

- Upgrade `idx_user_name` to `UNIQUE` index
- Rewrite `_create_entity()` to use `INSERT ... ON DUPLICATE KEY UPDATE` (raw SQL) instead of `self.client.upsert()` (which is `REPLACE INTO` and would delete the old row, breaking relationship foreign key references)
- After insert, `SELECT` to return the actual id (preserved on duplicate)
- Update `test_create_entity` to match new raw SQL path

## What's NOT in this PR

The transaction wrapper for `add()` delete+add atomicity is deferred — `pyobvector` does not support external connection injection, so wrapping `self.client.*` calls in a single transaction requires a larger refactor (raw SQL rewrite of the full mutation path).

## Test plan

- [x] `test_create_entity` updated for raw SQL mock path
- [x] Verified INSERT ON DUPLICATE KEY UPDATE preserves existing row id
- [x] SQL injection safe (all values parameterized)
- [x] Migration: existing drop+recreate logic (from PR #1010) picks up UNIQUE constraint
- [ ] CI tests pass

### DIFF
--- a/src/powermem/storage/oceanbase/oceanbase_graph.py
+++ b/src/powermem/storage/oceanbase/oceanbase_graph.py
@@ -3,13 +3,23 @@ OceanBase graph storage implementation
 
 This module provides OceanBase-based graph storage for memory data.
 """
+
 import json
 import logging
 import warnings
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import numpy as np
-from sqlalchemy import bindparam, text, MetaData, Column, String, Index, Table, BigInteger
+from sqlalchemy import (
+    BigInteger,
+    Column,
+    Index,
+    MetaData,
+    String,
+    Table,
+    bindparam,
+    text,
+)
 from sqlalchemy.dialects.mysql import TIMESTAMP
 from sqlalchemy.exc import SAWarning
 
@@ -20,7 +30,7 @@ warnings.filterwarnings(
     "ignore",
     message="Unknown schema content",
     category=SAWarning,
-    module="pyobvector.*"
+    module="pyobvector.*",
 )
 
 # Suppress pkg_resources deprecation warning from jieba internal usage
@@ -29,23 +39,29 @@ warnings.filterwarnings(
     "ignore",
     message="pkg_resources is deprecated as an API",
     category=UserWarning,
-    module="jieba.*"
+    module="jieba.*",
 )
 
-from pyobvector import ObVecClient, l2_distance, VECTOR, VecIndexType
+from pyobvector import VECTOR, ObVecClient, VecIndexType, l2_distance
 
 from powermem.integrations import EmbedderFactory, LLMFactory
 from powermem.storage.base import GraphStoreBase
-from powermem.utils.utils import format_entities, remove_code_blocks, generate_snowflake_id, get_current_datetime
 from powermem.utils.oceanbase_util import OceanBaseUtil
+from powermem.utils.utils import (
+    format_entities,
+    generate_snowflake_id,
+    get_current_datetime,
+    remove_code_blocks,
+)
 
 try:
     from rank_bm25 import BM25Okapi
 except ImportError:
-    raise ImportError("rank_bm25 is not installed. Please install it using pip install rank-bm25")
+    raise ImportError(
+        "rank_bm25 is not installed. Please install it using pip install rank-bm25"
+    )
 
 from powermem.prompts import GraphPrompts, GraphToolsPrompts
-
 from powermem.storage.oceanbase import constants
 
 logger = logging.getLogger(__name__)
@@ -54,8 +70,10 @@ logger = logging.getLogger(__name__)
 try:
     import jieba
 except ImportError:
-    logger.warning("jieba is not installed. Falling back to simple space-based tokenization. "
-                   "Install jieba for better Chinese text segmentation: pip install jieba")
+    logger.warning(
+        "jieba is not installed. Falling back to simple space-based tokenization. "
+        "Install jieba for better Chinese text segmentation: pip install jieba"
+    )
     jieba = None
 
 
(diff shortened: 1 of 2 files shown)


wayyoungboy:
Thanks for taking the entity de-dup side of #1003. I agree with the direction: `UNIQUE(user_id, name)` plus `INSERT ... ON DUPLICATE KEY UPDATE` is the right complement to the transaction fix.

A couple of coordination points after #1011:

1. Please rebase/merge this on top of #1011 and keep `_create_entity()` in the same `add()` transaction. #1011 now threads a SQLAlchemy `conn` through the delete/add/relationship write path, including entity inserts. If this PR keeps opening its own `engine.connect()` and calling `conn.commit()` inside `_create_entity()`, entity writes are split out of the `add()` transaction again and rollback will not cover them. The shape I would suggest is `_create_entity(..., conn=None)`: use the passed `conn` when present, and only create/commit its own connection for standalone helper calls.

2. The legacy migration still needs a guard for tables that already have `user_id` but only have a non-unique `idx_user_name`. The create-table change here only affects new/recreated tables; `_create_tables()` skips existing user-scoped tables, and the `ON DUPLICATE KEY` path will not fire without a real `UNIQUE(user_id, name)`. We should detect whether `idx_user_name` is unique and either migrate it to a unique index, or fail with a clear migration error if duplicate rows already exist. A unit test for "existing table has `user_id` but non-unique `idx_user_name`" would catch this.

With those two changes, #1011 and #1032 should fit together cleanly: #1011 provides atomic graph `add()` writes, and this PR provides cross-request entity de-duplication.

wayyoungboy (commented):
Code review — looked at the full diff in detail.

## Overall

The `INSERT ON DUPLICATE KEY UPDATE` approach is correct and the migration logic (`_needs_unique_index_migration` + `_migrate_to_unique_index`) is well-designed. Tests cover transaction/standalone modes and both dedup branches. The `conn` parameter support for #1011 transaction participation looks good.

## Issues

### 1. Dedup DELETE may orphan relationship rows

In `_migrate_to_unique_index`, the self-join DELETE:

```sql
DELETE t1 FROM graph_entities t1
INNER JOIN graph_entities t2
ON t1.user_id = t2.user_id AND t1.name = t2.name
WHERE t1.updated_at < t2.updated_at
   OR (t1.updated_at = t2.updated_at AND t1.id < t2.id)
```

When 3+ duplicate rows exist, this correctly keeps one. But if `graph_relationships.source_entity_id` or `destination_entity_id` references a deleted entity id, those relationship rows become orphans. The migration should either:
- CASCADE: also delete relationships referencing the to-be-removed entity ids, or
- REPOINT: UPDATE relationships to reference the surviving entity id before deleting duplicates

### 2. Formatting noise

~60% of the diff is pure formatting (trailing commas, import sorting, black reformatting). This makes it hard to review actual logic changes. Consider splitting formatting into a separate commit or applying it as a pre-commit pass beforehand.

## Minor

- The `_migrate_to_unique_index` RuntimeError message is excellent — it gives the user the exact manual SQL to recover.
- Test mocks correctly validate that standalone mode commits and transaction mode does not.

With the orphan-relationship fix, this is ready to merge.

wayyoungboy (commented):
Code review — read the full diff.

## Overall

The `INSERT ON DUPLICATE KEY UPDATE` approach is correct — it preserves the existing row id (unlike `REPLACE INTO`) so relationship FKs stay valid. The migration logic (`_needs_unique_index_migration` + `_migrate_to_unique_index`) and `conn` parameter for #1011 transaction participation are well-designed. Tests cover transaction/standalone modes and both dedup branches.

## Issue: Dedup DELETE may orphan relationship rows

In `_migrate_to_unique_index`, the self-join DELETE keeps one entity row per `(user_id, name)` pair. But if `graph_relationships.source_entity_id` or `destination_entity_id` references a *deleted* entity id, those relationship rows become orphans pointing to non-existent entities.

Suggestion: before deleting duplicate entities, UPDATE relationships that reference the to-be-removed ids to point to the surviving entity id. Something like:

```sql
-- For each duplicate group, repoint relationships to the keeper
UPDATE graph_relationships SET source_entity_id = :keeper_id
WHERE source_entity_id IN (:deleted_ids);
UPDATE graph_relationships SET destination_entity_id = :keeper_id
WHERE destination_entity_id IN (:deleted_ids);
```

Or alternatively, delete orphan relationships after the dedup and log what was removed.

## Minor

- ~60% of the diff is pure formatting (trailing commas, import sorting, black). Consider splitting formatting into a separate commit next time — it makes logic review harder.
- The `RuntimeError` message in the migration fallback is excellent — gives the user exact manual SQL to recover.

wayyoungboy (approved):
LGTM

lightzt99:
LGTM.

Locally verified coverage gaps (6 additional cases, all pass):

- **`_create_tables` integration**: verified that when the entities table already exists with `user_id` and `_needs_unique_index_migration()` returns `True`, `_create_tables` calls `_migrate_to_unique_index()`; and that it is skipped when the method returns `False`. The existing suite only covers the full-recreate path (table missing `user_id`).

- **`_needs_unique_index_migration` edge branches**: (1) when `INFORMATION_SCHEMA.STATISTICS` has no row for `idx_user_name` (e.g., new table), returns `False`; (2) when the DB query raises an exception, returns `False` without propagating — startup is not aborted.

- **`_migrate_to_unique_index` failure path**: when `ALTER TABLE` raises (e.g., duplicate rows appeared between the dedup commit and the DDL), a `RuntimeError` is raised and the message contains "Manual intervention required" along with the table name, so operators know what to fix.

- **`_create_entity` SELECT-returns-None fallback**: when the `SELECT id` after `INSERT ... ON DUPLICATE KEY UPDATE` unexpectedly returns no row, the pre-generated snowflake ID is returned rather than crashing. Shouldn't happen with a healthy UNIQUE index, but the fallback exists and is correct.

All 39 original tests still pass alongside the 6 new ones (45 total).

PR #1050, opened 2026-06-16
[codex] Add isolated no-LLM Claude hook regression tests

### LINKED ISSUE #1049: [Test]: Add isolated no-LLM Docker regression tests for Claude Code hooks
## Background
The Claude Code plugin should be testable in a fully isolated environment without host mounts, host networking, or LLM credentials. Hook behavior also needs contract tests that do not depend on the real service or external network.

## Proposed design
Add a dedicated no-LLM hook regression suite:

- Add `tests/regression/test_claude_hook_no_llm.py`.
- Add fixtures under `tests/regression/fixtures/claude_hook/` for transcript JSONL, compact payloads, prompts, and file samples.
- Add `docker/Dockerfile.claude-hook-regression` for the test environment, separate from the production Dockerfile.
- Add `make test-claude-hook-docker`.
- Build the Go hook binary inside the container.
- Start a Python stdlib fake HTTP server inside the container implementing only:
  - `POST /api/v1/memories`
  - `POST /api/v1/memories/search`
  - optional `GET /api/v1/system/health`
- Simulate Claude Code events through stdin JSON instead of requiring live Claude Code auth.
- Run Docker with `--network none`; the fake server should be reachable only via container-local loopback.

Recommended env:

```bash
POWERMEM_BASE_URL=http://127.0.0.1:8848
POWERMEM_INFER_TRANSCRIPT=0
POWERMEM_INFER_COMPACT=0
POWERMEM_INFER_FILE=0
POWERMEM_PROMPT_SEARCH=1
POWERMEM_HOOK_SCRUB=1
POWERMEM_DATA_DIR=/tmp/powermem-data
```

## No-LLM behavior
The suite must pass without LLM provider configuration, external network, host-mounted credentials, or a real PowerMem server. It verifies hook protocol contracts and payload safety.

## Full LLM behavior
This suite is intentionally no-LLM. Full-service integration tests can reuse the same fixtures but should live in a separate test target so local fast validation remains deterministic.

## Acceptance criteria
- `UserPromptSubmit`, `SessionEnd`, and `PostCompact` are covered initially.
- Future hooks such as `SessionStart`, `PreCompact`, `PostToolUse`, `PostToolUseFailure`, and `Stop` can add fixtures to the same suite.
- Detached workers are awaited with bounded polling and leave no residual processes.
- Missing transcript, empty transcript, malformed JSONL, search disabled, API key header, user/agent/run metadata, and scrubbed secrets are tested.
- The test container uses no host mount and no host network.

## Tests
- `make test-claude-hook-docker` builds the image and runs the suite.
- The fake server records request bodies and headers for exact assertions.
- Sentinel secrets must not appear in captured requests, stdout, or stderr.

wayyoungboy:
## Summary

- Add an isolated Docker regression target for Claude Code hook behavior.
- Build the hook binary inside the regression image and run the suite without host mounts or container network access.
- Add stdlib-only fake-server tests for `UserPromptSubmit`, `SessionEnd`, and `PostCompact`.
- Add opt-in hook scrubbing for request and stdout payload safety.

Closes #1049

## Validation

- branch name check
- `python -m py_compile tests/regression/test_claude_hook_no_llm.py`
- `git diff --check`
- `go build -trimpath ./cmd/powermem-hook`
- `python tests/regression/test_claude_hook_no_llm.py`
- `make test-claude-hook-docker`
- `cd apps/claude-code-plugin && go test ./...`


lightzt99:
Thanks for the clean isolation approach — stdlib-only test runner and `--network none` Docker are well thought out.

Two issues that will break CI on merge:

**1. Dockerfile missing `go.sum`**

`Dockerfile.claude-hook-regression` copies `go.mod` and `cmd/` but not `go.sum`. Go 1.16+ requires `go.sum` for module verification; `go build` will fail without it.

```dockerfile
COPY apps/claude-code-plugin/go.mod ./
+ COPY apps/claude-code-plugin/go.sum ./
COPY apps/claude-code-plugin/cmd ./cmd
```

If `cmd/powermem-hook` also imports local packages outside `cmd/`, those directories need to be copied too.

**2. `actions/checkout@v6` does not exist**

The new `claude-hook-regression` job references `actions/checkout@v6`. The latest stable release is `v4`. This will fail immediately at job setup.

```yaml
- uses: actions/checkout@v4
```

---

A few smaller items worth noting:

- **`SENTINEL_SECRET` scrub pattern** is production code shaped around a test fixture. If the intent is to test scrubbing behavior, use one of the existing patterns (`sk-`, `Bearer`, etc.) as the fixture trigger instead of adding a test-only rule to the scrub table.
- **`scrubText` applied to injected context** (`formatSearchResults`): scrubbing outbound requests makes sense; silently rewriting memory content that gets injected into Claude's context does not — it can corrupt legitimate values that happen to match a pattern.
- **Double-scrub**: `prompt` is scrubbed in `handleUserPromptSubmit` then scrubbed again inside `searchMemoriesForPrompt`. Harmless but redundant.
- **`hookScrubEnabled()`** calls `os.Getenv` on every `scrubText` invocation. Cache it once at startup.

wayyoungboy:
Thanks for the detailed pass. Addressed in the latest push:

- Updated the regression Dockerfile to copy `go.*`, so it works with the current `go.mod`-only module and will include `go.sum` if/when present.
- Changed the new `claude-hook-regression` job checkout step to `actions/checkout@v4`.
- Removed the test-only `SENTINEL_SECRET` production scrub rule; the regression fixtures now exercise existing Bearer-pattern scrubbing instead.
- Stopped scrubbing returned memory content before injecting it into Claude context.
- Removed the prompt double-scrub; the outbound search body still scrubs the query once.
- Cached `POWERMEM_HOOK_SCRUB` at process startup.

Validation run:
- `go test ./...` from `apps/claude-code-plugin`
- `python tests/regression/test_claude_hook_no_llm.py`
- `make test-claude-hook-docker`

PR #1057, opened 2026-06-17
feat: move embedding model download into PowerMem with CN-aware routing

lightzt99:
## Summary

Moves the `all-MiniLM-L6-v2` embedding model download from the shell-layer
`preload-model.sh` into `PyseekdbDefaultEmbedding` itself, with automatic
CN-network detection and ModelScope fallback.  Previously the download only
ran when `POWERMEM_INIT_PRELOAD_MODEL=1` was set; on first use without that
flag the server started healthy but hung for up to 30 s on the first embed
request.

## Root cause

`_load_sentence_transformer_with_fallback` attempted a direct HuggingFace
download with a 30-second thread timeout, with no CN-aware routing.  CN
users behind the GFW consistently timed out or never completed the download.
The ModelScope path existed only as a manual recovery command printed in the
error message, not as an automatic fallback.

The `POWERMEM_INIT_PRELOAD_MODEL` opt-in in `init.sh` partially addressed
this for the Claude Code plugin path, but left bare `powermem-server`
invocations and non-plugin deployments with no CN support at all.

## Changes

- Add `_detect_country()` to `pyseekdb_default.py`: mirrors `common.sh`
  `detect_public_ip_country` logic — probes `ipapi.co`, `ifconfig.co`,
  `ipinfo.io` in sequence with a 3-second timeout each; returns ISO-3166
  two-letter code or empty string on failure.
- Add `_download_via_modelscope()`: downloads the model via
  `uvx --from modelscope`; respects `POWERMEM_UV_BIN`,
  `POWERMEM_BOOTSTRAP_PYTHON`, `POWERMEM_MODELSCOPE_PACKAGE`, and
  `POWERMEM_UV_INDEX_URL`; auto-applies the Tsinghua PyPI mirror when
  country is CN and `POWERMEM_UV_INDEX_URL` is unset.
- Add `_bridge_modelscope_to_hf_cache()`: ports the bridge logic from
  `preload-model.sh` to Python — copies the ModelScope cache into the
  HuggingFace hub snapshot layout so `huggingface_hub` and
  `sentence_transformers` find a standard HF cache.  Falls back to a
  hardcoded revision SHA when the HF API is unreachable.
- Restructure `_load_sentence_transformer_with_fallback()`: country
  detection and ModelScope download now run **before** the optional
  `sentence_transformers` import, so the model is fetched even when
  `sentence_transformers` is not installed; the non-CN HF path is
  unchanged.
- Remove `POWERMEM_INIT_PRELOAD_MODEL` branch from `init.sh`; replace
  with a deprecation notice.  `preload-model.sh` is kept as a standalone
  offline pre-warm tool but is no longer called from the main init flow.
- Add 24 unit tests covering all new functions and edge cases.

## Validation

- `python3.11 -m pytest tests/unit/test_pyseekdb_default_download.py -v` — 24 passed
- `python3.11 -m pytest tests/unit/test_pyseekdb_default_embeddings.py -v` — 8 passed, no regression
- `git diff --check`

### DIFF
--- a/apps/claude-code-plugin/scripts/init.sh
+++ b/apps/claude-code-plugin/scripts/init.sh
@@ -635,11 +635,8 @@ export_env_file_vars "$ENV_FILE"
 echo "Backend package: $PACKAGE"
 echo "Backend launcher: uvx --from '$PACKAGE' powermem-server"
 
-if [ "${POWERMEM_INIT_PRELOAD_MODEL:-0}" = "1" ] || [ "${POWERMEM_INIT_PRELOAD_MODEL:-}" = "true" ]; then
-  echo "Preloading default local embedding model through uvx."
-  sh "$SCRIPT_DIR/preload-model.sh" "$BOOTSTRAP_PYTHON"
-else
-  echo "Skipping model preload. Set POWERMEM_INIT_PRELOAD_MODEL=1 to download via ModelScope and bridge to HuggingFace cache."
+if [ -n "${POWERMEM_INIT_PRELOAD_MODEL:-}" ]; then
+  echo "POWERMEM_INIT_PRELOAD_MODEL is deprecated; the embedding model is now downloaded automatically by PowerMem at startup."
 fi
 
 if [ -n "${POWERMEM_INIT_PORT:-}" ]; then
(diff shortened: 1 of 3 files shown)


wayyoungboy:
LGTM

PR #1026, opened 2026-06-14
Support no-LLM mode

wayyoungboy:
## Summary
- Add a `noop` LLM provider for explicitly disabled LLM mode.
- Preserve basic memory add/search/update/delete while skipping fact extraction, action planning, graph extraction/search, profile extraction, query rewrite, compression, skill distillation/merge, and LLM-based importance scoring.
- Let the Claude Code plugin init fall back to no-LLM mode when LLM credentials are missing or incomplete, and report disabled LLM status in CLI/health checks.

## Validation
- `python -m pytest tests/unit/test_noop_llm.py tests/integration/test_noop_llm_mode.py -q`
- `python -m pytest tests/unit/test_noop_llm.py tests/integration/test_noop_llm_mode.py tests/unit/test_config_loader.py tests/unit/test_memory.py tests/unit/intelligence/test_memory_optimizer.py tests/unit/intelligence/test_intelligent_memory_manager.py tests/unit/server/test_user_service.py -q`
- `python -m compileall -q src/powermem/integrations/llm src/powermem/core src/powermem/user_memory src/powermem/intelligence src/powermem/agent/components src/server apps/claude-code-plugin/scripts`
- `sh -n apps/claude-code-plugin/scripts/init.sh && git diff --check`

PR #1058, opened 2026-06-18
feat: add NoopEmbedding provider for EMBEDDING_PROVIDER=none

lightzt99:
## Summary

Adds explicit support for disabling embedding entirely via `EMBEDDING_PROVIDER=none`. When set, PowerMem skips all model loading and vector generation, allowing deployments that rely purely on LLM-based memory without any embedding dependency.

This mirrors the existing `LLM_PROVIDER=noop` pattern introduced in #1026.

## Motivation

There was no supported way to run PowerMem without an embedding model. Setting an invalid provider raised `ValueError: Unsupported Embedder provider`; omitting the config fell back to the local `all-MiniLM-L6-v2` model which downloads on first use. Users who want LLM-only memory (e.g. simple fact extraction without semantic search) had no clean opt-out.

## Changes

**New files**
- `src/powermem/integrations/embeddings/noop.py` — `NoopEmbedding(EmbeddingBase)` with `is_noop = True`; `embed()` and `embed_batch()` return `[]` without touching any model or network.
- `src/powermem/integrations/embeddings/config/providers.py` — `NoopEmbeddingConfig` registered under `provider="none"`.

**Modified files**
- `factory.py` — route `provider_name == "none"` to `NoopEmbedding()` before the generic registry lookup.
- `adapter.py` — two guards:
  - `add_memory`: change `if vector is None` → `if not vector` so that both `None` and `[]` (returned by `NoopEmbedding.embed()` in the regular memory add path) trigger the mock-vector fallback instead of calling `create_col(vector_size=0)`.
  - `add_memory`: skip `embedding_service.embed()` when `embedding_service.is_noop is True`.
- `memory.py` — add `_is_embedding_disabled()` and `_embed()` helpers (mirrors `_is_llm_disabled()`). The five direct `self.embedding.embed()` calls in the skill store methods (`add_skill`, `search_skills`) are replaced with `self._embed()`, which returns `None` when disabled so downstream stores receive `query_embedding=None` and fall back to full-text search.

**Behaviour when disabled**

| Operation | Result |
|-----------|--------|
| `memory.add()` | Stored with mock vector `[0.1]*1536` |
| `memory.get()` / `update()` / `delete()` | Work normally |
| `memory.search()` | Returns `[]` (no vector index) |
| `add_skill` / `search_skills` | Store/search with `None` embedding (full-text only) |

## Validation

```
pytest tests/unit/test_noop_embedding.py -v             # 17 passed
pytest tests/integration/test_noop_embedding_mode.py -v # 4 passed
pytest tests/unit/test_noop_llm.py tests/unit/test_pyseekdb_default_embeddings.py -v  # no regression
```

### DIFF
--- a/src/powermem/core/memory.py
+++ b/src/powermem/core/memory.py
@@ -805,6 +805,20 @@ class Memory(MemoryBase):
         """Return True when PowerMem is running without LLM-backed features."""
         return self.llm_provider == "noop" or getattr(self.llm, "is_noop", False) is True
 
+    def _is_embedding_disabled(self) -> bool:
+        """Return True when embedding is explicitly disabled (EMBEDDING_PROVIDER=none)."""
+        return self.embedding_provider == "none" or getattr(self.embedding, "is_noop", False) is True
+
+    def _embed(self, text: str) -> Optional[List[float]]:
+        """Embed text, returning None when embedding is disabled or fails."""
+        if self._is_embedding_disabled():
+            return None
+        try:
+            return self.embedding.embed(text)
+        except Exception as e:
+            logger.warning("Embedding failed: %s", e)
+            return None
+
     def _get_component_config(self, component: str) -> Dict[str, Any]:
         """
         Helper method to get component configuration uniformly.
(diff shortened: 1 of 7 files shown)


wayyoungboy:
LGTM

PR #1075, opened 2026-06-22
fix(intelligence): replace unsafe number-grabbing fallback in _parse_importance_response

### LINKED ISSUE #1074: fix(intelligence): _parse_importance_response second-level fallback blindly grabs first number
## Problem

In `src/powermem/intelligence/importance_evaluator.py`, the `_parse_importance_response()` method has a three-level fallback for parsing LLM responses:

1. **L1**: Extract JSON substring via `find("{")` / `rfind("}")` and look for `importance_score`
2. **L2**: `re.findall(r'\d+\.?\d*', response)` — grab the **first number** from the entire response text
3. **L3**: Return fixed `0.5`

**The second-level fallback is semantically unverifiable** — it grabs any number without context, which can produce incorrect importance scores that cascade into downstream systems.

## Impact

`importance_score` directly affects:
- `_classify_memory_type()` — determines `working` / `short_term` / `long_term` classification
- `initial_retention = initial_retention * importance_score` — initial retention rate in Ebbinghaus algorithm
- `adjusted_interval = interval * (1 - importance_score * 0.3)` — review schedule intervals

### Concrete failure scenarios

| Scenario | Number grabbed | Actual meaning | Consequence |
|----------|---------------|----------------|-------------|
| JSON parse fails but text contains criteria scores | First dimension score (e.g. `0.9`) | A single criterion, not overall | Wrong classification |
| Reasoning mentions "6 dimensions" | `6` | Text description | `min(1.0, 6)` = **1.0** → forced `long_term` |
| "Importance 85%" | `85` | Percentage expression | clamp → **1.0** |
| Text mentions threshold "0.6 short_term" | `0.6` | Config explanation | Unrelated to content |

## Inconsistency with other modules

Other parsers in the same codebase (`parse_fact_extraction_json`, `parse_memory_actions_json`, `skill_manager._parse_skills`) return empty structures on parse failure and never guess numbers. The importance evaluator is an outlier with higher downstream cost.

## Asymmetry with exception fallback

- **LLM call throws exception** → falls back to `_rule_based_evaluation(content, metadata, context)` (content-aware)
- **LLM call succeeds but parse is ambiguous** → may return a random number via L2, **never triggers rule engine**

L2 "pretends to succeed" rather than "failing safely."

## Suggested fix direction

1. **L1**: Use `parse_json_from_text()` (existing utility); support both `importance_score` and `overall_score` field names; when only `criteria_scores` is present, compute weighted sum using `criteria_weights`
2. **L2**: Only extract numbers in field-name context (e.g. `importance_score"\s*:\s*`); reject values outside `[0.0, 1.0]` instead of clamping
3. **L3**: Return `None` → caller falls back to `_rule_based_evaluation` (consistent with exception path)

## Related files

- `src/powermem/intelligence/importance_evaluator.py` — `_parse_importance_response`, `_llm_based_evaluation`
- `src/powermem/prompts/importance_evaluation.py` — Prompt JSON structure with six criteria
- `src/powermem/intelligence/intelligent_memory_manager.py` — consumes `importance_score`
- `src/powermem/intelligence/ebbinghaus_algorithm.py` — thresholds, retention, review scheduling
- `src/powermem/utils/utils.py` — `parse_json_from_text`, `extract_json` (reusable utilities)

knqiufan:
## Summary

Closed #1074

The second-level fallback in `_parse_importance_response()` blindly grabbed the first number from the entire LLM response text via `re.findall(r'\d+\.?\d*')`. This could misinterpret dimension counts (`6` → clamped to `1.0`), percentages (`85%` → `1.0`), or unrelated config references as the `importance_score`, causing incorrect memory classification and review scheduling downstream.

### Changes

- **L1 (JSON parsing):** Replace hand-rolled `find("{")`/`rfind("}")` with `parse_json_from_text()` (existing utility); support both `importance_score` and `overall_score` field names; when only `criteria_scores` is present, compute weighted sum using `criteria_weights`
- **L2 (field-name regex):** Only extract numbers anchored to recognized field names (`importance_score`, `overall_score`, `score`); reject values outside `[0.0, 1.0]` instead of clamping
- **L3 (safe failure):** Return `None` instead of fixed `0.5`; caller `_llm_based_evaluation` falls back to `_rule_based_evaluation` (consistent with LLM exception path)
- **Tests:** Add 17 unit tests covering all parsing paths, rejection of invalid numbers, and fallback behavior

### Why

`importance_score` directly determines:
- Memory type classification (`working` / `short_term` / `long_term`)
- Initial retention rate (`initial_retention * importance_score`)
- Review schedule intervals (`interval * (1 - importance_score * 0.3)`)

A silently wrong score cascades through the entire memory lifecycle and is nearly impossible to diagnose from external behavior.

## Test plan

- [x] All 17 new unit tests pass (`tests/unit/intelligence/test_importance_evaluator.py`)
- [x] All 63 existing intelligence unit tests pass with no regression
- [x] CI regression suite passes

### DIFF
--- a/src/powermem/intelligence/importance_evaluator.py
+++ b/src/powermem/intelligence/importance_evaluator.py
@@ -5,9 +5,11 @@ This module evaluates the importance of memory content using LLM.
 """
 
 import logging
+import re
 from typing import Any, Dict, Optional
-import json
+
 from ..prompts.importance_evaluation import ImportanceEvaluationPrompts
+from ..utils.utils import parse_json_from_text
 
 logger = logging.getLogger(__name__)
 
@@ -192,6 +194,13 @@ class ImportanceEvaluator:
             # Parse the response to extract importance score
             importance_score = self._parse_importance_response(response)
             
+            if importance_score is None:
+                logger.warning(
+                    "LLM response could not be parsed reliably, "
+                    "falling back to rule-based evaluation"
+                )
+                return self._rule_based_evaluation(content, metadata, context)
+            
             logger.debug(f"LLM evaluated importance: {importance_score}")
             
             return importance_score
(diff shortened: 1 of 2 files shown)


lightzt99:
LGTM

Issue #1056, opened 2026-06-17
Refactor: make MCP session handling stateless to support multi-worker deployment


lightzt99:
## Background

The PowerMem HTTP server and MCP service run in the same process. When `workers > 1` (default is 4 in Docker), uvicorn forks N worker processes, each importing the FastMCP ASGI app independently.

## Problem

FastMCP's `streamable-http` transport is **stateful at the process level**. On `initialize`, the MCP client receives a `Mcp-Session-Id` response header and includes it in all subsequent requests. FastMCP maintains an in-process session registry to look up these sessions.

With multiple workers, each worker has its own FastMCP instance and its own session registry:

```
Client → initialize → worker 0  (session S1 created in worker 0's registry)
Client → next call  → worker 1  (S1 not in worker 1's registry → error)
```

The MCP ASGI app is created once at module import in `src/server/main.py:47`:

```python
mcp_asgi_app = _load_mcp_asgi_app()  # one per worker process
```

Each worker creates its own `mcp` object (`src/powermem/mcp/server.py:51`) with an independent session map. Requests for the same session can land on any worker, making every cross-worker call fail.

## Current Mitigation (Incomplete)

`src/server/cli/server.py:291-299` forces `workers=1` when embedded storage is detected:

```python
if not config.reload and config.workers != 1 and _is_embedded_storage():
    config.workers = 1
```

This protects SQLite users but does **not** protect OceanBase users, who run the Docker default of 4 workers with MCP fully enabled.

## Proposed Fix

**Option A — Guard at startup (immediate safeguard)**
Extend the existing check: if MCP is enabled and `workers > 1`, force `workers=1` (or print a hard error). No architectural change required.

```python
if not config.reload and config.workers != 1 and mcp_is_enabled():
    config.workers = 1
```

**Option B — Stateless MCP transport (full refactor)**
Move session state to a shared backend (Redis or DB) so any worker can serve any MCP session. This requires changes to FastMCP session management or replacing `streamable-http` with `stdio` per-connection.

Option A is the lowest-risk short-term fix; Option B is the correct long-term solution for horizontal scalability.

PsiACE:
PowerContext currently runs MCP in a single process and does not expose a multi-worker mode. This PowerMem-specific issue does not apply, so we are closing it as not planned.

> We are now developing PowerContext, the next generation of PowerMem, on the `master` branch.

PR #1076, opened 2026-06-22
Force single worker when MCP is mounted

wayyoungboy:
Refs #1056

Note: this PR implements the short-term single-worker guard from #1056. The longer-term stateless MCP session handling needed for true multi-worker support remains tracked there.

Summary:
- Detect when the server CLI can mount the streamable HTTP MCP app.
- Reuse the existing single-worker guard so MCP sessions are not split across multiple worker processes.
- Keep multi-worker startup unchanged when no process-local runtime mode is active, and avoid running the guard checks in reload mode.

Validation:
- Using the project virtual environment, ran `python -m pytest tests/unit/server/test_server_cli.py tests/unit/server/test_main_mcp_optional.py -q` (31 passed, 1 warning from an existing dependency deprecation).
- Using the project virtual environment, ran `python -m py_compile src/server/cli/server.py tests/unit/server/test_server_cli.py`.
- Ran `git diff --check`.
- E2E: with MCP dependencies installed and non-embedded storage configuration, started `powermem-server --host <loopback host> --port <local test port> --workers 2 --no-open-browser`; startup printed the MCP single-worker guard and forced `workers=1` before server initialization. The process was stopped after the guard was observed.

### DIFF
--- a/src/server/cli/server.py
+++ b/src/server/cli/server.py
@@ -13,6 +13,7 @@ import time
 import urllib.error
 import urllib.request
 import webbrowser
+from importlib.util import find_spec
 from typing import Optional
 
 
@@ -240,6 +241,11 @@ def _is_embedded_storage() -> bool:
     return False
 
 
+def _is_mcp_enabled() -> bool:
+    """Return whether the in-process server will mount the streamable HTTP MCP app."""
+    return find_spec("fastmcp") is not None
+
+
 def _run_server_app(**kwargs) -> None:
     """Start uvicorn after verifying server extras are installed."""
     _require_server_deps()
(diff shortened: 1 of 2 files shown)


lightzt99:
Prefer longer-term stateless MCP mentioned in #1056 after discuss.

wayyoungboy:
Self-review status for the current head: I reviewed the single-worker guard for MCP-mounted server runs. The change matches the current mount condition: when `fastmcp` is available, the server can mount the MCP ASGI app, so forcing `workers=1` avoids running that mounted MCP app in multiple worker processes. Reload mode remains untouched, and the current checks are green.

I did not find a blocking issue in this static review. I am not marking this LGTM because this PR is authored from the `wayyoungboy` branch and still needs independent maintainer review/acceptance.

PR #1052, opened 2026-06-16
Fix memory search fallback when embeddings are unavailable

wayyoungboy:
## Summary

- fall back to text search when query embeddings cannot be generated
- add retrieval mode and fusion controls across memory search APIs
- preserve vector-only and OceanBase backend compatibility while keeping FTS threshold behavior distinct from result quality
- keep metadata filters consistent across vector, FTS, and hybrid retrieval paths
- rebase onto the latest `main` and preserve metadata filters whose names collide with payload fields such as `hash` and `data`
- preserve explicit no-embedding mode behavior: default search stays empty when embeddings are intentionally disabled, while transient embedding failures can still fall back to FTS

## Tests

- after activating the project virtual environment, `python -m pytest --ignore=tests/regression -m "not e2e and not e2e_config"`
- after activating the project virtual environment, `python -m pytest tests/unit/test_sqlite_fts.py tests/unit/test_memory.py tests/unit/server/test_search_score_threshold.py tests/unit/test_list_memory_filters.py tests/integration/test_sqlite_fts_integration.py`
- after activating the project virtual environment, `python -m pytest tests/unit/test_intelligence_forget_soft_mark.py tests/integration/test_noop_embedding_mode.py tests/unit/test_noop_embedding.py -q`
- public `Memory` API E2E with SQLite in-memory: forced embedding generation failure, then verified FTS fallback returns the single result matching collision-prone metadata filters; also verified explicitly disabled embeddings keep default search empty
- public `Memory` API E2E with a remote OceanBase test database: forced embedding generation failure, then verified OceanBase FTS fallback returns the expected metadata-filtered result; also verified explicitly disabled embeddings keep default search empty while explicit FTS retrieval still works

Fixes #1046

PR #1091, opened 2026-06-25
fix(storage): enable FTS fallback for NoopEmbedding across core and adapter layers

lightzt99:
## Summary

When `EMBEDDING_PROVIDER=none` (NoopEmbedding, added in #1058), memory search returned `[]` unconditionally. Two layers blocked the FTS fallback:

1. **Core layer** — `Memory.search()` and `AsyncMemory.search()` short-circuited to `{"results": [], "relations": []}` as soon as `_is_embedding_disabled()` was true, before the adapter ever saw the request.
2. **Adapter layer** — `StorageAdapter.search_memories()` returned `[]` when `query_embedding` was falsy, and `_supports_text_search_without_vector()` gated OceanBase FTS fallback behind the `hybrid_search` flag, so deployments with `hybrid_search=False` had no text-only path even when `query` was present.

This PR removes both blockers so the request flows through to the storage layer's FTS path when embedding is disabled, giving users keyword-based recall instead of silent empty results.

### Core layer
- `src/powermem/core/memory.py` — drop the `_is_embedding_disabled()` early-return in `Memory.search()`; keep `query_embedding = None` and let the adapter handle the fallback. Logs `Embedding disabled; falling back to FTS for query` for visibility.
- `src/powermem/core/async_memory.py` — same change in `AsyncMemory.search()`.

### Adapter layer
- `src/powermem/storage/adapter.py` — `_supports_text_search_without_vector()` now returns `True` for OceanBase unconditionally (no longer gated by `hybrid_search`). The existing FTS fallback path in `search_memories()` (added in #1052) already forwards `query_vector=None` when `query_embedding` is empty and `query` is present, so SQLite and OceanBase both route to full-text search.
- `src/powermem/storage/oceanbase/oceanbase.py` — `search()` now branches to `_fulltext_search()` when `vectors is None` and `query` is present, so OceanBase deployments with `hybrid_search=False` also get FTS fallback (previously the `not self.hybrid_search` branch would call `_vector_search` with no vectors and return `[]`).

### Tests
- `tests/integration/test_noop_embedding_mode.py` — renamed `test_noop_embedding_search_returns_empty` → `test_noop_embedding_search_returns_fts_hits`; asserts the FTS hit for both sync `Memory` and async `AsyncMemory`.
- `tests/unit/test_adapter_fts_fallback.py` — 5 unit tests covering:
  - Empty list embedding + query text → FTS hits
  - `None` embedding + query text → FTS hits
  - Empty embedding + no query → `[]`
  - Unrelated query → `[]`
  - FTS respects `user_id` scoping

## Validation

- `pytest tests/integration/test_noop_embedding_mode.py tests/unit/test_noop_embedding.py tests/unit/test_adapter_fts_fallback.py tests/unit/test_oceanbase_fts_quality_score.py tests/unit/test_sqlite_fts.py tests/unit/test_memory.py -q` — 130 passed
- No regressions in `test_list_memory_filters.py`, `test_search_score_threshold.py`, etc. (138 related unit tests pass)

## Behavior matrix

| Configuration | Before | After |
|---|---|---|
| SQLite + `EMBEDDING_PROVIDER=none` + query text | `[]` (core short-circuit) | FTS hits |
| OceanBase + `hybrid_search=True` + `none` + query text | `[]` (core short-circuit) | FTS hits |
| OceanBase + `hybrid_search=False` + `none` + query text | `[]` (core + adapter + store all block) | FTS hits |
| Any backend + `none` + no query text | `[]` | `[]` (degenerate, unchanged) |
| Real embedding + query text | Vector / hybrid search | Vector / hybrid search (unchanged) |

## Notes

- OceanBase `hybrid_search=False` deployments now also get FTS fallback. Previously the `not self.hybrid_search` branch in `search()` would call `_vector_search(query, vectors=None, ...)` and return `[]`; the new `vectors is None and query` branch routes to `_fulltext_search` first.
- The skill store path (`add_skill` / `search_skills`) was already correct and is not touched here — `Memory._embed()` returns `None` (not `[]`) when `_is_embedding_disabled()` is true, so `OceanBaseSkillStore.search(query_embedding=None, query_text=...)` already skipped vector search.
- Rebased onto `upstream/main` (`71003f5`) to pick up #1092 and #1130.

### DIFF
--- a/src/powermem/core/async_memory.py
+++ b/src/powermem/core/async_memory.py
@@ -1095,20 +1095,20 @@ class AsyncMemory(MemoryBase):
             query_embedding = None
             if retrieval_mode != "fts":
                 if self._is_embedding_disabled():
-                    return {
-                        "results": [],
-                        "relations": []
-                    }
-                try:
-                    query_embedding = await asyncio.to_thread(
-                        embedding_service.embed, query, memory_action="search"
+                    logger.info(
+                        "Embedding disabled; falling back to FTS for query"
                     )
-                except Exception as exc:
-                    logger.warning(
-                        "Search embedding failed; falling back to text search "
-                        "when available: %s",
-                        exc,
-                    )
+                else:
+                    try:
+                        query_embedding = await asyncio.to_thread(
+                            embedding_service.embed, query, memory_action="search"
+                        )
+                    except Exception as exc:
+                        logger.warning(
+                            "Search embedding failed; falling back to text search "
+                            "when available: %s",
+                            exc,
+                        )
             
 
             # Search in storage asynchronously - pass query text to enable hybrid search
--- a/src/powermem/core/memory.py
+++ b/src/powermem/core/memory.py
@@ -1779,20 +1779,20 @@ class Memory(MemoryBase):
             query_embedding = None
             if retrieval_mode != "fts":
                 if self._is_embedding_disabled():
-                    return {
-                        "results": [],
-                        "relations": []
-                    }
-                try:
-                    query_embedding = embedding_service.embed(
-                        query, memory_action="search"
+                    logger.info(
+                        "Embedding disabled; falling back to FTS for query"
                     )
-                except Exception as exc:
-                    logger.warning(
-                        "Search embedding failed; falling back to text search "
-                        "when available: %s",
-                        exc,
-                    )
+                else:
+                    try:
+                        query_embedding = embedding_service.embed(
+                            query, memory_action="search"
+                        )
+                    except Exception as exc:
+                        logger.warning(
+                            "Search embedding failed; falling back to text search "
+                            "when available: %s",
+                            exc,
+                        )
             
 
             # Search in storage - pass query text to enable hybrid search
--- a/src/powermem/storage/adapter.py
+++ b/src/powermem/storage/adapter.py
@@ -81,7 +81,7 @@ class StorageAdapter:
         if store_module.endswith("sqlite.sqlite_vector_store"):
             return True
         if ".oceanbase." in store_module:
-            return bool(getattr(target_store, "hybrid_search", False))
+            return True
         return False
 
     def _metadata_filter_key_for_store(
--- a/src/powermem/storage/oceanbase/oceanbase.py
+++ b/src/powermem/storage/oceanbase/oceanbase.py
@@ -921,6 +921,18 @@ class OceanBaseVectorStore(VectorStoreBase):
                 ]
             return results
 
+        if vectors is None and query:
+            logger.info(
+                "OceanBase search: no query embedding; falling back to FTS"
+            )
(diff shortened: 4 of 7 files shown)


wayyoungboy:
Thanks for the PR. I cannot mark this ready yet because both CI test jobs are currently failing:

- `test (3.11)`
- `test (3.12)`

Both fail in `tests/integration/test_noop_embedding_mode.py::test_noop_embedding_search_returns_empty`: the test still expects an empty result in noop embedding mode, while this change now returns an FTS hit for the added memory. Please reconcile the intended noop/`EMBEDDING_PROVIDER=none` search contract before merge, either by updating the test and documenting automatic FTS fallback as the new default behavior, or by keeping default search empty and requiring an explicit FTS retrieval mode. This also overlaps with #1052, which adds explicit FTS retrieval controls, so it would be good to align the two PRs before one lands.

wayyoungboy (changes requested):
Requesting changes because the current CI failures line up with the behavior change in this PR.

Both `test (3.11)` and `test (3.12)` fail in `tests/integration/test_noop_embedding_mode.py::test_noop_embedding_search_returns_empty`: the existing integration contract expects default search to return no results when embeddings are explicitly disabled, but this PR now returns an FTS hit.

Please settle the intended contract before merge: either document/update tests so automatic FTS fallback is the new default for `EMBEDDING_PROVIDER=none`, or keep default no-embedding search empty and require an explicit FTS retrieval mode. This should also be aligned with #1052 before either path lands.

wayyoungboy (changes requested):
Rechecked the current head. The Python checks are green now, but the original contract issue is still unresolved: this PR changes default search with `EMBEDDING_PROVIDER=none` to automatic FTS fallback and updates the integration test to expect a hit.

That still conflicts with the direction in #1052, where explicit no-embedding mode stays empty by default and full-text lookup is exposed through an explicit retrieval mode such as `retrieval_mode="fts"`. Please align these two PRs before merge. A clean path would be to drop the automatic default fallback here, wait for #1052 to land, and add the complementary explicit retrieval-mode exposure where this PR needs it.

Keeping changes requested until there is one consistent public search contract for no-embedding mode.

wayyoungboy:
Thanks for continuing on the NoopEmbedding fallback work. I do not think this is ready to merge yet because the public search contract is still unresolved.

This PR changes `EMBEDDING_PROVIDER=none` default search behavior to automatically return FTS hits. That conflicts with the direction discussed in #1052, where no-embedding search stays empty by default and FTS is exposed through an explicit retrieval mode such as `retrieval_mode="fts"`.

Please align this PR with that contract before merge: either drop the automatic default fallback here and wait for the explicit retrieval-mode path, or get maintainer agreement to change and document the default behavior.

Teingi:

PR #1099, opened 2026-06-26
fix(server): push run_id filter to storage for session timeline

### LINKED ISSUE #1086: Session timeline run_id filter can miss older sessions
## Evidence

Recent commit `1d13b003349d06c5441411fb6df3758381a13161` / PR #1051 added the session timeline APIs. The behavior is still present on `github/main` at `9648cd53fc8e5bd97f491571c45c564777bf6443`.

In `src/server/services/memory_service.py`, `_load_session_memories()` accepts `run_id`, but the call to `self.list_memories(...)` does not pass that `run_id` down to storage. It always loads a bounded snapshot (`SESSION_TIMELINE_FETCH_CAP = 5000`) and only then filters by `run_id` in memory.

Relevant code on `9648cd53fc8e5bd97f491571c45c564777bf6443`:

- `src/server/services/memory_service.py:55` defines `SESSION_TIMELINE_FETCH_CAP = 5000`
- `src/server/services/memory_service.py:869-885` calls `list_memories(user_id=..., agent_id=..., limit=5000, offset=0, sort_by="created_at", order=...)` without `run_id`
- `src/server/services/memory_service.py:890` filters `run_id` after the bounded snapshot has already been fetched

Minimal reproduction using a mocked `MemoryService.list_memories` on `9648cd53fc8e5bd97f491571c45c564777bf6443`:

```text
list_memories_run_id None
fetched_limit 5000
timeline_total 0
events []
```

The mock returns the target session if `run_id="target-run"` is pushed to storage, but returns 5000 newer unrelated records when `run_id` is omitted. Because `_load_session_memories()` omits `run_id`, `/api/v1/memories/timeline?run_id=target-run` can report zero events even though matching records exist outside the latest 5000-record global snapshot.

This affects all APIs using `_load_session_memories()` with `run_id`:

- `GET /api/v1/memories/sessions?run_id=...`
- `GET /api/v1/memories/session-stats?run_id=...`
- `GET /api/v1/memories/timeline?run_id=...`

## Why this is user-visible

The route query parameter says it filters by run/session ID, but the current implementation only filters inside the newest global snapshot. Busy installations can therefore return empty or partial timelines for older sessions.

## Minimal fix proposal

When `run_id` is provided, fetch targeted records before applying the snapshot projection:

1. Call `list_memories(..., run_id=run_id, limit=SESSION_TIMELINE_FETCH_CAP, ...)` for records stored with the canonical run ID.
2. Also preserve alias support by fetching bounded metadata matches for `session_id == run_id` and `thread_id == run_id` (the current projection treats those as run aliases).
3. Merge records by memory ID before the existing `_memory_run_id()` filter and sorting.
4. Add a regression test where more than `SESSION_TIMELINE_FETCH_CAP` newer unrelated memories exist, and `list_timeline_events(run_id=...)` still returns the targeted older session.

This keeps the current alias behavior while making exact `run_id` queries use storage-side filtering.

kic635:
Fixes #1086. Targeted run_id/session_id/thread_id fetches replace the global 5000-row snapshot so older sessions are not dropped under noise.

## Summary

When querying session timeline APIs with a `run_id` filter, `_load_session_memories` previously loaded the latest 5000 memories globally (sorted by `created_at`) and filtered in memory. In deployments with many records, an older session could be completely missing from that snapshot even though its data still exists in storage.

This PR fixes #1086 by pushing the `run_id` filter down to the storage layer for targeted fetches, so older sessions remain reachable regardless of newer unrelated records.

## Solution Description

**Root cause:** `_load_session_memories` ignored the `run_id` parameter at fetch time and always called `list_memories` without storage-side filtering.

**Changes in `memory_service.py`:**

1. **`_fetch_session_memories_for_run_id`** — When `run_id` is provided, perform three storage-side queries:
   - `run_id=<target>` (canonical match)
   - `filters={"session_id": <target>}` (legacy alias)
   - `filters={"thread_id": <target>}` (legacy alias)

2. **`_merge_memories_by_id`** — Deduplicate results by memory ID; later batches win (run_id batch is merged last and is authoritative).

3. **`_cap_merged_session_memories`** — Bound merged results to `SESSION_TIMELINE_FETCH_CAP` (5000). Alias-only rows (legacy records not returned by the run_id query) are preserved first; remaining slots are filled from the run_id batch, sorted by `created_at` per `sort_order`.

4. **`_load_session_memories`** — Route `run_id`-filtered requests through `_fetch_session_memories_for_run_id`; keep the global snapshot path when `run_id` is absent. Post-fetch filter drops rows whose projected run id conflicts with the requested filter (e.g. `run_id="other-run"` with `session_id="target-run"`).

**Tests:** Extended `tests/unit/server/test_session_timeline.py` from 10 to 20 cases, including:
- Storage-side run_id push-down vs global snapshot
- Alias-only records preserved when run_id results hit the fetch cap
- Conflicting explicit run_id filtering
- SQLite integration test: 5001+ noise records + old target session still reachable

**Known limits:**
- Bounded snapshot: sessions with >5000 events may truncate oldest run_id matches.
- Remote HTTP storage may still perform client-side filtering (pre-existing architecture debt, out of scope for this PR).

**Test plan:**
- [x] `pytest tests/unit/server/test_session_timeline.py -v` (20 passed)

### DIFF
--- a/src/server/services/memory_service.py
+++ b/src/server/services/memory_service.py
@@ -866,6 +866,95 @@ class MemoryService:
             or observation.get("source") == OBSERVATION_SOURCE
         )
 
+    @classmethod
+    def _merge_memories_by_id(
+        cls,
+        *memory_batches: List[Dict[str, Any]],
+    ) -> List[Dict[str, Any]]:
+        """Merge memory batches. Later batches overwrite earlier ones for each memory ID."""
+        merged: Dict[str, Dict[str, Any]] = {}
+        for batch in memory_batches:
+            for memory in batch:
+                memory_id = cls._memory_id(memory)
+                if memory_id is None:
+                    logger.debug("Skipping memory without identifiable ID during merge")
+                    continue
+                merged[str(memory_id)] = memory
+        return list(merged.values())
+
+    def _cap_merged_session_memories(
+        self,
+        run_results: List[Dict[str, Any]],
+        merged: List[Dict[str, Any]],
+        sort_order: str,
+    ) -> List[Dict[str, Any]]:
+        """Bound merged fetch results while preserving alias-only rows not in run_results."""
+        cap = SESSION_TIMELINE_FETCH_CAP
+        if len(merged) <= cap:
+            return merged
+
+        run_result_ids = {
+            str(memory_id)
+            for memory in run_results
+            if (memory_id := self._memory_id(memory)) is not None
+        }
+        alias_only = [
+            memory
+            for memory in merged
+            if (memory_id := self._memory_id(memory)) is not None
+            and str(memory_id) not in run_result_ids
+        ]
+        run_batch = [
+            memory
+            for memory in merged
+            if (memory_id := self._memory_id(memory)) is not None
+            and str(memory_id) in run_result_ids
+        ]
+
+        reverse = sort_order == "desc"
+        sort_key = lambda memory: self._memory_datetime(memory) or datetime.min
+        run_batch.sort(key=sort_key, reverse=reverse)
+
+        remaining = cap - len(alias_only)
+        if remaining <= 0:
+            alias_only.sort(key=sort_key, reverse=reverse)
+            return alias_only[:cap]
+
+        alias_only.sort(key=sort_key, reverse=reverse)
+        return alias_only + run_batch[:remaining]
+
+    def _fetch_session_memories_for_run_id(
+        self,
+        run_id: str,
+        user_id: Optional[str] = None,
+        agent_id: Optional[str] = None,
+        sort_order: str = "desc",
+    ) -> List[Dict[str, Any]]:
+        """Fetch session memories using storage-side run/alias filters."""
+        run_kwargs = {
+            "user_id": user_id,
+            "agent_id": agent_id,
+            "limit": SESSION_TIMELINE_FETCH_CAP,
+            "offset": 0,
+            "sort_by": "created_at",
+            "order": sort_order,
+        }
+        run_results = self.list_memories(**run_kwargs, run_id=run_id)
+        if len(run_results) >= SESSION_TIMELINE_FETCH_CAP:
+            alias_limit = SESSION_TIMELINE_FETCH_CAP
+        else:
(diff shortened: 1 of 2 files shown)


wayyoungboy (commented):
Static re-review of the current head:

- The storage-side `run_id` fetch addresses the main global-snapshot problem: targeted timeline reads no longer depend on the first global 5000-row page.
- `list_timeline_events()` re-sorts projected events before pagination, so the intermediate merge order from `run_id` / `session_id` / `thread_id` fetches should not affect event ordering.
- I did not find a code blocker in this pass.

One remaining review gate: the PR status checks I can see only include the translator check and skipped regression jobs, not the standard Python test matrix. Before final approval, please run or trigger the relevant Python tests, at least `tests/unit/server/test_session_timeline.py`, and post the result on the PR.

kic635:
> Local test run for the remaining review gate:

* Ran `python -m pytest tests/unit/server/test_session_timeline.py -v --tb=short` on branch `fix/1086-session-run-id` at commit `5501a56`.
* Result: **20 passed** in 5.36s.
* The 4 warnings are unrelated third-party `DeprecationWarning`s from `google` namespace packages via `pkg_resources`, not from this change.

```text
platform win32 -- Python 3.11.9, pytest-9.1.1
collected 20 items

tests/unit/server/test_session_timeline.py ....................  [100%]

======================= 20 passed, 4 warnings in 5.36s ========================
<img width="1242" height="718" alt="Snipaste_2026-06-29_15-55-15" src="https://github.com/user-attachments/assets/2bc05724-c9d4-4ef4-8411-b04137a2ff26" />

wayyoungboy (approved):
LGTM. I re-reviewed current head `5501a56` and did not find blocking issues.

Validation:
- `python -m pytest tests/unit/server/test_session_timeline.py -v --tb=short`: 20 passed.
- Local SQLite service-chain check with 5001 unrelated records plus an older target run passed across session summaries, session stats, and timeline results (`sessions=1`, `stats_events=2`, `timeline_events=2`).

Notes: GitHub checks currently show translator and CLA passing while the regression jobs are skipped; the targeted local tests cover the reported `run_id` timeline regression.

PR #1112, opened 2026-06-29
Fix user-scoped memory consolidation

### LINKED ISSUE #1109: [Bug]: mixed usage of user id in the powermem dashboard
### Describe the bug

The memory for userid=123 was added via the pmem client, 
The memory for userid=lixiaodong was added via the Cursor agent.

Image

**After the last memory was added via the Cursor agent, viewing userid=123 on the PowerMem dashboard...**

### Environment

powermem dashboard 1.1.4

### Fast reproduce steps

----
Mac login username: lixiaodong
----

# pmem client
pmem memory add "我喜欢喝红茶" --user-id user123
pmem memory add "我现在不喜欢喝红茶，我喜欢喝咖啡" --user-id user123

# cursor agent1
记住我2022年居住在北京

# cursor agent2
记住 2023年从北京回到了成都居住了

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional context

_No response_

wayyoungboy:
## Summary

- add adapter-side scope and logical-filter validation for memory search/list results
- avoid reporting intelligent UPDATE/DELETE actions when storage rejects them by scope
- cover user-scope leakage, logical filters, simple-backend pushdown, and sync/async rejected updates

## Validation

- `python -m pytest tests/unit/test_issue_1109_user_scope.py tests/unit/test_list_memory_filters.py tests/unit/test_oceanbase_native_filter.py -q`
- `python -m pytest tests/integration/test_noop_embedding_mode.py tests/integration/test_storage_integration.py tests/integration/test_async_memory_integration.py -q`
- Local SQLite E2E: forced search to return another user's candidate during intelligent add; verified the existing user memory was unchanged and the current user received an ADD result.

Closes #1109.

### DIFF
--- a/src/powermem/core/async_memory.py
+++ b/src/powermem/core/async_memory.py
@@ -806,7 +806,10 @@ class AsyncMemory(MemoryBase):
                     results.append({
                         "id": memory_id,
                         "memory": action_text,
-                        "event": event_type
+                        "event": event_type,
+                        "user_id": user_id,
+                        "agent_id": agent_id,
+                        "run_id": run_id,
                     })
                     action_counts["ADD"] += 1
                     
@@ -814,7 +817,7 @@ class AsyncMemory(MemoryBase):
                     # Use ID mapping to get the real memory ID (Snowflake ID - integer)
                     real_memory_id = temp_uuid_mapping.get(str(action_id))
                     if real_memory_id:
-                        await self._update_memory_async(
+                        updated_memory = await self._update_memory_async(
                             memory_id=real_memory_id,
                             content=action_text,
                             user_id=user_id,
@@ -822,13 +825,22 @@ class AsyncMemory(MemoryBase):
                             existing_embeddings=fact_embeddings,
                             metadata=metadata
                         )
-                        results.append({
-                            "id": real_memory_id,
-                            "memory": action_text,
-                            "event": event_type,
-                            "previous_memory": action.get("old_memory")
-                        })
-                        action_counts["UPDATE"] += 1
+                        if updated_memory:
+                            results.append({
+                                "id": real_memory_id,
+                                "memory": action_text,
+                                "event": event_type,
+                                "user_id": user_id,
+                                "agent_id": agent_id,
+                                "run_id": run_id,
+                                "previous_memory": action.get("old_memory")
+                            })
+                            action_counts["UPDATE"] += 1
+                        else:
+                            logger.warning(
+                                "Skipped UPDATE for memory %s because it is outside the requested scope",
+                                real_memory_id,
+                            )
                     else:
                         logger.warning(f"Could not find real memory ID for action ID: {action_id}")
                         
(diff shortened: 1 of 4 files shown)


lightzt99 (commented):
Thanks for this fix. CI is green and the sync/async + SQLite/OceanBase + nested AND/OR test coverage is solid.

Before merging, please consider the following:

**P1 — Two fail-open paths conflict with this PR's goal (scope isolation)**

1. `$nin` silently passes all rows when `op_value` is not a list/tuple/set. `_filter_value_matches` only checks `isinstance(op_value, (list, tuple, set)) and actual in op_value → return False`; when `isinstance` fails it falls through to the trailing `return True`. `$in` is fail-closed (`not isinstance(...) or actual not in op_value → return False`); `$nin` should be symmetric.
2. `_memory_matches_filter_expression`: `if not isinstance(filters, dict): return True`. A stray string/number passed as `filters` would let every memory through. Should `return False` (fail-closed) or raise.

**P1 — `search_memories` overscan cap returns fewer than `limit` for selective filters**

`search_limit = max(limit * 5, 100)`. With 10k rows, 1% selectivity (100 matches), `limit=10`: overscan 100, Python filters to 1, return 1 — but 100 actually match and the user asked for 10. `get_all_memories` uses `limit=None` (full scan) so it doesn't have this bug, but `search_memories` shouldn't under-return.

**P2 — Unknown filter operators silently drop all rows**

`_filter_value_matches` returns `False` for `$regex` / `$exists` / `$between` / `$all`. Users get an empty result with no warning, believing the filter worked. Suggest `raise ValueError(f"Unsupported filter operator: {op}")`.

**P2 — `like` / `$like` anchor semantics lost**

`str(op_value).replace("%", "")` substring match collapses `like "private%"` (prefix), `like "%private"` (suffix), and `like "%private%"` (contains) to the same `contains` behavior. SQL LIKE anchors silently don't work — risky for permission/scope filters. Either implement SQL semantics or document "substring contains" only.

**P2 — DELETE rejection path untested**

Tests cover UPDATE rejection (sync + async), but no test verifies that `_intelligent_add` correctly omits the result when `delete` is rejected by scope. Add a sync + async DELETE-rejected case.

**P3 — `count_all_memories` full-table scan has no upper bound**

`filters_fully_pushed=False` → `len(get_all_memories(limit=None))` pulls every matching row into Python. Dashboard calls `count` to show totals — at 100k memories a single count is O(N). Suggest a cap (e.g. 100k) or document the limitation.

**P3 — Warning message is inaccurate**

`_intelligent_add` logs `"Skipped UPDATE for memory %s because it is outside the requested scope"`, but `_update_memory` returning `None` may also mean the memory was concurrently deleted or never existed. Suggest `"not found or outside requested scope"`.

**P3 — Single 740-line commit is hard to review/rollback**

Three independent changes (`_intelligent_add` truthy return / search-list defensive filter / OceanBase logical filter translation) are bundled. Splitting into 3 commits would help.

**Band-aid nature**

The defensive re-check at the search/list tail is a patch, not a root-cause fix. If the real cause is SQLite `list()` not pushing `user_id` filters, the next new store path will bypass this defense too. Worth a follow-up to fix the backend's missing filter; keep this layer as defense-in-depth.

Summary: direction is right; please fix the two P1 fail-open paths and the overscan cap before merge, the rest can be follow-ups.

Teingi:

PR #1115, opened 2026-06-29
Split regression workflow into parallel shards

wayyoungboy:
## Summary
- Split scenario regression tests into four file-level shards and non-scenario regression tests into six file-level shards.
- Add a small shard-selection helper that balances files by test function count and keeps new `test_*.py` files covered automatically.
- Preserve the existing JUnit artifact aggregation while limiting each regression job to two concurrent shards.
- Include the regression workflow and helper script paths in the workflow trigger filters so workflow-only changes can be tested.

## Validation
- `actionlint .github/workflows/regression.yml`
- YAML parse of `.github/workflows/regression.yml`
- Shard planner dry-run covered 21 regression files with no missing or duplicate files.
- `uv run --no-project --python "3.11" --with-editable ".[dev,test,server,seekdb]" pytest --collect-only -q tests/regression/test_claude_hook_privacy_e2e.py`

### DIFF
--- a/.github/scripts/select_regression_shard.py
+++ b/.github/scripts/select_regression_shard.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python3
+"""Select a balanced file-level shard for regression tests."""
+
+from __future__ import annotations
+
+import argparse
+import ast
+import os
+from pathlib import Path
+
+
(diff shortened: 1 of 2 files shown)


lightzt99:
LGTM

PR #1125, opened 2026-06-30
fix(storage): sanitize SQLite FTS5 queries for technical tokens

### LINKED ISSUE #1119: [Bug]: SQLite FTS5 search fails on hyphenated technical tokens
## Summary

During the v1.1.6 post-release add/search smoke test, SQLite hybrid search logged an FTS5 failure when the query contained a hyphenated technical token such as `oceanbase-blue-116`.

The vector-search side still returned a result, but the full-text side failed and was silently degraded.

## Reproduction

Using `powermem==1.1.6` with SQLite storage and the default 384-dim local embedder:

```python
from powermem import Memory
from powermem.config_loader import create_config

config = create_config(
    database_provider="sqlite",
    database_config={
        "database_path": "/tmp/powermem-postrelease-verify.sqlite",
        "collection_name": "verify_memories",
    },
    llm_provider="noop",
    llm_model="noop",
    embedding_provider="default",
    embedding_model="all-MiniLM-L6-v2",
    embedding_dims=384,
)
config["intelligence"] = {"enabled": False}

memory = Memory(config=config)
memory.add(
    "PowerMem release verification: the marker phrase is oceanbase-blue-116 and it supports add search.",
    user_id="postrelease-user",
)
result = memory.search("oceanbase-blue-116 add search", user_id="postrelease-user", limit=3)
```

## Observed

The search returned one result through vector fallback, but SQLite FTS5 emitted this warning:

```text
WARNING - FTS5 search failed, returning empty: no such column: blue
```

A query without the hyphenated token, for example `PowerMem release verification add search`, returned cleanly without the FTS5 warning.

## Expected

Queries containing common technical tokens should not break the SQLite FTS5 path. Examples include:

- `oceanbase-blue-116`
- `powermem-mcp`
- `fix/release-dispatch-repo`
- `v1.1.6`
- package names, branch names, issue keys, and version strings

## Impact

- Hybrid search quality is reduced because the FTS side returns empty for common developer/coding-agent vocabulary.
- Users may see warnings during normal searches even when vector fallback still returns a result.
- Technical memories often contain package names, branch names, version strings, issue IDs, and other punctuation-heavy terms, so this is likely to occur in real usage.

## Suggested Fix

- Escape or sanitize user input before passing it to SQLite FTS5 `MATCH`.
- Consider a helper such as `escape_fts5_query()` that preserves useful terms while quoting special characters safely.
- If the raw query fails, retry with a sanitized phrase query before falling back to empty FTS results.
- Add tests for hyphen, slash, dot, colon, and mixed version/package tokens.

## Acceptance Criteria

- SQLite search with `oceanbase-blue-116 add search` does not log an FTS5 syntax/column warning.
- SQLite search with `powermem-mcp`, `v1.1.6`, and `fix/release-dispatch-repo` does not break the FTS path.
- Existing vector-only and hybrid search behavior remains compatible.

kic635:
Fixes #1119. Normalize punctuation-heavy search terms before FTS5 MATCH so hybrid search does not silently drop the full-text path or log spurious warnings for pure-punctuation input.

## Summary

SQLite hybrid search passes user queries to FTS5 `MATCH` unchanged. Technical tokens such as `oceanbase-blue-116`, `v1.1.6`, and `fix/release-dispatch-repo` are parsed as FTS5 query syntax (NOT operators, column filters, etc.), causing `OperationalError` and silently dropping the full-text path while vector search still returns results.

This PR adds `_sanitize_fts5_input()` and applies it in `SQLiteVectorStore._fulltext_search()` before `MATCH`, so common developer/coding-agent vocabulary no longer breaks the FTS leg of hybrid search.

Fixes #1119

## Solution Description

1. **Add `_sanitize_fts5_input()`** in `sqlite_vector_store.py`  
   - Strip FTS5 metacharacters and selected ASCII separator characters (`-`, `/`, `.`, `:`, `@`, `|`, etc.) by converting them to spaces, aligning query tokens with unicode61 indexing behavior.  
   - Tokens containing `@` or `|` are handled by stripping to word tokens (e.g. `@username` → `username`, `foo|bar` → `foo bar`) so `MATCH` succeeds without syntax errors.  
   - Fast path: queries with no special characters are returned unchanged.  
   - Pure punctuation input (e.g. `---`, `#@!`) sanitizes to an empty string; `_fulltext_search()` returns `[]` immediately without attempting `MATCH` or logging a warning.

2. **Behavior change (ops note)**  
   Previously, queries like `---` reached FTS5, logged `FTS5 search failed`, then returned `[]`. After this change, such input is rejected at sanitization time and returns `[]` silently.

3. **Tests**  
   - Unit tests in `tests/unit/test_sqlite_fts.py` (`TestFTS5QuerySanitization`)  
   - Integration test in `tests/integration/test_sqlite_fts_integration.py` (`test_search_technical_tokens`)

**Test plan**

- [x] `pytest tests/unit/test_sqlite_fts.py`
- [x] `pytest tests/integration/test_sqlite_fts_integration.py`>

### DIFF
--- a/src/powermem/storage/sqlite/sqlite_vector_store.py
+++ b/src/powermem/storage/sqlite/sqlite_vector_store.py
@@ -41,10 +41,8 @@ def _json_path_for_key(key: str) -> str:
     return path
 
 
-# FTS5 query metacharacters and selected ASCII separator characters stripped before MATCH.
-_FTS5_QUERY_CLEANUP_RE = re.compile(
-    r"""[-/.#:+!&*"()^$%,@|'~{}\[\]=\\?<>;`]+"""
-)
+# ASCII word fragments aligned with unicode61 token characters.
+_FTS5_TOKEN_RE = re.compile(r"[a-zA-Z0-9_]+")
 
 
 def _quote_fts5_literal_token(token: str) -> str:
@@ -54,37 +52,15 @@ def _quote_fts5_literal_token(token: str) -> str:
 
 
 def _sanitize_fts5_input(query: str) -> str:
-    """Sanitize user search text for safe SQLite FTS5 MATCH queries.
+    """Build a literal FTS5 MATCH string from user search text.
 
-    PowerMem callers pass natural-language or technical tokens, not raw FTS5
-    syntax. Separator characters and code punctuation (aligned with common
-    unicode61 tokenization) are converted to spaces, then each remaining token
-    is wrapped in FTS5 double quotes so ``AND``/``OR``/``NOT`` and other
-    reserved words are matched literally rather than as query operators.
-
-    Returns an empty string when the input contains only punctuation or
-    whitespace after cleanup (for example ``---`` or ``#@!``). Purely
-    sanitized punctuation queries therefore return ``[]`` without attempting
-    MATCH.
-
-    Numeric-only tokens (for example ``1119`` from ``#1119``) are quoted but
-    FTS5 typically does not index bare digits; matches still rely on
-    co-occurring word tokens such as ``oceanbase`` or ``blue``.
-
-    FTS5 advanced syntax (phrase quotes, prefix ``*``, column ``:`` filters,
-    boolean operators) is not available on user input. There is no public API
-    to bypass this sanitization.
+    Keeps ASCII alphanumerics and underscore, quotes each token, and drops
+    non-ASCII characters. Returns empty when nothing remains after extraction.
     """
     if not query or not query.strip():
         return ""
 
-    tokens: List[str] = []
-    for raw_token in query.split():
-        token = _FTS5_QUERY_CLEANUP_RE.sub(" ", raw_token)
-        for part in token.split():
-            if part:
-                tokens.append(part)
-
+    tokens = _FTS5_TOKEN_RE.findall(query)
     if not tokens:
         return ""
 
@@ -338,6 +314,9 @@ class SQLiteVectorStore(VectorStoreBase):
 
         fts_query = _sanitize_fts5_input(query)
         if not fts_query:
+            logger.debug(
+                "FTS5 query sanitized to empty, skipping fulltext search"
+            )
             return []
 
         fts_table = f"{self.collection_name}_fts"
--- a/src/powermem/storage/sqlite/sqlite_vector_store.py
+++ b/src/powermem/storage/sqlite/sqlite_vector_store.py
@@ -41,30 +41,37 @@ def _json_path_for_key(key: str) -> str:
     return path
 
 
-# ASCII word fragments aligned with unicode61 token characters.
-_FTS5_TOKEN_RE = re.compile(r"[a-zA-Z0-9_]+")
+_FTS5_MAX_QUERY_TOKENS = 64
 
+# Word fragments aligned with unicode61-style token characters.
+_FTS5_TOKEN_RE = re.compile(r"[\w]+", re.UNICODE)
 
-def _quote_fts5_literal_token(token: str) -> str:
-    """Wrap a single token in FTS5 double quotes for literal (non-syntax) matching."""
-    escaped = token.replace('"', '""')
-    return f'"{escaped}"'
+_FTS5_RESERVED = frozenset({"AND", "OR", "NOT"})
 
 
 def _sanitize_fts5_input(query: str) -> str:
     """Build a literal FTS5 MATCH string from user search text.
 
-    Keeps ASCII alphanumerics and underscore, quotes each token, and drops
-    non-ASCII characters. Returns empty when nothing remains after extraction.
+    Extracts Unicode word tokens, drops FTS5 boolean keywords, and wraps each
+    remaining token in double quotes. Punctuation-separated forms such as
+    ``v1.1.6`` become independent token matches (``"v1" "1" "6"``) without
+    preserving original spacing. Returns empty when nothing remains.
     """
     if not query or not query.strip():
         return ""
 
-    tokens = _FTS5_TOKEN_RE.findall(query)
+    tokens = [
+        token
+        for token in _FTS5_TOKEN_RE.findall(query)
+        if token.upper() not in _FTS5_RESERVED
+    ]
     if not tokens:
         return ""
 
-    return " ".join(_quote_fts5_literal_token(token) for token in tokens)
+    if len(tokens) > _FTS5_MAX_QUERY_TOKENS:
+        tokens = tokens[:_FTS5_MAX_QUERY_TOKENS]
+
+    return " ".join('"' + token.replace('"', '""') + '"' for token in tokens)
 
 
 def _check_sqlite_features(conn) -> None:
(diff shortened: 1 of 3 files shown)


src/powermem/storage/sqlite/sqlite_vector_store.py
```diff
@@ -40,6 +41,52 @@ def _json_path_for_key(key: str) -> str:
     return path
 
 
+# FTS5 query metacharacters and selected ASCII separator characters stripped before MATCH.
+_FTS5_QUERY_CLEANUP_RE = re.compile(
```
wayyoungboy: This cleanup still lets FTS5 syntax leak through. `AND`/`OR`/`NOT` survive as operators, and punctuation such as `<`, `>`, `;`, and backticks is not stripped, so literal technical input like `fix-NOT-working`, `std::vector<int>`, `foo;bar`, or `` `foo` `` can still log `FTS5 search failed` or drop the FTS leg. Since callers pass user text rather than FTS5 syntax, please make the sanitizer construct a literal token query or use an allowlist matching tokenizer output.

tests/unit/test_sqlite_fts.py
```diff
@@ -363,3 +367,185 @@ def test_update_fts_sync_no_stale_entry(self, store):
...
+            ("~1.2.3", "1 2 3"),
+            ("key=value", "key value"),
+            ("my_var", "my_var"),
+        ],
+    )
+    def test_sanitize_fts5_input(self, raw, expected):
+        assert _sanitize_fts5_input(raw) == expected
+
+    def test_sanitize_preserves_boolean_operators(self):
+        assert _sanitize_fts5_input("neural AND network") == "neural AND network"
```
wayyoungboy: This expectation preserves the unsafe behavior above. The tests should assert literal `AND`/`OR`/`NOT` terms and additional code punctuation do not reach FTS5 as syntax, otherwise hybrid search can still look successful through vector fallback while the FTS leg failed.

wayyoungboy (changes requested):
Thanks for the focused fix and the added coverage for the examples from #1119. I can confirm the listed cases now pass without FTS5 warnings, but I think this still needs one more pass before merge.

The sanitizer still leaves parts of user input as FTS5 query syntax. Standalone `AND`/`OR`/`NOT` survive cleanup, so literal technical terms such as `fix-NOT-working` sanitize to `fix NOT working` and drop the FTS result. The blacklist also misses common code punctuation such as `<`, `>`, `;`, and backticks, so inputs like `std::vector<int>`, `foo;bar`, and `` `foo` `` still hit `MATCH` and log `FTS5 search failed`.

Because callers pass natural-language or technical search text rather than FTS5 syntax, I would treat every sanitized token as literal text, or build the query from an allowlist that matches SQLite tokenization, so these cases cannot reach FTS5 as operators or syntax. The current tests cover the original examples, but they also encode the unsafe operator behavior as expected, so please add coverage for literal `AND`/`OR`/`NOT` tokens and the punctuation cases above.

Local validation:

- `pytest tests/unit/test_sqlite_fts.py tests/integration/test_sqlite_fts_integration.py`: 66 passed
- Extra in-memory probes reproduced the remaining warnings for `std::vector<int>`, `foo;bar`, `` `foo` ``, and `fix-NOT-working`.

tests/unit/test_sqlite_fts.py
kic635: @wayyoungboy Agreed — removed the unsafe “preserve boolean operators” expectation.

Test changes

test_sanitize_fts5_input now asserts literal quoted tokens (e.g. neural AND network → "neural" "AND" "network", fix-NOT-working → "fix" "NOT" "working").
Added test_boolean_keywords_quoted_as_literal_tokens_no_warning: end-to-end search for ham AND eggs, standalone AND, and AND OR NOT with no FTS5 search failed warning.
Added coverage for code punctuation (std::vector<int>, foo;bar, backticks) and // empty-input path.
Validation

pytest tests/unit/test_sqlite_fts.py tests/integration/test_sqlite_fts_integration.py -v
# 83 passed
Please take another look when CI is green.

src/powermem/storage/sqlite/sqlite_vector_store.py
kic635: wayyoungboy Thanks for the detailed review — pushed another pass on the sanitizer.

Sanitizer changes

Replaced blacklist cleanup with a Unicode word allowlist ([\w]+, re.UNICODE) and wrap every extracted token in double quotes for literal MATCH.
AND / OR / NOT are quoted like any other token, so they never reach FTS5 as operators (e.g. fix-NOT-working → "fix" "NOT" "working").
Punctuation-separated technical literals split into quoted tokens (e.g. std::vector<int> → "std" "vector" "int", foo;bar → "foo" "bar", `foo` → "foo").
Pure punctuation (//, <>;, !!!) sanitizes to empty; FTS is skipped with a debug log only (no warning).
Validation

pytest tests/unit/test_sqlite_fts.py tests/integration/test_sqlite_fts_integration.py -v
# 83 passed
Please take another look when CI is green.

kic635:
@wayyoungboy Thanks for the follow-up — pushed another pass that switches from blacklist cleanup to literal token MATCH.

Sanitizer changes

Replaced the blacklist with a Unicode word allowlist ([\w]+, re.UNICODE) and wrap every extracted token in double quotes, so user text never reaches FTS5 as operators or raw syntax.
AND / OR / NOT are quoted like any other token (e.g. fix-NOT-working → "fix" "NOT" "working", not fix NOT working; neural AND network → "neural" "AND" "network").
Code punctuation is stripped at tokenization time and only \w+ segments are quoted (e.g. std::vector<int> → "std" "vector" "int", foo;bar → "foo" "bar", `foo` → "foo", <>; → empty).
Standalone // (and other punctuation-only input) sanitizes to empty; FTS is skipped with a debug log only — no FTS5 search failed warning.
Tests

Removed the unsafe “preserve boolean operators” expectation; tests now assert literal quoted tokens end-to-end.
Added parametrize + no-warning search coverage for //, std::vector<int>, foo;bar, backticks, fix-NOT-working, and standalone AND / AND OR NOT.
Integration test uses composite + isolated docs for the technical-token cases from #1119.
Validation

pytest tests/unit/test_sqlite_fts.py tests/integration/test_sqlite_fts_integration.py -v
# 83 passed
The in-memory probes you listed (std::vector<int>, `foo`, foo;bar, fix-NOT-working) now complete without FTS5 warnings on this branch. Please take another look when CI is green.

wayyoungboy (approved):
LGTM. Thanks for the follow-up fix.

I rechecked the latest head and the previous blockers look addressed: the FTS5 input is now tokenized through an allowlist and every token is quoted as literal text, so `AND` / `OR` / `NOT` and code punctuation no longer leak into `MATCH` as query syntax.

Validation:

- `pytest tests/unit/test_sqlite_fts.py tests/integration/test_sqlite_fts_integration.py`: 83 passed
- Extra Memory API and direct SQLiteVectorStore probes covered `fix-NOT-working`, `std::vector<int>`, `foo;bar`, backticks, standalone `AND`, punctuation-only input, filters, and no FTS5 warning cases.
- GitHub checks on this head are green; the remaining skipped jobs appear to be workflow-conditional release/regression jobs.

PR #1129, opened 2026-07-01
feat(plugin): dual-backend fallback (remote primary + local fallback)

lightzt99:
## Summary
- Add optional local fallback server so the Claude Code hook keeps working when the remote PowerMem backend is unreachable. Default read/write goes to remote; on net error / timeout / 5xx, the hook retries on the local fallback.
- When `POWERMEM_FALLBACK_BASE_URL` is unset, behavior is byte-identical to single-backend mode (fast path) — no state file written, no probe.
- Circuit-breaker state persisted to `~/.powermem/fallback-state.json` with up/down TTLs (clamp 5..300s) to avoid probing primary on every hook event.
- `POWERMEM_FALLBACK_DISABLED=1` is a runtime kill switch that decays to primary-only without editing `runtime.env`.
- `init.sh` learns dual mode: `POWERMEM_INIT_FALLBACK_BASE_URL` (non-interactive) or a TTY prompt. `write_runtime_dual` writes both URLs + keys with shell-safe quoting.
- `skills/init/SKILL.md` gains a 4th Question 0 option "Remote + local fallback" with a dedicated onboarding round (fallback URL, port-conflict handling, MCP disabled note).
- `SETUP.md` trimmed (~1076 → ~947 lines): dropped PRE-CHECK / STEP BY STEP / FINAL VALIDATION / SUMMARY sections that were redundant with `init.sh` or conflicted with SKILL.md; E006 now points to `scripts/preload-model.sh`.
- Replay/sync deferred — needs a server-side metadata filter / list endpoint; will follow up separately.

## Env vars
| Var | Purpose | Default |
|---|---|---|
| `POWERMEM_FALLBACK_BASE_URL` | fallback URL; empty = single-backend | empty |
| `POWERMEM_FALLBACK_API_KEY` | fallback API key | empty |
| `POWERMEM_FALLBACK_DISABLED` | runtime kill switch | `0` |
| `POWERMEM_FALLBACK_DOWN_TTL_SECONDS` | "primary down" cache TTL (5..300) | `30` |
| `POWERMEM_FALLBACK_UP_TTL_SECONDS` | "primary up" cache TTL (5..300) | `30` |
| `POWERMEM_FALLBACK_TRIGGER_5XX` | treat 5xx as fallback trigger | `1` |
| `POWERMEM_FALLBACK_LOG_FILE` | fallback event log path | `$DATA_DIR/powermem-hook.log` |
| `POWERMEM_INIT_FALLBACK_BASE_URL` | init.sh non-interactive fallback URL | empty |

## Fallback triggers
| Failure mode | Triggers fallback? |
|---|---|
| client timeout (`context.DeadlineExceeded`) | yes |
| dial refused / DNS / TLS / conn-reset (`net.Error`) | yes |
| HTTP 5xx with `TRIGGER_5XX=1` (default) | yes |
| HTTP 5xx with `TRIGGER_5XX=0` | no |
| HTTP 4xx | no |
| HTTP 2xx | no (success) |

## Known limitations (v1)
- **No replication.** Memories written to the fallback during an outage are not automatically replayed to the primary after recovery; the two backends diverge until you manually reconcile.
- **Stale reads.** Searches during fallback read from the local backend, which may lag behind the remote.
- **No conflict resolution.** No automatic deduplication if both backends receive writes independently.
- **MCP transport not covered.** Fallback applies to the hook's REST calls only, not to MCP server transport; dual mode disables MCP.

## Test plan
- [x] `go test ./...` in `apps/claude-code-plugin/cmd/powermem-hook` — 14 new unit tests (helpers, state TTL, routing, 5xx trigger, kill switch, single-backend fast path)
- [x] `python3 apps/claude-code-plugin/tests/test_runtime_dual.py` — 5 cases including shell-metachar quoting for `write_runtime_dual`
- [x] Existing `make test-claude-hook-docker` regression — single-backend fast path preserved, unaffected
- [x] `make test-claude-hook-dual` — new Docker dual-backend regression (`docker/Dockerfile.claude-hook-dual`, `tests/regression/test_claude_hook_dual_backend.py`, 8 tests: both-up → primary, primary-down → fallback, cached-down skips probe, recovery marks up, 5xx trigger on/off, kill switch, single-backend no state file). Passes in 11s with `--network none`.
- [x] Manual: `POWERMEM_INIT_BASE_URL=http://remote:8001 POWERMEM_INIT_FALLBACK_BASE_URL=http://localhost:8848 sh init.sh` writes dual `runtime.env`; subsequent hook events route to primary, fall back on primary failure, recover after DOWN_TTL.

## Non-goals (v1)
- No background replay daemon (hook is short-lived; daemon is a follow-up)
- No dual-write / multi-primary / load balancing
- No MCP transport fallback
- No exponential backoff within a single hook invocation (one primary + one fallback, then exit)

## Files
- `apps/claude-code-plugin/cmd/powermem-hook/fallback.go` (new) — routing, state, helpers
- `apps/claude-code-plugin/cmd/powermem-hook/fallback_test.go` (new) — Go unit tests
- `apps/claude-code-plugin/cmd/powermem-hook/main.go` — `searchMemoriesForPrompt` / `postMemoryRaw` routed via `doRequestWithFallback`
- `apps/claude-code-plugin/scripts/init.sh` — dual mode branch, `write_runtime_dual`, reconfigure prompt
- `apps/claude-code-plugin/scripts/common.sh` — `write_runtime_dual` helper
- `apps/claude-code-plugin/skills/init/SKILL.md` — 4th Question 0 option + dual onboarding round
- `apps/claude-code-plugin/SETUP.md` — trimmed; new "Remote primary + local fallback" section
- `apps/claude-code-plugin/tests/test_runtime_dual.py` (new) — shell tests
- `tests/regression/test_claude_hook_dual_backend.py` (new) — Docker regression
- `docker/Dockerfile.claude-hook-dual` (new) — regression image
- `Makefile` — `test-claude-hook-dual` target


wayyoungboy:
I found one draft blocker in the dual fallback init flow. `apps/claude-code-plugin/skills/init/SKILL.md` starts the local fallback by passing `POWERMEM_INIT_BASE_URL=<fallback_url>` to `apps/claude-code-plugin/scripts/init.sh`, but the local init path does not derive the port from that URL; it uses `POWERMEM_INIT_PORT` if set, otherwise 8848 if free, otherwise 8849-8899.

If the user chooses `http://localhost:8849` while 8848 is free, init starts the server on 8848 but the next step writes `POWERMEM_FALLBACK_BASE_URL=http://localhost:8849`, so fallback points at a port where nothing was started. Please either pass/derive `POWERMEM_INIT_PORT` from the selected fallback URL, or change the skill/script flow to only write a fallback URL that was actually started and verified.

wayyoungboy:
Thanks for the dual-backend fallback work. I do not think this is ready to merge yet because the fallback init flow can still write a URL that does not match the server that was actually started.

If the user selects a fallback URL such as `http://localhost:8849` while port 8848 is free, the init script may start the local server on 8848 but still persist `POWERMEM_FALLBACK_BASE_URL=http://localhost:8849`. Please either derive/pass `POWERMEM_INIT_PORT` from the selected fallback URL, or only write a fallback URL after that exact endpoint has been started and verified.

Teingi:

PR #1130, opened 2026-07-01
fix(config): honor disabled graph store flag

### LINKED ISSUE #1118: [Bug]: graph_store enabled=false still initializes OceanBase graph store
## Summary

When using a dict config, setting `graph_store` to `{"enabled": false}` can still cause PowerMem to initialize the default OceanBase graph store. This was found during the v1.1.6 post-release smoke test while trying to run an isolated `sqlite + noop LLM + default embedding` add/search verification.

## Reproduction

In a clean environment with `powermem==1.1.6` installed:

```python
from powermem import Memory
from powermem.config_loader import create_config

config = create_config(
    database_provider="sqlite",
    database_config={
        "database_path": "/tmp/powermem-postrelease-verify.sqlite",
        "collection_name": "verify_memories",
    },
    llm_provider="noop",
    llm_model="noop",
    embedding_provider="default",
    embedding_model="all-MiniLM-L6-v2",
    embedding_dims=384,
)
config["intelligence"] = {"enabled": False}
config["graph_store"] = {"enabled": False}

memory = Memory(config=config)
```

## Observed

`Memory(config=config)` fails during initialization:

```text
ValueError: embedding_model_dims is required for OceanBase graph operations. Please configure embedding_model_dims in your OceanBaseGraphConfig.
```

The failure indicates that the graph store is still considered enabled and the default OceanBase graph provider is being initialized, even though the caller explicitly passed `enabled: false`.

## Expected

`graph_store: {"enabled": false}` should disable graph initialization completely, matching the behavior users would naturally expect from the `enabled` flag.

At minimum, PowerMem should reject this shape with a clear validation error instead of silently treating it as an enabled graph-store config.

## Impact

- No-LLM / SQLite smoke tests are harder to configure reliably.
- Users may reasonably assume `enabled: false` disables graph features, but the process can fail while initializing an unrelated OceanBase graph store.
- This is especially confusing because omitting `graph_store` entirely allows the same `sqlite + noop LLM` add/search flow to work.

## Suggested Fix

- Normalize `graph_store` before creating `MemoryConfig`, or teach `MemoryConfig` / `_get_graph_enabled()` to honor `enabled: false` consistently.
- Ensure both dict mode and `MemoryConfig` mode treat disabled graph config the same way.
- Consider documenting the canonical way to disable graph store.

## Acceptance Criteria

- `Memory(config={..., "graph_store": {"enabled": False}})` does not initialize a graph store.
- The same configuration can complete a simple add/search smoke test with SQLite.
- Add regression coverage for dict config and `MemoryConfig` config paths.

knqiufan:
## Summary
- Normalize graph_store disabled dicts to disabled MemoryConfig state so they do not initialize the default OceanBase graph store.
- Respect explicit graph_store.enabled in both Memory and AsyncMemory before falling back to provider-config presence.
- Add regression coverage for dict config, MemoryConfig, async initialization, env-disabled graph config, and SQLite + noop LLM add/search smoke.

Closes #1118.

## Test plan
- [x] uv run --python 3.12 --no-project --with-editable .[dev,server] pytest tests/unit/test_issue_1118_graph_store_disabled.py -v
- [x] uv run --python 3.12 --no-project --with-editable .[dev,server] pytest tests/unit/test_config_loader.py -v
- [x] uv run --python 3.12 --no-project --with-editable .[dev,server] pytest tests/integration/test_noop_llm_mode.py -v
- [ ] Broader non-e2e suite attempted locally on Windows; it still has unrelated failures in Claude plugin shell subprocess tests, ModelScope cache-dependent tests, and release binary subprocess tests. The #1118 regression tests are green.

### DIFF
--- a/scripts/build_binary_package.sh
+++ b/scripts/build_binary_package.sh
@@ -18,6 +18,10 @@ else
     --with pyinstaller
     --with "setuptools<81"
     --with wheel
+    # Pillow 12.2.1+ stopped publishing manylinux_2_17 wheels, so the CentOS 7
+    # binary build (glibc 2.17) falls back to a source build that fails without
+    # libjpeg/zlib headers. Pin to the last release with a manylinux_2_17 wheel.
+    --with "pillow<12.2.1"
     python
   )
 fi
--- a/src/powermem/configs.py
+++ b/src/powermem/configs.py
@@ -38,6 +38,33 @@ def _default_vector_store_config() -> BaseVectorStoreConfig:
     return OceanBaseConfig()
 
 
+def _is_disabled_graph_store_flag(value: Any) -> bool:
+    if value is None:
+        return False
+    if isinstance(value, bool):
+        return value is False
+    if isinstance(value, str):
+        return value.strip().lower() in {"false", "0", "no", "off", "disabled"}
+    if isinstance(value, int):
+        return value == 0
+    return False
+
+
+_GRAPH_STORE_WRAPPER_KEYS = {
+    "enabled",
+    "provider",
+    "config",
+}
+
+_GRAPH_STORE_METADATA_KEYS = {
+    "llm",
+    "custom_prompt",
+    "custom_extract_relations_prompt",
+    "custom_update_graph_prompt",
+    "custom_delete_relations_prompt",
+}
+
+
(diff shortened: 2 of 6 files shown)


wayyoungboy (approved):
LGTM

PR #1132, opened 2026-07-01
test(storage): add issue #1093 filter key collision regression tests

### LINKED ISSUE #1093: Add coverage for metadata and top-level filter key collisions
Follow-up from review of PR #1052.

## Summary
Add regression coverage that documents the intended behavior when user metadata keys collide with promoted or top-level payload fields used by search filters.

## Suggested coverage
- Plain metadata filters versus explicit metadata-prefixed filters for keys such as hash, data, created_at, updated_at, and category.
- SQLite search, list, and count paths.
- OceanBase-like and PGVector-like adapter filter translation paths.

## Acceptance criteria
- Tests make the intended filter semantics explicit for colliding keys.
- Existing public filter behavior is either preserved or intentionally documented with updated tests.

kic635:
Refactor list post-filter matching with _normalized_filter_value and store-aware _memory_matches_filter, restore OceanBase metadata fallback for bare unknown keys, and keep public search/list return shapes unchanged.

## Summary

Add regression tests for filter key collisions between top-level payload fields and nested `metadata` (issue #1093), and refactor `StorageAdapter` list post-filter matching so it follows the same store-specific key translation as `_build_db_filters`.

Fixes a regression where OceanBase bare unknown keys (e.g. `{"scope": "personal"}`) were incorrectly rejected during list post-filtering. Public `search_memories` / `get_all_memories` return shapes are unchanged.

Fix #1093

### Adapter changes (`src/powermem/storage/adapter.py`)

- Add `_normalized_filter_value()` with a `_FILTER_VALUE_MISSING` sentinel to distinguish absent keys from `None` values.
- Refactor `_memory_matches_filter()` to reuse `_metadata_filter_key_for_store()`:
  - `metadata.*` filters always match nested metadata first (e.g. OceanBase `metadata.hash` vs top-level `hash`).
  - Other keys are translated per store, then resolved via `_normalized_filter_value()`.
  - For bare unknown keys outside `_PAYLOAD_FILTER_KEYS` / `_SYSTEM_FILTER_KEYS`, restore metadata fallback when the top-level lookup is missing (OceanBase legacy behavior).
- Add `_filter_match_view_from_payload()` as an internal-only row shape for list post-filtering, without changing public API return fields.
- Document that `hash` / `data` are intentionally excluded from `_PAYLOAD_FILTER_KEYS`; use `payload.*` or `metadata.*` for explicit cross-store semantics.

### Test coverage (`tests/unit/test_issue_1093_filter_key_collisions.py`)

- SQLite: end-to-end `count` / `list` / `search` for collision keys.
- OceanBase / PGVector: `_build_db_filters` translation and `_memory_matches_filter` via lightweight backend stubs.
- Unit cases for bare `hash` / `data` (SQLite vs OceanBase), OceanBase `metadata.hash` preference, and bare unknown-key metadata fallback.

### Test plan

```bash
python -m pytest tests/unit/test_issue_1093_filter_key_collisions.py tests/unit/test_list_memory_filters.py -v

### DIFF
--- a/src/powermem/storage/adapter.py
+++ b/src/powermem/storage/adapter.py
@@ -15,11 +15,21 @@ from powermem.utils.utils import serialize_datetime, get_current_datetime
 
 logger = logging.getLogger(__name__)
 
+_FILTER_VALUE_MISSING = object()
+
 
 class StorageAdapter:
     """Adapter that bridges VectorStoreBase interface with Memory class expectations."""
 
     _SYSTEM_FILTER_KEYS = {"user_id", "agent_id", "run_id"}
+    # Promoted payload fields addressed by bare filter keys on nested stores
+    # (SQLite / PGVector). On OceanBase-like stores bare keys map to the top-level
+    # payload instead; see _metadata_filter_key_for_store().
+    #
+    # hash/data are intentionally excluded from this set: on nested stores unqualified
+    # hash/data filters target metadata.{key}; on OceanBase they target the top-level
+    # payload field. Use payload.hash / payload.data or metadata.hash / metadata.data
+    # when you need explicit cross-store semantics.
     _PAYLOAD_FILTER_KEYS = {
         "actor_id",
         "category",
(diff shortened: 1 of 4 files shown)


wayyoungboy:
I did not find a code-level blocker in the adapter/test changes themselves, but this PR is still blocked by the failing CentOS 7 binary build check. The failure is in the dependency build path for Pillow 12.3.0, where the source build requires C99 support. Please rebase/sync the packaging fix that avoids this CI failure, or otherwise get the binary build check green before merge.

kic635:
Thanks for the review! I've rebased this branch onto the latest `main` and synced the packaging fix from #1130 (`pillow<12.2.1` in `scripts/build_binary_package.sh`). The PR head is now at `174daec`, and all checks are green. Please take another look when you have a moment.

wayyoungboy (changes requested):
Requesting changes because the OceanBase execution path still does not honor the explicit collision-key semantics this PR is trying to document.

1. `metadata.<column>` filters are stripped before they reach OceanBase, so real list/count/search calls can filter on the top-level column and drop the row before `_memory_matches_filter()` ever sees it. For example, `metadata.hash` is translated to `hash`; OceanBase then resolves `hash` as the table column first, not nested JSON metadata. A row with top-level `hash="payload-hash"` and nested `metadata.hash="metadata-hash"` is therefore missed by `filters={"metadata.hash": "metadata-hash"}` in `get_all_memories()`, `count_all_memories()`, and `search_memories()`. The same problem applies to other column-backed collision keys such as `metadata.category`, `metadata.created_at`, and `metadata.updated_at`.

2. `payload.data` / bare `data` also do not target OceanBase's stored text content. The adapter translates them to `data`, but the OceanBase text column is `document`; because `data` is not a table column, OceanBase falls back to metadata JSON lookup instead. That makes `filters={"payload.data": "<memory text>"}` miss the row unless user metadata also happens to contain the same `data` value.

The new tests cover SQLite end-to-end and adapter-level OceanBase stubs, but they do not exercise OceanBase's actual `_generate_where_clause()` column-first behavior. I verified the declared tests pass locally and CI is green, but this gap means issue #1093's collision semantics are still not reliably covered for OceanBase-like stores.

kic635:
Thanks for the detailed and careful review — sorry I missed the OceanBase execution-path issues in the first pass. The follow-up commit 6961fee addresses each point below.

metadata.<column> filters are no longer stripped before they reach OceanBase.
_metadata_filter_key_for_store() now preserves the metadata.* prefix on OceanBase-like stores, so metadata.hash stays metadata.hash instead of being translated to hash. Plain collision keys (hash, category, created_at, updated_at, etc.) still map to top-level table columns as intended.

In oceanbase.py, _generate_where_clause() → get_column() now checks for the metadata. prefix before resolving a table column. So metadata.hash targets nested JSON (metadata ->> '$.hash'), while hash still targets the top-level hash column. A row with top-level hash="payload-hash" and nested metadata.hash="metadata-hash" is now correctly matched by filters={"metadata.hash": "metadata-hash"} in get_all_memories(), count_all_memories(), and search_memories(). The same fix applies to other column-backed collision keys such as metadata.category, metadata.created_at, and metadata.updated_at.

payload.data / bare data now target OceanBase's stored text content.
The adapter maps payload.data and bare data to the OceanBase text field (document by default via text_field), instead of passing data through. Because document is a real table column, _generate_where_clause() resolves it directly rather than falling back to metadata JSON lookup. filters={"payload.data": "<memory text>"} now hits the actual stored text content.

Tests now exercise OceanBase's actual _generate_where_clause() column-first behavior.
Added regression tests against the real table schema (mock table, no live DB):

test_oceanbase_where_clause_metadata_hash_targets_json_not_hash_column — verifies metadata.hash compiles to JSON lookup, distinct from top-level hash
test_oceanbase_where_clause_document_column_for_text_content — verifies document compiles to the text column, distinct from metadata.data
test_oceanbase_adapter_db_filters_use_document_for_payload_data — verifies adapter translation produces document + preserved metadata.hash
These go beyond adapter-level stubs and validate the compiled SQL path that list/count/search actually use. All 43 related unit tests pass locally.

We don't have a live OceanBase instance in CI; if the project has an OB test environment, I'm happy to add an integration test in a follow-up.

Sorry again for the oversight — your review helped catch a real gap. Please take another look when you have a chance.

wayyoungboy (approved):
Re-reviewed the current head. The previous OceanBase filter concerns are addressed: `metadata.*` now reaches the OceanBase SQL generator as an explicit metadata JSON lookup before column resolution, and `payload.data` / bare `data` map to the stored text field instead of metadata JSON.

I also verified the new regression coverage locally: `tests/unit/test_issue_1093_filter_key_collisions.py` and `tests/unit/test_list_memory_filters.py` both pass together (43 tests). No further blocker from my side.

wayyoungboy:
LGTM

PR #1136, opened 2026-07-03
fix(server): sanitize startup errors in API responses

### LINKED ISSUE #1133: Service startup errors are exposed in API responses
## Evidence
Recent commit `3b282aa7fb4f487eeb1b90814d3297aa9b69c150` (`feat: auto-fallback to SQLite when embedded SeekDB is unavailable (#993)`) added service readiness handling that stores and returns raw startup exceptions:

- `src/server/main.py:92-94` catches service initialization failures and stores `str(e)` in `app.state.service_startup_error`.
- `src/server/utils/service_errors.py:10-12` appends that raw `service_startup_error` to the user-facing 503 message used by memory/search/user/agent endpoints.
- `src/server/middleware/error_handler.py:46-52` serializes `APIError.to_dict()` unchanged into the HTTP response body.
- `src/server/api/v1/system.py:97`, `:128`, and `:151` also place the raw startup error into `/api/v1/system/status` response data and dependency details.

The same commit adds `tests/unit/server/test_system_health.py`, which confirms `/api/v1/system/health` does not expose startup errors, but the authenticated status endpoint and the service-unavailable paths are not covered.

## Why this is a bug
Startup exceptions from storage/backend initialization can include credentials, internal hostnames, DSNs, or local filesystem paths. For example, a synthetic failure string like `failed to connect to postgresql://user:SYNTHETIC_SECRET@db.example/powermem from /synthetic/local/path` would be copied into API response JSON by the paths above.

## Minimal safe fix
Keep the raw startup exception in logs/internal state only, and return a generic public message in API responses. Concretely:

- Remove `str(startup_error)` from `service_unavailable_message()`.
- Do not populate `StatusResponse.startup_error` or `DependencyStatus.error_message` with the raw exception; use a generic startup failure message plus the existing storage recommendation.
- Add regression tests with a synthetic DSN/path secret for `/api/v1/system/status` and one service-unavailable endpoint, similar to the existing health test.

No real secrets or local environment details are included here.

knqiufan:
## Summary
- Stop appending raw service startup exceptions to user-facing 503 responses.
- Return sanitized startup failure summaries from `/api/v1/system/status` and its fallback path while keeping raw errors internal.
- Add regression coverage for synthetic DSN/secret/path exposure in health, status, fallback, and service-unavailable responses.

## Test plan
- `python -m pytest tests/unit/server/test_system_health.py`
- `python -m pytest tests/unit/server`
- `git diff --check`
- `python -m pytest tests/unit` was also attempted locally; it fails on unrelated local environment/tool assumptions, including existing API/model environment overrides, missing ModelScope cache, and Windows script/tool availability.

Fixes #1133

### DIFF
--- a/src/server/api/v1/system.py
+++ b/src/server/api/v1/system.py
@@ -18,6 +18,10 @@ from ...middleware.rate_limit import limiter, get_rate_limit_string
 from ...config import config
 from ...utils.metrics import get_metrics_collector
 from ...utils.health_check import check_all_dependencies
+from ...utils.service_errors import (
+    public_startup_error_message,
+    public_startup_error_with_recommendation,
+)
 from powermem import auto_config
 from powermem.version import __version__ as powermem_version
 
@@ -94,7 +98,9 @@ async def get_status(
         dependencies["memory_service"] = DependencyStatus(
             name="memory_service",
             status="healthy" if memory_service_ready else "unavailable",
-            error_message=(str(startup_error)[:200] if startup_error else None),
+            error_message=(
+                public_startup_error_with_recommendation() if startup_error else None
+            ),
             last_checked=datetime.utcnow(),
         )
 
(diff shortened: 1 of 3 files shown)


wayyoungboy (approved):
Reviewed the startup-error redaction changes. The raw startup exception is no longer included in /system/status, memory_service dependency details, or service-unavailable APIError messages; those paths now use a fixed public message while keeping the operational recommendation.

Validation:
- GitHub checks are green, including test (3.11) and test (3.12).
- Locally ran a focused FastAPI TestClient regression over health, status normal/fallback, and 503 APIError serialization using a synthetic DSN/path startup error; the sensitive fragments stayed out of the JSON responses.

LGTM.

PR #1138, opened 2026-07-03
fix(server): sanitize runtime exception responses

### LINKED ISSUE #1137: Runtime dependency check and request-path errors still expose raw exceptions in API responses
## Evidence

Issue #1133 fixed the startup-error leak path (commit ea1095f), but the same vulnerability class persists in two other code paths that still echo `str(e)` verbatim into public API responses.

**1. Authenticated `/api/v1/system/status` — runtime dependency probe**

`src/server/utils/health_check.py` runs `Memory()` / `auto_config()` initialization lazily during status checks, and on failure stores `str(e)` directly into `DependencyStatus.error_message`:

- `src/server/utils/health_check.py:131` — `_check_database_sync()`
- `src/server/utils/health_check.py:224` — `_check_llm_sync()`

These statuses are returned by `check_all_dependencies()` and merged into the response at `src/server/api/v1/system.py:97-105`. Any authenticated caller of `/api/v1/system/status` can read `dependencies.database.error_message` and `dependencies.llm.error_message`.

**2. Request-path service errors**

`APIError` messages throughout the service layer interpolate raw exceptions:

- `src/server/services/memory_service.py` (multiple sites: `:503`, `:563`, `:626`, `:1370`, `:1450`, `:1762`, plus `:368`, `:1566`, `:1651` for error dicts)
- `src/server/services/user_service.py` (`:67`, `:153`, `:215`, `:266`, `:323`, `:381`, `:420`)
- `src/server/services/agent_service.py` (`:80`, `:165`, `:315`)
- `src/server/services/search_service.py:112`
- `src/server/api/v1/system.py:246` — `delete-all-memories`

## Why this is a bug

Storage-layer exceptions surfaced via these paths can include DSNs, credentials, internal hostnames, or filesystem paths. The same synthetic example from #1133 applies: a connection string like

`failed to connect to postgresql://user:SYNTHETIC_SECRET@db.example/powermem from /synthetic/local/path`

would be echoed into the JSON response by the paths above.

The `/status` endpoint is arguably worse than the startup path #1133 patched:

- Status can be queried repeatedly while startup fires once.
- An operator reading the startup traceback in logs has full context; an API consumer only needs valid auth to read the dependency `error_message` field.

## Minimal safe fix

- In `health_check.py`, stop storing `str(e)` in `DependencyStatus.error_message`. Reuse the helpers added in #1133 (e.g. `public_startup_error_with_recommendation()`) or add a `dependency_unavailable_message(name)` that does not interpolate `e`. Log the raw exception via `logger.exception(...)` so operators don't lose diagnostics.
- For request-path service errors, stop interpolating `str(e)` into user-visible `APIError.message`. Return a generic public message; keep the raw exception in logs only.
- Add regression tests modeled on `tests/unit/server/test_system_health.py` that seed a synthetic DSN/path secret into a database/LLM probe failure (and into a service-layer call failure) and assert it does not appear in the response body.

## Scope note

Same vulnerability class as #1133 but distinct code paths. Filed separately so each fix can be reviewed independently. No real secrets included.

knqiufan:
## Summary
- Sanitize runtime dependency probe failures before they reach `/api/v1/system/status`, and keep raw probe exceptions in server logs only.
- Replace request-path service `APIError` messages and batch item errors that previously embedded `str(e)` with generic public messages.
- Add regression coverage for synthetic DSN/secret/path exposure in dependency checks, status dependencies, service-layer errors, and batch response data.

## Stack note
- This branch is stacked on #1136 because #1137 reuses the safe startup-error helpers added there.
- The #1137-specific commit is `a37bb5d` (`fix(server): sanitize runtime exception responses`). After #1136 merges, this PR diff should collapse to the #1137 changes only.

## Test plan
- `python -m pytest tests/unit/server/test_health_check.py tests/unit/server/test_system_health.py`
- `python -m pytest tests/unit/server`
- `git diff --check`
- `python -m pytest tests/unit` was attempted locally; it still fails on unrelated local environment/tool assumptions, including API/model environment overrides, missing ModelScope cache, Windows script/tool availability, and an existing sqlite FTS setup error.

Fixes #1137

### DIFF
--- a/src/server/api/v1/system.py
+++ b/src/server/api/v1/system.py
@@ -2,6 +2,8 @@
 System management API routes
 """
 
+import logging
+
 from fastapi import APIRouter, Depends, Request, Response, Query
 from slowapi import Limiter
 from typing import Optional
@@ -19,6 +21,8 @@ from ...config import config
 from ...utils.metrics import get_metrics_collector
 from ...utils.health_check import check_all_dependencies
 from ...utils.service_errors import (
+    dependency_unavailable_message,
+    operation_failed_message,
     public_startup_error_message,
     public_startup_error_with_recommendation,
 )
@@ -29,6 +33,14 @@ from powermem.version import __version__ as powermem_version
 from ...state import SERVER_START_TIME
 
 router = APIRouter(prefix="/system", tags=["system"])
+logger = logging.getLogger("server")
+
+
+def _dependency_status_response(name: str, dependency: DependencyStatus) -> dict:
+    data = dependency.model_dump(mode='json')
+    if name in {"database", "llm"} and data.get("error_message"):
+        data["error_message"] = dependency_unavailable_message(name)
+    return data
 
 
 @router.get(
@@ -121,7 +133,7 @@ async def get_status(
         
         # Convert dependencies to dict for response
         dependencies_dict = {
-            name: dep.model_dump(mode='json')
+            name: _dependency_status_response(name, dep)
             for name, dep in dependencies.items()
         }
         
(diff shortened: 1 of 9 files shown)


wayyoungboy:
LGTM

Teingi:

PR #1142, opened 2026-07-04
fix(intelligence): unify rule-based importance scoring

### LINKED ISSUE #1141: [Enhancement]: Unify rule-based importance evaluation with six-dimension weighted scoring
## Summary

`ImportanceEvaluator` defines six importance dimensions with `criteria_weights`, and the LLM parsing path already synthesizes a weighted total from `criteria_scores` via `_synthesize_from_criteria()`. However, the rule-based fallback path and the breakdown API still use a separate, inconsistent scoring model.

These two gaps should be addressed together so all importance scoring paths share the same six-dimension weighted formula.

## Current State

**File:** `src/powermem/intelligence/importance_evaluator.py`

### 1. Rule-based fallback does not use six-dimension weighting

`_rule_based_evaluation()` (approx. L91–154) still uses a standalone heuristic:

- Content length thresholds
- English-only keyword matching (`important`, `critical`, `remember`, etc.)
- Half-width `?` / `!` punctuation
- `metadata.priority`, tags, and context engagement flags

It does **not** call the existing `_evaluate_*` dimension methods and does **not** apply `criteria_weights`.

By contrast, the LLM path already supports weighted synthesis in `_parse_importance_response()` → `_synthesize_from_criteria()` (see tests in `tests/unit/intelligence/test_importance_evaluator.py`).

This fallback is used when:

- No LLM is configured
- LLM is disabled (`is_noop`)
- LLM response parsing fails
- LLM call raises an exception

### 2. `get_importance_breakdown()` omits weighted total

`get_importance_breakdown()` (approx. L212–246) iterates over `criteria_weights` only to dispatch `_evaluate_*` calls. It returns raw per-dimension scores and:

- Does not include a `weighted_total`
- Does not use the `weight` variable in any calculation
- Is not called by the main `evaluate_importance()` path

## Expected

1. **Single scoring model:** Rule-based evaluation should compute importance as `Σ w_i × score_i` using the same dimension scorers and weights as the LLM criteria synthesis path.
2. **Shared implementation:** `_rule_based_evaluation()` and `get_importance_breakdown()` should reuse the same dimension evaluation helpers, avoiding two parallel algorithms.
3. **Breakdown completeness:** `get_importance_breakdown()` should either return a `weighted_total` field or document explicitly that it returns dimension details only (prefer returning `weighted_total` for consistency).
4. **Better non-LLM coverage:** Extend keyword lists and punctuation handling for Chinese content and full-width punctuation where practical, so fallback scoring is not biased toward English input.

## Impact

- Without LLM (or on LLM fallback), Chinese memories are often under-scored because the rule path ignores the six-dimension model.
- Product/docs narrative around six-dimension importance (e.g. forgetting-curve engineering docs) is inconsistent with the actual fallback behavior.
- Callers of `get_importance_breakdown()` must manually recompute weighted totals, even though weights are already configured in the evaluator.

## Suggested Fix

1. Refactor `_rule_based_evaluation()` to:
   - Call `_evaluate_relevance`, `_evaluate_novelty`, `_evaluate_emotional_impact`, `_evaluate_actionable`, `_evaluate_factual`, and `_evaluate_personal`
   - Aggregate with `criteria_weights` (reuse `_synthesize_from_criteria()` or an equivalent shared helper)
2. Update `get_importance_breakdown()` to include `weighted_total` computed from the same weights.
3. Optionally wire `evaluate_importance()` to reuse the shared aggregation helper to prevent future drift.
4. Add unit tests covering:
   - Rule-based path uses weighted six-dimension scoring
   - Chinese / full-width punctuation inputs are handled reasonably
   - `get_importance_breakdown()` returns consistent `weighted_total`

## Acceptance Criteria

- `_rule_based_evaluation()` produces scores via weighted six-dimension aggregation, not the legacy heuristic sum.
- `get_importance_breakdown()` returns per-dimension scores plus a `weighted_total` consistent with rule-based and LLM synthesis paths.
- Existing LLM parsing tests continue to pass; new tests cover the rule-based and breakdown paths.
- Fallback behavior is documented as aligned with the six-dimension importance model.

## Contributor Note

I am willing to submit a PR implementing the consolidation.

knqiufan:
## Summary
- Closes #1141.
- Refactors rule-based importance evaluation to reuse the six-dimension weighted scoring model.
- Adds `weighted_total` to importance breakdown and shares weighted aggregation with LLM criteria synthesis.
- Expands rule-based keyword signals for Chinese content and full-width punctuation.

## Test plan
- `python -m pytest tests/unit/intelligence/test_importance_evaluator.py`
- `python -m pytest tests/unit/intelligence`
- `git diff --check`

### DIFF
--- a/src/powermem/intelligence/importance_evaluator.py
+++ b/src/powermem/intelligence/importance_evaluator.py
@@ -105,53 +105,9 @@ class ImportanceEvaluator:
         Returns:
             Importance score between 0 and 1
         """
-        score = 0.0
-        
-        # Length factor
-        if len(content) > 100:
-            score += 0.1
-        elif len(content) > 50:
-            score += 0.05
-        
-        # Keyword importance
-        important_keywords = [
-            "important", "critical", "urgent", "remember", "note",
-            "preference", "like", "dislike", "hate", "love",
-            "password", "secret", "private", "confidential"
-        ]
-        
-        content_lower = content.lower()
-        for keyword in important_keywords:
-            if keyword in content_lower:
-                score += 0.1
-        
-        # Question factor
-        if "?" in content:
-            score += 0.05
-        
-        # Exclamation factor
-        if "!" in content:
-            score += 0.05
-        
-        # Metadata factors
-        if metadata:
-            if metadata.get("priority") == "high":
-                score += 0.2
-            elif metadata.get("priority") == "medium":
-                score += 0.1
-            
-            if metadata.get("tags"):
-                score += 0.05
-        
-        # Context factors
-        if context:
-            if context.get("user_engagement") == "high":
-                score += 0.1
-            elif context.get("user_engagement") == "medium":
-                score += 0.05
-        
-        # Cap the score at 1.0
-        return min(score, 1.0)
+        dimension_scores = self._compute_dimension_scores(content, metadata, context)
+        score = self._weighted_dimension_total(dimension_scores)
+        return score if score is not None else 0.0
     
     def _llm_based_evaluation(
         self,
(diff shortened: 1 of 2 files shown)


wayyoungboy (changes requested):
Requesting changes because the current rule-based scoring path drops existing metadata/context importance signals.

The new rule-based path delegates to dimension scoring, but that scoring path does not appear to consume the existing `metadata.priority`, `metadata.tags`, or `context.user_engagement` signals. A direct probe on the current head with neutral content plus high-priority/profile/high-engagement metadata/context returns `0.0`; the previous rule fallback assigned a positive score from those signals. The new tests patch the dimension scorer, so they do not exercise the real evaluator behavior.

Please carry the metadata/context boosts into the real dimension scoring path, or keep the old signal path covered by a test that calls the real evaluator instead of patching it.

knqiufan:
@wayyoungboy Thanks for the review! Fixed in 0d0da9a.

The rule-based path now carries the existing metadata/context signals into the real dimension scoring path:
- `metadata.priority` and `context.user_engagement` contribute through relevance.
- `metadata.tags` contributes through novelty.
- Added a regression test that calls the real `evaluate_importance` path with neutral content plus high-priority/profile/high-engagement metadata/context and verifies the legacy positive score is preserved without patching the scorer.

Verified:
- `python -m pytest tests/unit/intelligence/test_importance_evaluator.py`
- `python -m pytest tests/unit/intelligence`
- `git diff --check`

wayyoungboy:
Thanks for the follow-up. This is still not ready to merge while the previous `CHANGES_REQUESTED` review is active.

Please request re-review after confirming that the rule-based path carries the existing metadata/context importance signals into the six-dimension scoring path, with regression coverage for those signals. Once that review is cleared, this can be reconsidered for merge.

Teingi:

PR #1146, opened 2026-07-05
fix(agent): normalize permission checks

### LINKED ISSUE #1144: [Bug]: PermissionController.check_permission() fails role/default lookup due to enum-vs-string type mismatch
## Summary

`PermissionController.check_permission()` uses direct `in` comparison between an `AccessPermission` enum object and the string-based `role_permissions` lists loaded from config. Since Python enum instances do not match their `.value` strings in `in` checks, role-based and default permission lookups silently return `False`, effectively denying all role/default access.

Additionally, the public `MultiAgentMemoryManager.check_permission()` wrapper converts incoming string permissions with `.upper()`, but `AccessPermission` enum values are lowercase (e.g. `"read"`, `"write"`), causing `ValueError` on every call.

### 1. `PermissionController.check_permission()` — enum vs string

```python
# src/powermem/agent/components/permission_controller.py ~L119-131

# Check role-based permissions
if agent_id in self.agent_roles:
    for role in self.agent_roles[agent_id]:
        if role in self.role_permissions:
            if permission in self.role_permissions[role]:   # <-- enum vs string list
                ...
                return True

# Check default permissions
if "public" in self.role_permissions:
    if permission in self.role_permissions["public"]:       # <-- same issue
        ...
        return True
```

When config defines `default_permissions = {"owner": ["read", "write", ...], "public": ["read"]}`, these are **string lists**. But `permission` is an `AccessPermission` enum (e.g. `AccessPermission.READ` with `.value == "read"`). In Python, `AccessPermission.READ in ["read", "write"]` evaluates to `False`.

**Result:** Role-based and default permission checks always fail; only direct per-memory grants work.

### 2. `MultiAgentMemoryManager.check_permission()` — `.upper()` mismatch

```python
# src/powermem/agent/implementations/multi_agent.py ~L1104-1125

def check_permission(self, agent_id, memory_id, permission):
    try:
        if isinstance(permission, AccessPermission):
            permission_enum = permission
        else:
            permission_enum = AccessPermission(permission.upper())   # <-- "read" -> "READ"
        return self.permission_controller.check_permission(...)
    except (ValueError, Exception):
        return False
```

`AccessPermission` values are lowercase (`"read"`, `"write"`, `"delete"`, `"share"`, `"admin"`). Passing `.upper()` (e.g. `"READ"`) to the enum constructor raises `ValueError`, which is caught and returns `False`.

**Result:** String-based permission checks from external callers (SDK, HTTP API) always fail.

### 3. Missing `revoke_all_for_memory()` method

When a memory is deleted, there is no method to bulk-clear all permission records for that memory from `self.memory_permissions`. The current `revoke_permission()` only handles one agent+permission pair at a time and requires knowing which agents have grants.

## Expected

1. Role-based and default permission lookups should compare the enum's `.value` string against the config list, not the enum object itself.
2. The public `check_permission()` wrapper should normalize to `.lower()` to match `AccessPermission` enum values.
3. A `revoke_all_for_memory(memory_id)` method should exist for cleanup during memory deletion.

### Fix 1: Add a normalization helper in `permission_controller.py`

```python
def _permission_value(permission: AccessPermission) -> str:
    """Normalize permission to lowercase string value for role list comparison."""
    if isinstance(permission, AccessPermission):
        return permission.value
    return str(permission).lower()
```

Then replace `permission in self.role_permissions[role]` with `_permission_value(permission) in self.role_permissions[role]` (same for default check and `validate_permission_chain()`).

### Fix 2: Change `.upper()` to `.lower()` in `multi_agent.py`

```python
permission_enum = AccessPermission(str(permission).lower())
```

### Fix 3: Add bulk revocation

```python
def revoke_all_for_memory(self, memory_id: str) -> None:
    """Remove all permission records for a memory."""
    if memory_id in self.memory_permissions:
        del self.memory_permissions[memory_id]
```

## Impact

- **All role-based and default permission checks are broken** — agents without direct per-memory grants are always denied.
- **SDK / HTTP API callers** passing string permissions are always denied.
- **Memory deletion** leaves stale permission records in `memory_permissions` dict.
- Scope-based access partially masks the issue since `ScopeController.check_scope_access()` also grants access independently, but any flow relying solely on RBAC is broken.

## Acceptance Criteria

- [ ] `PermissionController.check_permission()` correctly matches enum against string-based config lists.
- [ ] `MultiAgentMemoryManager.check_permission("agent_1", "mem_1", "read")` returns `True` when agent has role with `"read"` permission.
- [ ] `revoke_all_for_memory()` or equivalent bulk cleanup exists and is called during `delete_memory()`.
- [ ] Unit tests cover: role permission match, default permission match, string input normalization, bulk revocation.

knqiufan:
## Summary
- Normalize enum permissions before comparing against string role/default permission lists.
- Add bulk permission revocation for memory cleanup.
- Preserve int/string memory-id compatibility for direct permission records.

Fixes #1144

## Test plan
- pytest tests/unit/agent/test_permission_controller.py tests/unit/agent/test_memory_id_normalize.py
- pytest tests/unit/agent/test_permission_controller.py tests/unit/agent/test_memory_id_normalize.py tests/unit/agent/test_multi_agent_crud.py tests/unit/agent/test_multi_user_crud.py tests/unit/agent/test_hybrid_cleanup.py

### DIFF
--- a/src/powermem/agent/components/permission_controller.py
+++ b/src/powermem/agent/components/permission_controller.py
@@ -10,11 +10,30 @@ import logging
 from datetime import datetime
 from typing import Any, Dict, List, Optional, Set
 
-from typing import Any, Dict
+from powermem.agent.utils.memory_id import memory_key_variants
 from powermem.agent.types import AccessPermission
 from powermem.agent.abstract.permission import AgentPermissionManagerBase
 
 logger = logging.getLogger(__name__)
+
+
+def _permission_value(permission: AccessPermission) -> str:
+    """Normalize permission to lowercase string value for role list comparison."""
+    if isinstance(permission, AccessPermission):
+        return permission.value
+    return str(permission).lower()
+
+
+def _memory_id_keys(memory_id: Any) -> List[Any]:
+    """Return memory id lookup keys while preserving non-numeric legacy ids."""
+    keys = [memory_id]
+    try:
+        for key in memory_key_variants(memory_id):
+            if key not in keys:
+                keys.append(key)
+    except ValueError:
+        pass
+    return keys
 
 
 class PermissionController(AgentPermissionManagerBase):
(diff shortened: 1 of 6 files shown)


wayyoungboy (approved):
Reviewed the current head. Permission checks now handle int/string memory IDs and enum/string permission values consistently, and the targeted permission + memory-id normalization tests pass. No blocker from my side.

wayyoungboy:
LGTM

PR #1153, opened 2026-07-10
test(storage): add OceanBase embedding-failure text fallback coverage…

### LINKED ISSUE #1094: Add OceanBase coverage for embedding-failure text fallback
Follow-up from review of PR #1052.

## Summary
Add OceanBase-focused coverage for search behavior when query embeddings cannot be generated and the search path must fall back to text retrieval.

## Suggested coverage
- retrieval_mode=auto with no query embedding and a text query.
- retrieval_mode=fts, retrieval_mode=vector, and retrieval_mode=hybrid behavior when vectors are unavailable.
- Threshold and filter behavior for the fallback path.

## Acceptance criteria
- The fallback behavior is covered by either a lightweight OceanBase-like unit test or a real backend integration test.
- Vector-only backends still return no false-positive results when vectors are unavailable.

kic635:
## Summary
Fix #1094.
Add required-path unit coverage for OceanBase search when query embeddings are unavailable and the path must fall back to text retrieval (follow-up from PR #1052).
No production code changes.
## Solution Description
Add `tests/unit/test_oceanbase_embedding_fallback.py` with two layers of coverage:
1. **StorageAdapter routing**
   - `auto` / `fts` / `hybrid` with `query_embedding=None` and a text query are forwarded to an OceanBase-like store
   - `vector` mode without embedding returns empty and does not call the store
   - vector-only OceanBase backends (`hybrid_search=False`) return no false-positive results
   - threshold / filter / sparse / fusion / `candidate_limit` controls are forwarded correctly
2. **OceanBaseVectorStore production paths**
   - real `_vector_search` safety belt: `vectors=None` / `[]` returns empty and does not call `ann_search`
   - FTS SQL generation and threshold behavior on the production FTS path
   - `auto` / `hybrid` without vectors still run FTS after the empty vector branch
   - weighted fusion with FTS-only results
   - remote parallel hybrid path
   - native hybrid falls back to application hybrid when vectors are missing
Verified locally with:
```bash
python -m pytest tests/unit/test_oceanbase_embedding_fallback.py -v

### DIFF
--- a/tests/unit/test_oceanbase_embedding_fallback.py
+++ b/tests/unit/test_oceanbase_embedding_fallback.py
@@ -1,0 +1,611 @@
+"""OceanBase-focused coverage for embedding-failure text fallback (Issue #1094).
+
+Follow-up from PR #1052: when query embeddings are unavailable, search must fall
+back to text retrieval on OceanBase-capable backends and must not invent hits on
+vector-only backends.
+
+Adapter tests use a call-recording spy only (no parallel search logic).
+Store tests construct OceanBaseVectorStore via __init__ with mocked client/table
+boundaries, then exercise production search / _vector_search / _fulltext_search.
+"""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+
+from powermem.storage.adapter import StorageAdapter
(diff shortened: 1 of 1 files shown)


wayyoungboy:
LGTM

Teingi:

PR #1164, opened 2026-07-13
Feat/langchain middleware

xiaotongsaying:

### DIFF
--- a/.github/workflows/vldb-comment.yml
+++ b/.github/workflows/vldb-comment.yml
@@ -1,0 +1,197 @@
+name: VLDB 2026 Contract Comment
+
+on:
+  workflow_run:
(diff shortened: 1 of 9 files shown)


wayyoungboy:
Thanks for the PR. I do not think we should merge this as-is.

Current blockers:

- The PR targets `main`, but the VLDB LangChain middleware work should target `vldb_2026`.
- The CLA check is still pending.
- This adds workflow/package skeleton files that overlap with the existing `vldb_2026` setup, so merging it to `main` would duplicate that work instead of landing only the middleware implementation.

Please retarget/rebase onto the intended branch, complete the CLA, and coordinate with the other LangChain middleware submissions before merge.

PR #1198, opened 2026-07-27
fix(docker): handle bind-mounted log directory permissions

cr7258:
## Problem

Starting `powermem-server` with the Compose bind mount `./logs:/app/logs` can leave the memory service unavailable:

```text
Failed to initialize service singletons: [Errno 13] Permission denied: '/app/logs/powermem.log'
PermissionError: [Errno 13] Permission denied: '/app/logs/powermem.log'
```

The health response then reports:

```json
{
  "status": "degraded",
  "memory_service_ready": false,
  "storage_type": null
}
```

## Reproduction

On a clean Linux checkout, follow the [recommended Docker Compose startup guide](https://github.com/oceanbase/powermem/blob/main/docker/README_CN.md#2-%E4%BD%BF%E7%94%A8-docker-compose-%E5%90%AF%E5%8A%A8-powermem-server-%E5%92%8C-seekdb%E6%8E%A8%E8%8D%90) with a root-owned host log directory:

```bash
git clone https://github.com/oceanbase/powermem.git
cd powermem

cp .env.example docker/.env
# Edit docker/.env and configure the LLM provider, model, API key, and base URL.

docker compose -f docker/docker-compose.yml up -d --build

docker compose -f docker/docker-compose.yml logs powermem-server
curl http://localhost:8848/api/v1/system/health
```

Before this fix, the server process runs as UID 1000 and cannot create `/app/logs/powermem.log` in the mounted directory, producing the error and degraded response above.

## Root Cause

The Compose service [bind-mounts `./logs` at `/app/logs`](https://github.com/oceanbase/powermem/blob/main/docker/docker-compose.yml#L42). The image switched to the non-root `powermem` user before the entrypoint ran. When the host log directory is created as `root`, the non-root process cannot repair its ownership before PowerMem opens `/app/logs/powermem.log`.

## Solution

- Install `gosu` in the runtime image.
- Create `/app/logs` with the expected ownership during the image build.
- Let the entrypoint prepare bind-mounted `/app/logs` before startup.
- Drop privileges with `gosu` before launching the requested command, so PowerMem still runs as the non-root `powermem` user.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -111,7 +111,7 @@ ENV PYTHONUNBUFFERED=1 \
 HEALTHCHECK --interval=30s --timeout=10s --start-period=40s --retries=3 \
     CMD curl -f http://localhost:8848/api/v1/system/health || exit 1
 
-# The entrypoint initializes bind-mounted directories, then drops to powermem.
+# Use entrypoint script
 ENTRYPOINT ["docker-entrypoint.sh"]
 
 # Default command: API server + PMem Dashboard on port 8848
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -78,10 +78,12 @@ RUN if [ -n "$DEBIAN_MIRROR" ]; then \
     apt-get install -y --no-install-recommends \
         postgresql-client \
         curl \
+        gosu \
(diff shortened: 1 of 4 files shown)


Teingi:

PR #1208, opened 2026-08-11
fix(e2e): redact replay evidence and clean up compose resources

thunguo:

Closes #.

# Rationale for this change

The Bub replay harness previously redacted credentials only at selected upstream call sites. Judge exceptions, span attributes, evaluation reasons, failures, and other values could bypass those scattered checks and persist provider credentials or authorization tokens in `replay.json`, `eval-report.json`, or `report.md`.

The Compose runner also used `set -e` while calling `docker compose down` only after the successful execution path. A failed image build, Server startup, or harness execution could therefore leave containers, networks, and database volumes behind. Cleanup failures could also obscure the original failure status.

This change enforces credential redaction at the evidence sink and Compose cleanup at the process lifecycle boundary, giving every execution path the same guarantees.

- Add a unified evidence redactor that:
  - collects sensitive environment values such as API keys, authorization values, tokens, passwords, and database URLs;
  - recursively processes scenarios, Memory snapshots, prepared context, spans, assertions, scores, labels, failures, and other JSON values;
  - recognizes bearer credentials, provider keys, authorization fields, URL userinfo, sensitive query parameters, and database URLs.
- Enforce redaction before writing all three evidence sinks:
  - `replay.json`
  - `eval-report.json`
  - `report.md`
- Preserve the JSON content and schema of normal live replays that contain no sensitive values.
- Register `EXIT`, `INT`, and `TERM` traps before the Compose build:
  - remove containers, networks, and volumes after both successful and failed runs;
  - preserve the original build, startup, or harness failure status;
  - prevent cleanup failures from masking an existing failure;
  - return the cleanup failure when the main execution succeeded.
- Add regression tests for evidence redaction and the Compose lifecycle.
- Run the Bub harness pytest suite as part of `make harness-check`.
- Document the automatic cleanup behavior and update the harness lock file.

NONE

# How was this change tested?

The added regression coverage verifies that:

- multiple sentinel credentials are injected across scenarios, Memory snapshots, prepared context, agent/model spans, assertions, scores, labels, failures, and the Markdown report;
- all three evidence artifacts are recursively scanned and contain zero sentinel credential matches;
- a normal live-mode replay remains exactly equal at the JSON level and still validates as the same Pydantic model;
- injected build, startup, and harness failures always invoke cleanup, leave container/network/volume state empty, and preserve exit codes `31`, `32`, and `33`;
- a cleanup failure does not mask an existing harness failure;
- the successful path uses the same cleanup behavior;
- a real Docker Compose run with a missing scenario preserves harness exit code `1`, while Compose output confirms that the container, volume, and network were removed;
- the real Docker validation used a fake sentinel key and did not call a cloud model.

Commands run:

```bash
UV_NO_SYNC=1 make harness-check
make harness-compose-check
/bin/sh -n e2e/bub/run.sh
/bin/dash -n e2e/bub/run.sh
```

Results:

```
8 passed
Ruff passed
format check passed
ty passed
Compose configuration check passed
CLI smoke test passed
```

# AI usage statement


PsiACE (changes requested):
I suggest simplifying this: redact known runtime secrets at the evidence sink, then use TruffleHog before publication. If scanning fails, skip summary and artifact upload without failing CI. Keep tests on observable behavior, not regex coverage, field locations, or Docker command order.

thunguo:
> I suggest simplifying this: redact known runtime secrets at the evidence sink, then use TruffleHog before publication. If scanning fails, skip summary and artifact upload without failing CI. Keep tests on observable behavior, not regex coverage, field locations, or Docker command order.

Got it, I'll adjust things in that direction.

PsiACE:
Thanks for catching this. After looking more closely at the CI path, I think live replay should stay local for now because uploaded evidence may leak credentials. Contributors should still run it when changing this flow. This does not block the fixes in this PR. What do you think?

thunguo:
> Thanks for catching this. After looking more closely at the CI path, I think live replay should stay local for now because uploaded evidence may leak credentials. Contributors should still run it when changing this flow. This does not block the fixes in this PR. What do you think?

👌，I fully agree with this approach.

PsiACE (approved):
thanks, totally lgtm

PR #1211, opened 2026-08-11
fix(powercontext): use --frozen for Codex plugin uv commands

### LINKED ISSUE #1210: bug: Codex plugin commands fail with --locked when uv uses a custom index
### Describe the bug

The PowerContext Codex plugin uses `uv run --locked` in the following files:

- `integrations/codex/plugins/powercontext/hooks/hooks.json`
- `integrations/codex/plugins/powercontext/skills/project-context/SKILL.md`

When the user's global uv configuration uses a custom package index, such as the Tsinghua PyPI mirror, while the bundled `uv.lock` records `https://pypi.org/simple`, uv considers the lockfile outdated.

This prevents the `UserPromptSubmit` hook from running when a prompt is submitted and prevents the `project-context` skill from resolving the project scope.

Possible solutions include:

- Replace `--locked` with `--frozen`, so the plugin uses the bundled lockfile without checking or updating it.
- Remove `--locked`, allowing uv to update the lockfile using the user's configured index.

Using `--frozen` may be preferable for a packaged plugin because it preserves the bundled dependency versions and does not modify the installed plugin's lockfile.

I am willing to submit a PR to fix this issue.

### Steps to reproduce

1. Configure a custom package index in `~/.config/uv/uv.toml`:

```toml
index-url = "https://pypi.tuna.tsinghua.edu.cn/simple"
```

2. Install and enable the PowerContext Codex plugin.

3. Submit a prompt in Codex, or invoke the `project-context` skill.

4. The plugin command fails with a lockfile error.

The problem can also be reproduced directly from the repository root:

```bash
uv run --locked --quiet \
  --project integrations/codex/plugins/powercontext \
  python integrations/codex/plugins/powercontext/scripts/project_scope.py \
  --cwd "$PWD"
```

### Expected behavior

The PowerContext Codex plugin should work when the user has configured a custom uv package index.

The plugin should be able to use its bundled lockfile without requiring the user to run `uv lock` or manually modify the installed plugin files.

### Actual behavior

The Hook and Skill commands fail with the following error:

```text
error: The lockfile at `uv.lock` needs to be updated, but `--locked` was provided.
To update the lockfile, run `uv lock`.
```

Temporarily overriding the package index back to `https://pypi.org/simple` allows the same `--locked` command to succeed.

Replacing `--locked` with `--frozen` also allows the command to run successfully with the custom uv index configuration.

### Environment

_No response_

Evenss:

Closes #1210.

# Rationale for this change

The PowerContext Codex plugin runs project-local helper scripts with `uv run --locked`.

When a user's global uv configuration uses a custom package index while the bundled `uv.lock` records PyPI, uv considers the lockfile outdated and exits. This breaks both the `UserPromptSubmit` hook and the `project-context` skill.

Using `--frozen` allows the plugin to trust its bundled lockfile without checking or modifying it.

- Replace `--locked` with `--frozen` in the `UserPromptSubmit` hook command.
- Replace `--locked` with `--frozen` in the `project-context` skill command.
- No dependency versions, public APIs, or persisted formats are changed.

Yes. The Codex plugin can now run when users configure a custom uv package index without requiring them to regenerate the plugin's lockfile.

There are no breaking changes.

# How was this change tested?

- Reproduced the lockfile error with `--locked` while using the Tsinghua PyPI mirror.
- Verified that the same plugin command succeeds with `--frozen`.
- Validated `hooks/hooks.json` with `python3 -m json.tool`.
- Ran `git diff --check upstream/master...HEAD`.

# AI usage statement

OpenAI Codex (GPT-5.6) was used to inspect the affected plugin paths, reproduce the uv index and lockfile behavior, and draft the change and PR description. The final diff and validation results were reviewed by the contributor.

### DIFF
--- a/integrations/codex/plugins/powercontext/hooks/hooks.json
+++ b/integrations/codex/plugins/powercontext/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "uv run --locked --quiet --project \"${PLUGIN_ROOT}\" python \"${PLUGIN_ROOT}/hooks/recall.py\"",
+            "command": "uv run --frozen --quiet --project \"${PLUGIN_ROOT}\" python \"${PLUGIN_ROOT}/hooks/recall.py\"",
             "timeout": 10,
             "statusMessage": "Syncing PowerContext"
           }
--- a/integrations/codex/plugins/powercontext/skills/project-context/SKILL.md
+++ b/integrations/codex/plugins/powercontext/skills/project-context/SKILL.md
@@ -18,7 +18,7 @@ to duplicate the current prompt.
 Before the first memory tool call, run:
 
 ```bash
-uv run --locked --quiet --project "$PLUGIN_ROOT" python "$PLUGIN_ROOT/scripts/project_scope.py" --cwd "$PWD"
+uv run --frozen --quiet --project "$PLUGIN_ROOT" python "$PLUGIN_ROOT/scripts/project_scope.py" --cwd "$PWD"
 ```
 
 Reuse that exact `scope_id` for the task.


Teingi (approved):
LGTM

Issue #1189, opened 2026-07-20
[Enhancement]: 增加配置校验 || [Enhancement]: Add configuration verification


ob-lzq:
### Description

目前手动调用用户画像时报错，发现配置 Embedding 的  base_url  格式写错时（缺少v1），部署时和使用其他服务过程也未有异常，用户无法感知，可能造成一些功能的缺失。

需要增加部署时和使用期间配置异常反馈，比如dashboard 异常提示等。

---
### Description

Currently, an error occurs when manually calling user portraits. When it is found that the base_url format of the Embedding configuration is written incorrectly (v1 is missing), there are no abnormalities during deployment and use of other services. Users cannot perceive it, which may cause the loss of some functions.

It is necessary to add configuration exception feedback during deployment and use, such as dashboard exception prompts, etc.

Keesan12:
I would split validation into two stages: static config checks before startup, then live capability checks after dependencies connect. Each feature should report working, disabled by policy, or degraded with a specific reason. “Process is up” is not enough when memory writes are quietly failing.

We are building MartinLoop around truthful runtime state because agents need to know when their context layer cannot be trusted. I can test the proposed checks against missing credentials, unreachable stores, and partial permissions and share the resulting receipts. https://github.com/Keesan12/martin-loop

PsiACE:
Superseded by #1215, which covers the broader `powercontext doctor` and Server readiness design.

PR #1218, opened 2026-08-12
improve(doctor): split integration checks and improve runtime readiness

### LINKED ISSUE #1215: feat: separate core and integration checks in powercontext doctor
## Feature description

`powercontext doctor` currently treats the Codex CLI and plugin as required checks, although PowerContext can run without the Codex integration.

Refine diagnostics into two groups:

- Core checks cover the installed package, Server liveness, and Server readiness.
- Server readiness owns configured runtime dependencies such as the database and inference providers.
- Integration checks, starting with Codex, run only when explicitly requested or during the related setup command.

The default command should not fail when Codex is not installed. It should distinguish an unreachable Server from a live Server that is not ready and preserve the detailed readiness checks returned by the Server.

A configured inference provider is one example of a Server-side readiness check. `doctor` should report that result without accessing provider credentials directly.

## Acceptance criteria

- A healthy PowerContext Server can pass the default diagnostics without Codex installed.
- An explicitly requested Codex check fails when the CLI or plugin is unavailable.
- The output distinguishes liveness failures from readiness failures.
- Readiness details remain available in human-readable and JSON output.
- `doctor` does not access database or provider credentials directly.

Related to #1189.

thunguo:

Closes #1215 .

# Rationale for this change

`powercontext doctor` currently mixes core PowerContext checks with optional Codex integration checks. This causes the default doctor command to fail when Codex is not installed, even when PowerContext itself is healthy.

Server readiness also needs to reflect the availability of configured runtime dependencies, while keeping database and inference credentials within the Server process.

- Split the doctor commands by responsibility:
  - `powercontext doctor` checks the installed package, Server liveness, and Server readiness.
  - `powercontext doctor codex` explicitly checks the Codex CLI and PowerContext plugin.
  - `powercontext setup codex` reuses the same Codex diagnostics after installation.
- Added explicit `ok`, `degraded`, `failed`, and `skipped` diagnostic states. JSON output includes `status` at the top level and for each check while preserving the existing `ok` field.
- Distinguished an unreachable Server from a live-but-not-ready Server.
- Preserved nested readiness details from expected HTTP 503 responses in both human-readable and JSON output.
- Added transport-independent Runtime readiness aggregation for:
  - Runtime availability.
  - Database connectivity using `SELECT 1`.
  - Configured generation and embedding providers.
- Added real inference canaries with:
  - A two-second timeout per dependency.
  - A 300-second cache for `ready` and `misconfigured` results, and a 30-second cache for `timeout` and `unavailable` results.
  - Single-flight protection for concurrent checks.
  - Stable and sanitized dependency statuses.
- Integrated Runtime readiness with the production Server factory, lifecycle events, and readiness metrics.
- Added startup readiness warm-up without making dependency failures terminate the Server.
- Updated English and Chinese documentation for the new doctor commands, health semantics, readiness statuses, and provider canary behavior.
- Added CLI, Runtime, Server, caching, security, and regression tests.

Yes.

- The default `powercontext doctor` command no longer requires Codex or the PowerContext Codex plugin.
- Codex integration checks are now run explicitly with `powercontext doctor codex`.
- `powercontext doctor --json` now reports:
  - `package`
  - `server_liveness`
  - `server_readiness`
- The previous `server`, `codex`, and `plugin` keys are no longer emitted by the default doctor command. Scripts consuming the old JSON keys will need to be updated.
- Readiness failures now include safe, nested dependency details.
- `/health/ready` returns HTTP 503 with `not_ready` when the Runtime or database is unavailable. Configured inference failures return HTTP 200 with `degraded`.
- Inference readiness canaries run during startup. `ready` and `misconfigured` results are cached for 300 seconds; `timeout` and `unavailable` results are retried after 30 seconds. Concurrent checks share one refresh. Generation probes request at most one output token.

There are no persisted-format changes. The HTTP/OpenAPI contract adds the `degraded` readiness status. The `doctor` JSON key changes are described above and require consumers of the previous keys to update.

# How was this change tested?

Automated validation:

- `make test`: 452 passed, 7 skipped.
- `make contract-test`: 24 passed.
- `make check`: passed.
- `make docs-test`: passed.
- `git diff --check`: passed.

Manual validation with real Server processes included:

- Verified that the default doctor succeeds without Codex when the Server is healthy.
- Verified that `doctor codex` fails correctly when the Codex CLI or plugin is missing.
- Verified that an unreachable Server skips readiness after liveness fails.
- Verified HTTP 200 for a healthy readiness response.
- Verified HTTP 503 with nested details for a live-but-not-ready Server.
- Verified generation readiness using an actual configured model adapter.
- Verified that an OpenAI-compatible provider returning HTTP 404 is reported as `misconfigured`, covering the #1189 regression.
- Verified that inference failures are cached and concurrent checks do not repeat provider canaries.
- Verified that API keys, provider URLs, database URLs, exception messages, and response bodies are not exposed in CLI output, health responses, or logs.
- Verified that the CLI does not load or use Server database and inference credentials.

# AI usage statement

### DIFF
--- a/src/powercontext/builtin/runtime/readiness.py
+++ b/src/powercontext/builtin/runtime/readiness.py
@@ -1,0 +1,136 @@
+"""Runtime-owned dependency readiness checks."""
+
+from __future__ import annotations
+
+import asyncio
+from collections.abc import Awaitable, Callable, Mapping
+from dataclasses import dataclass
+from enum import StrEnum
+from time import monotonic
+
+from powercontext.builtin.inference import InferenceConfigurationError
+
+READINESS_PROBE_TIMEOUT_SECONDS = 2.0
+READINESS_PROBE_CACHE_SECONDS = 300.0
+
+DependencyOperation = Callable[[], Awaitable[object]]
+Clock = Callable[[], float]
+
+
+class ReadinessCheckStatus(StrEnum):
+    """Stable outcomes exposed for one Runtime dependency."""
+
+    READY = "ready"
+    UNAVAILABLE = "unavailable"
+    TIMEOUT = "timeout"
+    MISCONFIGURED = "misconfigured"
+
+
+ReadinessProbe = Callable[[], Awaitable[ReadinessCheckStatus]]
+
+
+@dataclass(frozen=True, slots=True)
+class RuntimeReadiness:
+    """Aggregate the safe readiness outcomes for one Runtime."""
+
+    checks: Mapping[str, ReadinessCheckStatus]
+
+    @property
+    def ready(self) -> bool:
+        """Return whether every registered dependency is ready."""
+
+        return all(status is ReadinessCheckStatus.READY for status in self.checks.values())
+
+
+class RuntimeReadinessChecks:
+    """Run independent dependency probes concurrently."""
+
+    def __init__(self, probes: Mapping[str, ReadinessProbe] | None = None) -> None:
+        self._probes = {} if probes is None else dict(probes)
+
+    async def run(self) -> RuntimeReadiness:
+        """Return safe results in registration order without exposing failures."""
+
+        names = tuple(self._probes)
+        results = await asyncio.gather(*(self._run(self._probes[name]) for name in names))
+        return RuntimeReadiness(checks=dict(zip(names, results, strict=True)))
+
+    @staticmethod
+    async def _run(probe: ReadinessProbe) -> ReadinessCheckStatus:
+        try:
+            return await probe()
+        except asyncio.CancelledError:
+            raise
+        except Exception:
+            return ReadinessCheckStatus.UNAVAILABLE
+
+
+class CachedReadinessProbe:
+    """Cache one dependency result and collapse concurrent refreshes."""
+
+    def __init__(
+        self,
+        probe: ReadinessProbe,
+        *,
+        ttl_seconds: float = READINESS_PROBE_CACHE_SECONDS,
+        clock: Clock = monotonic,
+    ) -> None:
+        self._probe = probe
+        self._ttl_seconds = ttl_seconds
+        self._clock = clock
+        self._lock = asyncio.Lock()
+        self._result: ReadinessCheckStatus | None = None
+        self._expires_at = 0.0
+
+    async def __call__(self) -> ReadinessCheckStatus:
+        """Return a fresh cached result, refreshing it at most once."""
+
+        result = self._fresh_result()
+        if result is not None:
+            return result
+        async with self._lock:
+            result = self._fresh_result()
+            if result is not None:
+                return result
+            result = await self._probe()
+            self._result = result
+            self._expires_at = self._clock() + self._ttl_seconds
+            return result
+
+    def _fresh_result(self) -> ReadinessCheckStatus | None:
+        return self._result if self._result is not None and self._clock() < self._expires_at else None
+
+
+def dependency_readiness_probe(
+    operation: DependencyOperation,
+    *,
+    timeout_seconds: float = READINESS_PROBE_TIMEOUT_SECONDS,
+) -> ReadinessProbe:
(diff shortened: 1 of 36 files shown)


src/powercontext/builtin/runtime/readiness.py
```diff
@@ -0,0 +1,136 @@
...
+class RuntimeReadiness:
+    """Aggregate the safe readiness outcomes for one Runtime."""
+
+    checks: Mapping[str, ReadinessCheckStatus]
+
+    @property
+    def ready(self) -> bool:
+        """Return whether every registered dependency is ready."""
+
+        return all(status is ReadinessCheckStatus.READY for status in self.checks.values())
```
PsiACE: Can we avoid making the whole Server not ready when an inference provider fails? Database operations can still work, but this currently causes `/health/ready` to return 503.

Please use `not_ready` with HTTP 503 for Runtime or database failures, and `degraded` with HTTP 200 for inference failures. This should be part of the probe definition rather than based on check names.

Since this changes the public response, please update `openapi/powercontext.yaml`, run `make api-generate`, and then run `make contract-test`.

src/powercontext/builtin/runtime/readiness.py
```diff
@@ -0,0 +1,136 @@
...
+        result = self._fresh_result()
+        if result is not None:
+            return result
+        async with self._lock:
+            result = self._fresh_result()
+            if result is not None:
+                return result
+            result = await self._probe()
+            self._result = result
+            self._expires_at = self._clock() + self._ttl_seconds
```
PsiACE: Can we retry temporary failures sooner? A single timeout, rate limit, or network error is currently cached for five minutes.

Please keep concurrent checks sharing the same request, but use a shorter cache time for `timeout` and `unavailable`, such as 30 seconds. The 300-second cache can remain for `ready` and `misconfigured`.

src/powercontext/cli/system.py
```diff
@@ -72,10 +75,33 @@ class CodexSetupResult:
...
+    @property
+    def ok(self) -> bool:
+        """Return whether this check passed."""
+
+        return self.status is DiagnosticStatus.OK
+
+    def as_json(self) -> dict[str, object]:
+        """Return the stable external diagnostic representation."""
+
+        result: dict[str, object] = {"ok": self.ok, "detail": self.detail}
```
PsiACE: Please include `status` in the JSON output. `failed` and `skipped` currently both appear as `ok: false`, so callers have to parse `detail`. `degraded` would have the same problem.

Please keep `ok` for compatibility and add `status` to the top-level result and each check.

PsiACE (changes requested):
The command split and the Server-side credential handling look good.

The main issue is readiness behavior. Runtime and database failures should make the Server not ready. Inference failures should be reported as degraded without removing the Server from traffic.

Tests should focus on HTTP and CLI behavior, end-to-end acceptance, the #1189 regression, and provider request limits. They should avoid depending on internal helper calls or exact log and call counts.

thunguo:
> The command split and the Server-side credential handling look good.
> 
> The main issue is readiness behavior. Runtime and database failures should make the Server not ready. Inference failures should be reported as degraded without removing the Server from traffic.
> 
> Tests should focus on HTTP and CLI behavior, end-to-end acceptance, the #1189 regression, and provider request limits. They should avoid depending on internal helper calls or exact log and call counts.

Alright, I'll handle these issues.

thunguo:
> The command split and the Server-side credential handling look good.
> 
> The main issue is readiness behavior. Runtime and database failures should make the Server not ready. Inference failures should be reported as degraded without removing the Server from traffic.
> 
> Tests should focus on HTTP and CLI behavior, end-to-end acceptance, the #1189 regression, and provider request limits. They should avoid depending on internal helper calls or exact log and call counts.

Done.

PsiACE (approved):
lgtm, thanks

PR #1222, opened 2026-08-12
fix(scheduler): fail to create the parent directory for its SQLite job store

### LINKED ISSUE #1221: bug: scheduler fails to create the parent directory for its SQLite job store
### Describe the bug

When scheduled processing is enabled and `scheduler_path` points to a file whose parent directory does not exist, PowerContext does not create that directory before initializing APScheduler's `SQLAlchemyJobStore`.

As a result, scheduler startup fails with:

`sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) unable to open database file`

This is exposed when running the Server with OceanBase and a fresh `POWERCONTEXT_HOME`. The default scheduler database is stored at `<POWERCONTEXT_HOME>/scheduler.db`, but the OceanBase persistence backend does not create this local directory.

The default SQLite backend masks the issue because its main database profile creates the same parent directory as a side effect.

The scheduler owns its SQLite sidecar database and should initialize its parent directory independently of the main persistence backend.

### Steps to reproduce

1. Create a temporary project and install PowerContext:

```bash
uv init powercontext-repro
cd powercontext-repro
uv add "powercontext[builtin] @ git+https://github.com/oceanbase/powercontext.git@7aeeffa170f2e6be2a0fe7a581aaf6905b7a5fe6"
```

2. Save the following as repro.py:

```python
import asyncio
from pathlib import Path

from powercontext.builtin.runtime.scheduler import create_scheduler


async def main():
    with TemporaryDirectory() as tmp:
        database = Path(tmp) / "missing" / "scheduler.db"
        assert not database.parent.exists()

        scheduler = create_scheduler(database)
        scheduler.start()

        try:
            assert database.is_file()
        finally:
            scheduler.shutdown()


asyncio.run(main())
```

3. Run the reproduction:
```bash
uv run python repro.py
```

### Expected behavior

PowerContext should create the missing parent directories and initialize `scheduler.db`.

The scheduler should start successfully regardless of whether the main persistence backend is SQLite or OceanBase.

### Actual behavior

The scheduler fails during startup because the parent directory does not exist:

```text
sqlalchemy.exc.OperationalError:
(sqlite3.OperationalError) unable to open database file
```

The process exits with status 1, and the scheduler database parent directory remains absent.

A full OceanBase Server reproduction produced:

```
exception_type=OperationalError
unable_to_open=True
home_created=False
sqlite_control_scheduler_exists=True
```

This confirms that the SQLite control case succeeds only because the SQLite main database profile creates the directory first.

### Environment

- PowerContext version: 0.0.1
- Reproduction revision: `7aeeffa170f2e6be2a0fe7a581aaf6905b7a5fe6`
- Also reproduced on revision: `a5e9340d7334f8e442a0873d3bd8012e4a50554b`
- Python version: 3.12.13; also reproduced on 3.14.3
- OS: macOS 26.2, arm64
- SQLAlchemy: 2.0.51
- APScheduler: 3.11.3
- OceanBase integration environment: OceanBase CE 4.4.2.1
- OceanBase is not required for the minimal reproduction

thunguo:

Closes #1221 .

# Rationale for this change

When the scheduler database is located under a parent directory that does not yet exist, PowerContext passes the path directly to APScheduler's SQLite job store. Scheduler startup then fails with `sqlite3.OperationalError: unable to open database file`.

This is especially visible when using OceanBase with a fresh `POWERCONTEXT_HOME`, because the OceanBase backend does not create the local directory. The default SQLite backend previously masked the issue by creating the directory while initializing the main database.

The scheduler owns its SQLite sidecar database and should create the required parent directory independently of the main persistence backend.

- Create the scheduler database parent directory before initializing the SQLite job store.
- Support missing nested parent directories using `mkdir(parents=True, exist_ok=True)`.
- Add a regression test covering scheduler startup with a previously nonexistent parent directory.
- Preserve the existing validation for unsupported paths such as `:memory:`.

NONE

# How was this change tested?

Added `test_scheduler_creates_missing_database_parent_directory`, which verifies that `BuiltinRuntime.start_scheduler()` creates nested parent directories and initializes the scheduler database.

Validation performed:

- `uv run pytest tests/builtin/runtime/test_scheduler.py -q` — 8 passed
- `make check` — passed
- `make test` — 434 passed, 7 skipped
- Manually started the Server against OceanBase CE 4.4.2.1 with scheduled processing enabled and a fresh `POWERCONTEXT_HOME`; the Server started successfully and created `scheduler.db`.

The regression test was also confirmed to fail before the fix with `sqlite3.OperationalError: unable to open database file`.

# AI usage statement

### DIFF
--- a/src/powercontext/builtin/runtime/scheduler.py
+++ b/src/powercontext/builtin/runtime/scheduler.py
@@ -103,7 +103,9 @@ def _processor(runtime_key: str, name: str) -> Processor:
 def create_scheduler(scheduler_path: str | Path) -> AsyncIOScheduler:
     """Create an APScheduler instance with an isolated SQLite job-store."""
 
-    url = URL.create("sqlite+pysqlite", database=scheduler_database_path(scheduler_path))
+    database = Path(scheduler_database_path(scheduler_path))
+    database.parent.mkdir(parents=True, exist_ok=True)
+    url = URL.create("sqlite+pysqlite", database=str(database))
     return AsyncIOScheduler(
         jobstores={
             "default": SQLAlchemyJobStore(
--- a/tests/builtin/runtime/test_scheduler.py
+++ b/tests/builtin/runtime/test_scheduler.py
@@ -109,6 +109,21 @@ def test_scheduler_persists_one_stable_source_window_job(tmp_path) -> None:
     asyncio.run(scenario())
 
 
+def test_scheduler_creates_missing_database_parent_directory(tmp_path) -> None:
+    async def scenario() -> None:
+        database = tmp_path / "missing" / "nested" / "scheduler.db"
+        assert not database.parent.exists()
+
+        runtime = _runtime(_ScheduledTriggers())
+        try:
+            runtime.start_scheduler(database, 3_600)
+            assert database.is_file()
+        finally:
+            await runtime.close()
+
+    asyncio.run(scenario())
+
+
 def test_scheduler_interval_activates_the_source_window_policy(tmp_path) -> None:
     async def scenario() -> None:
         triggers = _ScheduledTriggers()


PsiACE (approved):
nice catch, lgtm, thanks

Issue #1214, opened 2026-08-11
Tracking Issue: Fine-Grained Tracing for PowerContext


PsiACE:
## Goal

Track the remaining tracing work on top of PowerContext's OpenTelemetry support.

This issue replaces #159 and the tracing requirements in #1163.

## Current support

- [x] W3C Trace Context propagation
- [x] Optional OTLP export
- [x] HTTP and MCP ingress spans
- [x] Application operation spans
- [x] Outbound PowerContext Client spans
- [x] Request, trace, and span correlation in logs

## Remaining work

- [x] Document OpenTelemetry Collector and trace backend setup
- [ ] Trace Memory search and retrieval stages
- [ ] Trace reranking and context preparation
- [ ] Trace Memory capture, extraction, indexing, and persistence
- [x] Trace generation, embedding, and other remote calls
- [ ] Trace scheduled and background processing
- [ ] Test span hierarchy, outcomes, and context propagation

## Constraints

- Do not record queries, prompts, responses, Memory content, or vectors.
- Keep span attributes bounded.
- Tracing must remain optional and must not affect Runtime behavior.
- Do not store traces or build a trace viewer in PowerContext.
- Keep the existing operation names and W3C propagation contract.

## References

- `docs/en/rfcs/0046_observability_foundations.md`
- `src/powercontext/server/tracing.py`
- `src/powercontext/client/tracing.py`

Ethan-Xingyue:
I'm very interested in this issue and would love to learn more about it. May I ask how difficult this issue is? I noticed that there seem to be quite a few tasks involved. Is there a specific deadline or expected timeline for completing it?

PsiACE:
Hi @Ethan-Xingyue . Thanks for your interest! #1214 is a tracking issue, so there is no expectation to complete everything and no fixed deadline.

I can split out a smaller task: integrate Pydantic AI's existing OpenTelemetry support, then document a minimal tracing setup. If you're interested, I can open a separate issue and assign it to you.

Ethan-Xingyue:
> Hi [@Ethan-Xingyue](https://github.com/Ethan-Xingyue) . Thanks for your interest! [#1214](https://github.com/oceanbase/powercontext/issues/1214) is a tracking issue, so there is no expectation to complete everything and no fixed deadline.
> 
> I can split out a smaller task: integrate Pydantic AI's existing OpenTelemetry support, then document a minimal tracing setup. If you're interested, I can open a separate issue and assign it to you.

[...]

PR #1227, opened 2026-08-13
feat(tracing): trace Pydantic AI generation and embedding calls

### LINKED ISSUE #1217: feat: integrate Pydantic AI tracing with PowerContext OpenTelemetry
## Goal

When PowerContext tracing is enabled, enable Pydantic AI's native OpenTelemetry instrumentation for generation and embedding calls. Export these spans through the Server's existing tracer provider so they can be viewed with the rest of the PowerContext trace.

## Requirements

- Keep Pydantic AI instrumentation disabled when tracing is disabled.
- Reuse the Server-owned tracer provider and OTLP exporter. A Logfire dependency is not needed.
- Preserve the current trace context so inference spans appear under the relevant PowerContext operation span when one exists.
- Embedded Runtime usage must not configure global tracing.
- Tracing failures must not affect Runtime behavior.
- This task does not need to define or expose content capture settings. Use the minimal instrumentation needed to verify generation and embedding traces.

## Documentation

Add an English and Chinese how-to using a local Phoenix container. Cover starting Phoenix, configuring OTLP export, triggering an inference request, and finding the resulting trace.

## Acceptance criteria

- Generation and embedding spans are exported through PowerContext's tracing pipeline only when tracing is enabled.
- Tests verify that instrumentation follows the tracing setting and preserves the expected span hierarchy.
- The documented setup shows transport, application, and inference spans in one trace.
- `make test` and `make docs-test` pass.

Part of #1214.

Ethan-Xingyue:

Closes #1217. Part of #1214.

# Rationale for this change

Transport and application spans already export, but the model calls underneath them were invisible. Pydantic AI ships
native OpenTelemetry instrumentation, so this is wiring rather than new telemetry code.

- `ServerTracing` builds an `InstrumentationSettings` bound to its own tracer provider, only when tracing is enabled.
  The global provider is never touched, and construction failure degrades to `None`.
- Content capture is hardcoded off, with no configuration surface. `include_content=False` alone is insufficient — the
  full prompt otherwise leaks via `model_request_parameters`.
- `open_builtin_runtime` takes one optional `instrumentation` argument, wrapping the generation model in
  `InstrumentedModel` and passing `instrument=` to `Embedder`.
- Readiness probing uses the unwrapped model, so health checks emit no parentless traces.
- The six generators pass stable `name=` values, so spans read `invoke_agent memory_extraction` rather than
  `invoke_agent agent`.
- New `Trace with Phoenix` how-to in both locales.

Span nesting needed no new code, and no dependencies changed.

All additive; tracing stays off by default. Known boundaries: injected pipelines and embedding models are not
instrumented, scheduler-triggered inference becomes a trace root, span names track the Pydantic AI conventions, and
`include_content=False` does not cover schema-validation retries (see comment below).

# How was this change tested?

```bash
make test
make docs-test
make check
```

Added tests for the tracing switch, the content switches, the `HTTP` → `powercontext` → `invoke_agent` → `chat` span
chain with a negative case, and embedding span nesting. Manually validated against `arizephoenix/phoenix:20.1.0`.


Ethan-Xingyue:
`include_content=False` does not cover the schema-validation retry path. Concretely:

A model returns `{"candidates":[{"text":"Ethan prefers aisle seats"}]}`, missing the required `intent` field. Pydantic
AI retries with validation feedback that quotes the model's own output back to it:

```json
{"type":"missing","loc":["candidates",0,"intent"],
 "input":{"text":"Ethan prefers aisle seats"}}
```

That feedback is recorded in full as a message in `gen_ai.input.messages` and `pydantic_ai.all_messages`, so the
proposed Memory content ends up in the trace.

The cause is that `RetryPromptPart.otel_message_parts` writes content unconditionally when `tool_name is None`; only
the `else` branch checks `include_content`. We use `PromptedOutput`, so output-validation failures take the first
branch. With the default `generation_max_requests=2` (`retries=1`) this path is live; reproduced locally.

I treated it as an upstream bug and only documented the exception in the new how-to. Worth confirming whether that is
enough, or whether we should strip those attributes before export so #1214 holds unconditionally.

PsiACE (approved):
totally lgtm

PsiACE:
> does not cover the schema-validation retry path. 

Upstream should fix it; we can track it in a new issue.

Ethan-Xingyue:
> > does not cover the schema-validation retry path.
> 
> Upstream should fix it; we can track it in a new issue.



> > does not cover the schema-validation retry path.
> 
> Upstream should fix it; we can track it in a new issue.

Get，if there are more #1214 subtasks available, I'd be glad to pick some up～～

PR #1235, opened 2026-08-13
fix(deps): require pydantic-ai 2.27.1 so tracing excludes retry content

### LINKED ISSUE #1228: fix: `include_content=False` does not cover the schema-validation retry path. Concretely:
`include_content=False` does not cover the schema-validation retry path. Concretely:

A model returns `{"candidates":[{"text":"Ethan prefers aisle seats"}]}`, missing the required `intent` field. Pydantic
AI retries with validation feedback that quotes the model's own output back to it:

```json
{"type":"missing","loc":["candidates",0,"intent"],
 "input":{"text":"Ethan prefers aisle seats"}}
```

That feedback is recorded in full as a message in `gen_ai.input.messages` and `pydantic_ai.all_messages`, so the
proposed Memory content ends up in the trace.

The cause is that `RetryPromptPart.otel_message_parts` writes content unconditionally when `tool_name is None`; only
the `else` branch checks `include_content`. We use `PromptedOutput`, so output-validation failures take the first
branch. With the default `generation_max_requests=2` (`retries=1`) this path is live; reproduced locally.

I treated it as an upstream bug and only documented the exception in the new how-to. Worth confirming whether that is
enough, or whether we should strip those attributes before export so #1214 holds unconditionally.

_Originally posted by @Ethan-Xingyue in https://github.com/oceanbase/powercontext/issues/1227#issuecomment-5277486467_

Ethan-Xingyue:

Closes #1228.

# Rationale for this change

`include_content=False` did not cover the schema-validation retry path: the retry feedback quoted the
model's raw output into `gen_ai.input.messages` and `pydantic_ai.all_messages`, and for Memory extraction
that output is the proposed Memory content. Upstream fixed it in 2.27.1, but our floor was `>=2.14.1`, so
a fresh install could still resolve a leaking version.

- Raise the `pydantic-ai-slim` floor to `>=2.27.1` and relock (resolves to 2.29.0).
- Drop the now-obsolete retry-exception paragraph from the Phoenix how-to, `docs/en` and `docs/zh`.

The minimum `pydantic-ai-slim` is now 2.27.1, and the tracing guarantee from #1227 holds without
exception. No API or persisted-format changes.

# How was this change tested?

`make test`, `make docs-test`, and `make check` pass. Leak checked with an in-memory span exporter
through the real extraction path: 2.14.1 leaks those two attributes, 2.29.0 leaks nothing, both with the
retry confirmed triggered.


Ethan-Xingyue:
Added a regression test so this can't come back: the model first returns output missing a nested required field, and we assert that text never lands in a span. The output type has to be nested — a flat one gives a `loc` of length 1, which upstream already excluded. Passes on 2.29.0, fails on 2.14.1.

PsiACE (approved):
LGTM, thanks

Issue #1219, opened 2026-08-12
Tracking Issue: Development Context and Scope Model


PsiACE:
## Goal

Track the design and follow-up work needed to define how Project, Workstream, Session, and scope relate across PowerContext.

PowerContext originally used a project-derived `scope_id` to share context across Sessions. Handoff later used the same scope as the identity of one linear Workstream. This assumption does not cover projects with parallel work or development spanning multiple projects.

## RFC required

Implementation requires a new RFC that defines:

- Project, Workstream, Session, and scope semantics;
- their relationships and identity boundaries;
- scope granularity for Memory, Handoff, Statistics, and other Runtime capabilities;
- whether `project_id` remains a separate identity;
- compatibility expectations for the API, CLI, integrations, and existing data.

The RFC should decide the model and invariants without fixing detailed user-flow orchestration prematurely.

## Tracking

- [ ] Draft and review the RFC.
- [ ] Record the accepted identity and scope model.
- [ ] Reconcile affected existing RFCs, including RFC 0019, RFC 0048, and RFC 0082.
- [ ] Create follow-up issues for API, CLI, Codex integration, Dashboard, and Handoff Report work.
- [ ] Implement changes only after the RFC is accepted.

## Constraints

- Preserve shared project context across Sessions.
- Keep independent Workstream histories isolated.
- Do not equate Session with Workstream.
- Do not assume the current `project_id` model is final.
- Treat migration and detailed orchestration as follow-up work.

larry-zy:
Hi! I'm really interested in this issue. If no one's on it yet, I'd like to take it and author the RFC (Project / Workstream / Session / scope_id model, reconciled with RFC 0019/0048/0082). I'll align with you on scope first, then open a draft PR. Could you assign it to me? Thanks!

PR #1238, opened 2026-08-14
RFC: scope model — Project sharing + Workstream isolation

larry-zy:
## Summary

PowerContext derives a `scope_id` from a project and uses it to share durable context across Sessions; Handoff later reused that same scope as the identity of one linear Workstream. As a result `scope_id` is overloaded with three conflated meanings at once — the **project** (sharing unit), the **memory-isolation / Runtime partition key**, and the **Workstream** (isolation unit) — where sharing wants scopes to be the *same* and isolation wants them *different*. One key cannot be both, so the current model cannot express parallel work inside one project or development spanning several projects.

This RFC fixes the identity boundaries relating **Project**, **Workstream**, **Session**, and **scope**, and records the accepted model and invariants (I1–I8) without prematurely fixing user-flow orchestration, migration tooling, or concrete API/CLI/Dashboard shapes.

## Decision — Alternative B

Three approaches were weighed (full comparison in *Rationale and alternatives*):

- **A** — three-layer model with a new `workstream_id` (cleanest, but reverses two RFC 0082 decisions and forces a backfill migration) → **deferred, not discarded**.
- **B** — keep `scope_id` as the Workstream identity and partition key; add a separate, additive, opt-in **Project** layer that carries the *shared* context → **adopted**.
- **C** — documentation-only clarification → **rejected** (no modeled capability).

Under B, **isolation stays on the Workstream** (`scope_id` ≡ Workstream: its own Sources / Memory head / linear Handoff history / Statistics) and **sharing lives on the Project** (`project_id → Memory Artifact`, read-only overlay from any member Workstream's Session). Both hold simultaneously — which a single overloaded `scope_id` could not express. No `workstream_id` this version; **zero migration, no Core Protocol change**; ungrouped scopes behave exactly as today.

## Scope of this RFC

- Reconciles the affected existing RFCs — **0019** (`MemoryBindingStore` one-to-one binding + the named multi-instance extension seam), **0048** (one linear history per scope, CAS conflict), **0082** (`Project → Workstream(≡scope_id)` catalog, aggregation-only rule, branch-is-not-identity).
- English and Chinese versions kept in sync (`docs/en/rfcs/` + `docs/zh/rfcs/`).

## Intentionally left Unresolved (deferred to follow-up)

- Read-composition precedence / merge / dedup policy between the two heads.
- Whether and how a Session may *write* Project context (default: read-shared only).
- Context Pack surface for Project-level reads (RFC 0028 stays single-scope for now).
- The Workstream-boundary declaration surface (`derive_scope_id` currently keys only on the repository).

## Notes for reviewers

- Filename keeps the `0000_` prefix per the RFC process; it will be renamed to this PR's number once assigned.

Refs #1219


PsiACE:
hi @larry-zy , pls only add your rfc files in this PR -> edit the target branch to `oceanbase:master`

larry-zy:
I've realized the problem — sorry, I'm fixing it now.

larry-zy:
This PR adds an RFC ("Development Context and Scope Model") proposing to resolve scope_id's three-way semantic overload by keeping scope_id as the Workstream identity (and Runtime partition key) and introducing an explicit application-layer Project to carry shared context, without adding a new workstream_id. It is a documentation-only change: no source code is modified. The only additions are two synchronized RFC files, in English and Chinese,  under docs/en/rfcs/ and docs/zh/rfcs/.

PsiACE:
I wonder if Alternative B can be simplified by treating Project as a typed scope, while keeping it as a domain abstraction:

```text
Project(scope=P)
  +-- Workstream(scope=A)
  +-- Workstream(scope=B)
```

The existing catalog already provides the `A/B -> P` relationship. If `project_id` uses the Runtime scope identity, Project context can reuse the current scoped Source and Memory path:

```text
A/B --publish snapshot--> Sources(P) --> Memory(P)
prepare(A) --> Memory(A) + Memory(P)
```

This avoids a separate `project_id -> Memory Artifact` binding. The membership part of the catalog should belong to the Runtime rather than Handoff Report, and Project scopes should not expose Handoff or Continue.

Could you compare this model with the current Alternative B?

larry-zy:
Thanks @PsiACE  — nice idea, and it's tempting since scope_id is already the universal partition key, so P would get
  its own partition for free. Comparing with B, three things stand out:

  1. It moves Project identity into the scope namespace. Today project_id is a server-minted prj_<uuid> (catalog.py),
  deliberately separate from scope_id (git:<host>/<path> / local:…). A plain checkout derives exactly git:<host>/<path>
  with no discriminator — branch and worktree don't change it (derive_scope_id). So for one repo you can't get both a
  Project scope and per-Workstream scopes from derivation as-is; one side needs a discriminator the current derivation
  doesn't produce, and whichever value a normal session derives is the one it writes into. B sidesteps this by keeping
  project_id out of the scope namespace entirely.
  2. prepare(A) = Memory(A) + Memory(P) is new cross-scope logic, not reuse. Every entrypoint is single-scope today
  (context.for_scope(...).prepare(...), memory.for_scope(...).search(...)), and memory heads are FK-locked to artifacts
  in the same scope_id (MEMORY_ENTRY_HEADS_TABLE). Merging two scopes needs precedence/dedup for overlapping entries
  plus cross-scope ranking — none of which exists yet. B specifies this once as a read-only overlay.
  3. "No Handoff/Continue on Project scopes" has nowhere to land yet. Scoped tables carry only an opaque scope_id —
  there's no scope-type column — and handoff.for_scope(id) / memory.for_scope(id).remember(...) accept any string, so
  nothing structurally stops a project scope from being written or handed off. B makes Project a distinct type that's
  never a Handoff scope (I5) and composes read-only, so that's structurally impossible.

  So the variant could be leaner if we settle the identity/discriminator question (1) and scope typing (3). Happy to
  record it as a "Variant: Project as a typed scope" under the alternatives with these as open questions — sound good?

PsiACE:
@larry-zy Thanks for the work here. Let’s put this RFC on hold for now while the PMC aligns on how hierarchy, isolation, and sharing should be modeled. We’ll follow up once the direction is clearer, so no further changes are needed yet.

PR #1239, opened 2026-08-14
feat(dsh): add DeepSeek Harness plugin

knqiufan:

Not linked to an existing issue.

# Rationale for this change

DeepSeek Harness is a supported agent host. PowerContext already has Codex and HTTP/MCP surfaces; this change adds a first-class DSH plugin so a running Server can recall bounded context and maintain Memory from `dsh web` without embedding Python or MCP.

- Add `integrations/dsh/plugins/powercontext` (`powercontext-dsh`): Cordis plugin that recalls via `POST /v1/context/prepare`, captures prompts as Source evidence, and exposes named `pc_*` HTTP tools. Server or transport failures fail open.
- Add `powercontext setup dsh` and `powercontext doctor dsh`. Setup installs the plugin with `dsh plugin --profile web add`. Doctor checks that dump-config lists `id: powercontext-dsh`.
- Require a built `lib/index.js` so a Git checkout can be loaded. On Windows, launch `dsh.cmd` and decode CLI output as UTF-8.
- Generate the JS operations table from `openapi/powercontext.yaml` (`make js-api-generate` / `js-api-generate-check` / `js-test`).
- Document the integration in en/zh how-to, configuration, troubleshoot, interfaces, and the root README.

Yes. New commands: `powercontext setup dsh` and `powercontext doctor dsh`. New docs: Configure DeepSeek Harness. No Server API or storage format changes. Existing Codex setup is unchanged.

# How was this change tested?

- `uv run python -m pytest tests/test_dsh_cli.py tests/e2e/test_dsh_http_chain.py tests/test_js_operations.py tests/test_system_cli.py -k dsh` (11 passed)
- `pnpm --dir integrations/dsh/plugins/powercontext test` (47 passed)
- `uv run python scripts/generate_js_operations.py --check`
- Manual on Windows with a local checkout:
  - `uv run powercontext setup dsh --source .`
  - `uv run powercontext doctor dsh` → `plugin: ok - powercontext-dsh is installed`
  - `dsh --profile web --dump-config` lists `id: powercontext-dsh`
  - `dsh web` on `http://127.0.0.1:3080` returns HTTP 200
  - `curl` `/health/live`, `/health/ready`, `POST /v1/memory/remember`, `POST /v1/context/prepare` against a local Server


integrations/dsh/plugins/powercontext/src/recall.ts
```diff
@@ -0,0 +1,126 @@
...
+  wrapContent: (text: string) => unknown
+  log: (event: Record<string, unknown>) => void
+}
+
+export function messagesToQuery(messages: PromptMessage[]): string {
+  return messages
+    .flatMap((message) => message.content)
+    .filter((block): block is TextBlock & { text: string } => block.type === 'text' && typeof block.text === 'string')
+    .map((block) => block.text)
+    .join('')
```
Teingi, line 46: [P1] Preserve message provenance before capturing prompts

`agent/pre-step` receives the complete claimed `UserMessage[]`, which can include plugin-sourced injections and other next-step context, not only the user's prompt. `PromptMessage` drops `source`, and `messagesToQuery()` concatenates every text block; `recallThenCapture()` then persists that combined string as `event: user_prompt_submit`. With prompt capture enabled by default, this can feed plugin or tool context back into Memory as user evidence, create recall feedback loops, and retain content the user never asked PowerContext to store.

Recommended fix: preserve the full `UserMessage.source`, derive the recall query and capture payload separately, and pass only explicitly user-originated messages to `captureUserPrompt` (using clear separators between messages). Please add a composition test with one user message and one plugin-sourced message, and assert that only the user text is captured.

integrations/dsh/plugins/powercontext/src/tools.ts
```diff
@@ -0,0 +1,309 @@
...
+function callTool(runtime: PluginRuntime, defineTool: DefineTool): unknown {
+  return pcTool(defineTool, {
+    name: 'pc_call',
+    description: 'Call any PowerContext OpenAPI operation by operation_id. Do not approve candidates unless the user explicitly asked. scope_id is injected automatically when omitted.',
+    kind: 'read',
+    parameters: {
+      operation_id: {
+        type: 'string',
+        required: true,
+        enum: [...OPERATION_IDS],
```
Teingi: [P1] Do not expose the full OpenAPI surface through a read-classified tool

`OPERATION_IDS` includes administrative and destructive mutations such as `purge_handoff_report_activities`, `detach_handoff_report_workspace`, and artifact approval or rejection. Exposing all of them through `pc_call` bypasses the curated agent-facing surface used by the Server MCP projection. The tool is also presented as `kind: read`, and `injectScope()` preserves a caller-provided non-empty `scope_id`, so the generic tool can target a scope other than the one derived for the current workspace.

Recommended fix: replace `OPERATION_IDS` here with an explicit agent allowlist aligned with the intended Server MCP surface (or remove `pc_call` entirely), exclude administrative and destructive operations, and overwrite `scope_id` with the derived current scope for every allowed scoped operation. If a mutation is genuinely required, expose it as a named mutation tool with the correct presentation and an explicit user-authorization rule. Add tests proving destructive operation IDs are unavailable and a caller-supplied `scope_id` cannot override the current scope.

Teingi (changes requested):
Thanks for adding the DeepSeek Harness integration. I am requesting changes for two blocking issues: prompt capture currently loses message provenance, and the generic OpenAPI tool exposes operations beyond the intended agent-facing surface. Please address the inline comments before merging.

integrations/dsh/plugins/powercontext/src/tools.ts
knqiufan: Thank you for calling out the boundary mismatch. Fixed in 490da63: I removed pc_call entirely, so administrative, destructive, and review-mutation OpenAPI operations are no longer model-visible. The retained mutations are explicit named tools with edit/delete presentation and DSH one-time user approval, failing closed when no approval channel is available. Scoped operations now always overwrite caller-provided scope_id with the workspace-derived scope. Regression tests cover the exact tool surface, destructive IDs being absent, approval routing, and scope override.

integrations/dsh/plugins/powercontext/src/recall.ts
knqiufan: Thank you for the careful review. Fixed in 490da63: the pre-step path now preserves UserMessage.source, uses the full claimed batch only for recall, and captures only messages whose source.kind is user, with explicit blank-line separators. I also added mixed user/plugin and plugin-only regression coverage so injected context cannot be persisted as user_prompt_submit.

Teingi (approved):
LGTM

PR #1245, opened 2026-08-16
feat(tracing): trace scheduled Runtime processing (#1243)

### LINKED ISSUE #1243: feat: trace scheduled Runtime processing
## Feature description

Add tracing for scheduled Source-window processing and Experience incubation.

Scheduled processing currently reports outcomes through logs but has no trace root.

Possible trace structure:

```text
scheduled.process_source_window
└── memory.flush

scheduled.incubate_experience_candidates
└── experience.incubation
```

The trace should distinguish successful, no-op, failed, and cancelled runs. It must not inherit an unrelated HTTP or MCP context or record scope IDs or content.

## Alternatives considered

Keep scheduled processing observable through logs only.

## Additional context

Part of #1214. Tracing remains optional and must not change scheduler or Runtime behavior.

guozhihao-224:

Closes #1243. Parent: #1214. Aligns scheduled tracing with RFC 0046.

# Rationale for this change

Scheduled Source-window processing and Experience incubation reported outcomes only through structured logs (`background.operation.completed`). With no trace root, a scheduled run was invisible in traces unless it happened to join an unrelated HTTP or MCP context. RFC 0046 requires background activations to start their own trace when tracing is enabled.

- Extend the existing OTel-free `RuntimeTracing` port with `background()` (fresh trace root, `unit=background`) and `RuntimeSpan.set_outcome()` so domain code can record `success` / `noop` / `failure` without changing control flow. This port is an internal Runtime hook, not a public extension API (RFC 0046).
- Implement `ServerTracing.background()` with an empty OTel `Context()` attached for the whole block, including when the root span fails to start, so child stages cannot inherit an inbound HTTP/MCP span.
- Wrap each scoped scheduled activation in `scheduled.process_source_window` / `scheduled.incubate_experience_candidates`, and add `memory.flush` / `experience.incubation` stages on the shared `flush()` / `incubate()` boundaries (the same `memory.flush` stage also appears under `HTTP flush_memory`).
- Outcomes are `success`, `noop`, `failure`, or `cancelled`, matching the existing logs. `failure` sets span status `ERROR`; the others stay `UNSET`. Attributes are bounded counts only — no `scope_id`, request ID, or Memory content.
- Document the scheduled tree in the Phoenix how-to (EN/ZH) and note `unit=background` in RFC 0046.

Observability-only. When tracing is enabled, scheduled activations emit independent traces. No breaking API or persisted-format changes. Tracing stays optional; with no tracer injected, scheduler and Runtime control flow are unchanged. Custom `RuntimeTracing` adapters are not a supported public surface in this release.

# How was this change tested?

- `uv run pytest tests/builtin/runtime/test_scheduler.py tests/e2e/test_observability.py tests/test_server_tracing.py`
- Unit coverage: scheduled roots, nested `memory.flush` / `experience.incubation`, outcomes `success` / `noop` / `failure` / `cancelled` for both processors, swallowed errors, isolation when the background root fails to start, and no `scope_id` in exported attributes
- E2E: real Server with a short scheduler interval; Source-window and Experience roots have `parent is None`, do not share a `trace_id` with HTTP spans, and export no scope ID or captured content

Exported span trees (issue requirement), from the e2e `InMemorySpanExporter`:

```text
scheduled.process_source_window
  powercontext.operation.name=process_source_window
  powercontext.operation.unit=background
  powercontext.operation.outcome=success
  powercontext.background.source_count=1
└── memory.flush
      powercontext.operation.unit=stage
      powercontext.operation.outcome=success
      powercontext.memory.flush.source_count=1

scheduled.incubate_experience_candidates
  powercontext.operation.name=incubate_experience_candidates
  powercontext.operation.unit=background
  powercontext.operation.outcome=success
  powercontext.background.source_count=1
  powercontext.background.candidate_count=0
└── experience.incubation
      powercontext.operation.unit=stage
      powercontext.operation.outcome=success
      powercontext.experience.incubation.source_count=1
      powercontext.experience.incubation.candidate_count=0
```

`scope.context` / `scope.lock` remain existing setup stages and close before the operation body, so they appear as siblings of `memory.flush` / `experience.incubation`. When inference is instrumented, `invoke_agent` / `chat` nest under those stages.

# AI usage statement

Cursor (Grok 4.6) was used to implement the scheduled tracing port, tests, and docs; to review the change against #1243; and to draft this PR description. The design, issue scope, and final review of the exported trees were directed by the author.


src/powercontext/server/tracing.py
```diff
@@ -138,7 +138,37 @@ def stage(
...
+        *,
+        operation: str,
+        attributes: Mapping[str, _TraceAttribute],
+    ) -> Iterator[_ActiveSpan]:
+        """Trace one scheduled activation as an independent trace root."""
+
+        span = self.start_span(
+            name,
+            kind=SpanKind.INTERNAL,
+            context=Context(),
```
Kairo-J: If creating or attaching this root span fails, `_ActiveSpan.start()` returns without installing the empty Context. The background task then continues under any ambient HTTP/MCP context, so child spans such as `memory.flush` may attach to the wrong trace.Although the theoretical probability of this occurring is low, it violates the guidelines established in the current issues.

Could we keep an empty Context attached for the whole background block, even when root-span creation fails, and add a regression test for this failure path?

src/powercontext/builtin/runtime/protocols.py
```diff
@@ -47,6 +49,14 @@ def stage(
         attributes: Mapping[str, TraceAttribute],
     ) -> AbstractContextManager[RuntimeSpan]: ...
 
+    def background(
```
Kairo-J, line 52: Could this break existing custom `RuntimeTracing` adapters?

Previously an adapter only needed `stage()`, and its span only needed `set_attributes()`. The Runtime now calls `background()` and `set_outcome()` unconditionally, so an older adapter would raise `AttributeError`. For `flush()` / `incubate()`, this may happen after the underlying operation has already completed.

Do we know whether there are many users currently implementing and injecting their own tracing adapters? If this is a supported extension point, could we add a backward-compatible fallback or document it as a breaking change? If it is strictly internal, it would be helpful to clarify that as well.

tests/test_server_tracing.py
```diff
@@ -230,6 +230,50 @@ def start_span(self, *_args: object, **_kwargs: object) -> None:
...
+    tracing, exporter = _tracing()
+
+    with tracing.stage("memory.flush", attributes={}) as noop:
+        noop.set_outcome("noop")
+    with tracing.stage("memory.flush", attributes={}) as failed:
+        failed.set_outcome("failure")
+
+    spans = [span for span in exporter.get_finished_spans() if span.name == "memory.flush"]
+    assert len(spans) == 2
+    outcomes = {span.attributes["powercontext.operation.outcome"] for span in spans if span.attributes is not None}
```
Kairo-J, line 244: `ruff format --check` fails on this generator expression, which causes `make check` and CI to fail. Please run `uv run ruff format tests/test_server_tracing.py` (or `uv run prek run -a`) and commit the formatting change.

tests/builtin/runtime/test_scheduler.py
```diff
@@ -222,6 +284,162 @@ async def scenario() -> None:
...
+            await task
+
+    asyncio.run(scenario())
+
+    spans = {span.name: span for span in exporter.get_finished_spans()}
+    root = spans["scheduled.process_source_window"]
+    assert root.attributes is not None and root.attributes["powercontext.operation.outcome"] == "cancelled"
+
+
+def test_scheduled_experience_records_root_and_incubation_spans() -> None:
```
Kairo-J, line 388: The PR description states that both processors have unit coverage for `success`, `noop`, `failure`, and `cancelled`. The Experience processor currently has success, failure, and cancellation tests, but no noop case. Please add an Experience noop test that asserts both the scheduled root and `experience.incubation` report `noop`, or update the PR description to match the actual coverage.

Kairo-J:
The PR description mentions that the exported Span tree can be regenerated using `scripts/trace_scheduled_demo.sh --phoenix`, yet the script is missing from the latest commit (HEAD) of the PR.

I noticed that the second commit removed it. Was there a specific reason for this? Could you please clarify, update the PR description, or add the script back?

Ethan-Xingyue:
plz fix ci

guozhihao-224:
> plz fix ci

hi @Ethan-Xingyue  The Main failure was `ruff format --check` on `tests/test_server_tracing.py`. That formatting is fixed in the follow-up commit together with the background isolation and Experience noop coverage.

guozhihao-224:
> The PR description mentions that the exported Span tree can be regenerated using `scripts/trace_scheduled_demo.sh --phoenix`, yet the script is missing from the latest commit (HEAD) of the PR.
> 
> I noticed that the second commit removed it. Was there a specific reason for this? Could you please clarify, update the PR description, or add the script back?

ha @Kairo-J  The script was only a local Phoenix helper and was dropped on purpose so it would not ship in the PR. The exported trees in the description come from the e2e `InMemorySpanExporter`. I will remove the `scripts/trace_scheduled_demo.sh --phoenix` lines from the PR body so it matches HEAD.

tests/builtin/runtime/test_scheduler.py
guozhihao-224: Added `test_scheduled_experience_records_noop_outcome`, which asserts `noop` on both `scheduled.incubate_experience_candidates` and `experience.incubation` (including the bounded counts). Source-window and Experience now cover success / noop / failure / cancelled.

tests/test_server_tracing.py
guozhihao-224: Fixed. The generator in `test_runtime_stage_records_explicit_noop_and_failure_outcomes` is wrapped to satisfy `ruff format --check`.

src/powercontext/builtin/runtime/protocols.py
guozhihao-224: `RuntimeTracing` is an internal Runtime port, not a supported public extension API (RFC 0046: "Internal observability hooks are not public extension APIs in the initial release"). The only in-tree implementation is `ServerTracing`. We are treating this as an additive change to that internal port rather than a user-facing break, and we are not adding a runtime fallback for older adapters.

src/powercontext/server/tracing.py
guozhihao-224: Agreed. `background()` now attaches an empty `Context()` for the whole block (including when `_ActiveSpan.start()` fails) and detaches it in `finally`. Child stages such as `memory.flush` therefore cannot join an ambient HTTP/MCP trace. Covered by `test_background_isolates_child_spans_when_root_start_fails`.

Ethan-Xingyue:
plz fix ci

Kairo-J (approved):

Kairo-J:
lgtm

PR #1252, opened 2026-08-17
feat(tracing): add memory read-path spans

### LINKED ISSUE #1241: feat: trace the Memory read path
## Feature description

Add bounded tracing for Memory search, reranking, and context preparation.

`search_memory` and `prepare_context` currently expose only the outer application span. Add enough detail to show which read stage ran, how many candidates were handled, and where latency or failure occurred.

Possible trace structure:

```text
powercontext search_memory
└── memory.search
    ├── embeddings <model>
    └── memory.rerank
        └── invoke_agent memory_rerank

powercontext prepare_context
├── memory.search
├── experience.search
└── context.prepare
```

Do not record queries, prompts, Memory content, or vectors.

## Alternatives considered

Keep the existing application span only.

## Additional context

Part of #1214. Tracing remains optional and must not change Runtime behavior.

Kairo-J:

Closes #1241

# Rationale for this change

Memory read requests currently expose only top-level application spans, making it difficult to determine whether latency comes from retrieval, embeddings, reranking, Experience recall, or context assembly.

This change adds stage-level tracing while keeping OpenTelemetry owned by the Server layer. The Runtime remains framework-neutral, and tracing failures do not affect search behavior.

- Add internal, framework-neutral `RuntimeSpan` and `RuntimeTracing` protocols.
- Inject optional Runtime tracing through `BuiltinRuntime`.
- Implement `ServerTracing.stage()` using OpenTelemetry `INTERNAL` spans.
- Record `success`, `failure`, and `cancelled` outcomes while preserving original exceptions.
- Add the following read-path spans:
  - `memory.search`
  - `memory.rerank`
  - `experience.search`
  - `context.prepare`
- Emit stable zero-result spans when Memory or Experience configuration is absent.
- Wrap actual reranker calls without changing reranking behavior or `policy_id`.
- Export only bounded booleans, enums, and counts.
- Do not record queries, content, vectors, IDs, or exception messages.
- Add unit and end-to-end coverage for span hierarchy, privacy, cancellation, disabled tracing, and tracing failure isolation.

Validated span trees:

```text
powercontext search_memory
└── memory.search
    ├── embeddings <model>
    └── memory.rerank
        └── invoke_agent memory_rerank

powercontext prepare_context
├── memory.search
├── experience.search
└── context.prepare
```

Operators with tracing enabled will see the additional internal spans and bounded attributes.

Search results and error behavior remain unchanged. This PR does not change public APIs, OpenAPI, CLI behavior, database schemas, persisted formats, or dependencies.

# How was this change tested?

```text
make test
486 passed, 7 skipped

make check
Lock-file validation, pre-commit checks, Ruff formatting/linting, and ty type checking passed.
```

The tests also cover missing Memory, unconfigured Experience recall, empty results, reranker fallback, embedding and agent span nesting, sensitive-data exclusion, cancellation, and tracing failure isolation.

# AI usage statement

OpenAI Codex (GPT-5.6-Sol) was used to help analyze the issue, assist with making changes, write tests.


src/powercontext/server/tracing.py
```diff
@@ -70,6 +73,34 @@ def start_span(
...
+    ) -> Iterator[_ActiveSpan]:
+        """Trace one bounded Runtime stage without changing its behavior."""
+
+        span = self.start_span(
+            name,
+            kind=SpanKind.INTERNAL,
+            attributes={
+                **attributes,
+                "powercontext.operation.name": name,
+                "powercontext.operation.unit": "stage",
```
Ethan-Xingyue, line 113: This adds a fifth value for powercontext.operation.unit, but the docs that define the vocabulary aren't updated: the unit table in docs/{en,zh}/rfcs/0046_observability_foundations.md 
and the span table in docs/{en,zh}/docs/how-to/trace-with-phoenix.md 
.

src/powercontext/builtin/runtime/composition.py
```diff
@@ -100,6 +103,44 @@ def project_source(self, source: Source, /) -> JsonValue:
...
+
+    def __init__(self, delegate: MemoryReranker, tracing: RuntimeTracing) -> None:
+        self._delegate = delegate
+        self._tracing = tracing
+
+    @property
+    def policy_id(self) -> str:
+        return self._delegate.policy_id
+
+    @policy_id.setter
```
Ethan-Xingyue: The policy_id setter has no callers — policy_id is only ever read, at service.py. I removed locally: uv run ty check stays green and tests/e2e/test_builtin_runtime.py + tests/e2e/test_observability.py still pass (8 passed). The getter is needed — _DynamicPolicyReranker covers the dynamic read — just not the setter.


What is the purpose of adding this setter?

src/powercontext/builtin/runtime/application.py
```diff
@@ -261,32 +274,72 @@ async def prepare(self, request: PrepareContextRequest, /) -> PreparedContext:
...
-                if self._runtime._experience_recall is None
-                else await self._runtime._experience_recall(
-                    self.scope_id,
-                    request.query,
-                    builder.experience_candidate_limit,
+                if span is not None:
+                    span.set_attributes({"powercontext.experience.search.result_count": len(experience_hits)})
+
+            with self._runtime._stage(
+                "context.prepare",
```
Ethan-Xingyue: Minor naming thought: this sits right next to the existing powercontext prepare_context span (app.py:1299), which covers the whole request, while this one wraps just the synchronous builder.build_result(). Might be worth something like context.assemble or context.build to make the difference obvious in a trace viewer?

Ethan-Xingyue:
Just a thought, not a blocker: _context() (which awaits self._provider.get() and opens a DB session) and the wait on _lock(scope_id) both happen before the first stage span opens, so they don't show up in the trace. Since the goal here is locating latency in the read path, and test_same_scope_read_only_searches_do_not_serialize_reranking suggests lock contention is already on your radar, those might be interesting to cover eventually.

Ethan-Xingyue:
Those are my thoughts so far; let's wait for @PsiACE's review before settling on the specifics.

src/powercontext/builtin/runtime/composition.py
Kairo-J: The setter was added because `MemoryReranker` declares `policy_id` as a read-write property, and the tracing decorator needs to fully proxy this interface.
It exists to forward future assignments made to the wrapper's `policy_id` to the underlying reranker; it is not required for the current business logic. Scenarios involving dynamic `policy_id` updates actually rely solely on the getter.
@Ethan-Xingyue @PsiACE Do you think it should be kept? Unless `policy_id` is explicitly defined as a read-only property, I would prefer to retain the setter functionality.

src/powercontext/server/tracing.py
Kairo-J: Get, it was an oversight on my part, and I will add it later.

src/powercontext/builtin/runtime/application.py
Kairo-J: Get, I think it could be changed to `context.build`, as `context.assemble` might misleadingly imply the inclusion of the preceding Memory/Experience search, making the scope seem a bit too broad.

src/powercontext/builtin/runtime/composition.py
PsiACE: `policy_id` is plain state, so please keep it as a regular attribute instead of proxying it through a property with a setter.

PsiACE (changes requested):
Please focus the tests on exported tracing behavior and real regressions: drop the dynamic policy_id, fake _RecordingTracing, and synthetic SystemExit cases, and cover the actual vector-search span tree plus the readiness probe's no-root-span behavior.

src/powercontext/builtin/runtime/composition.py
Kairo-J: Okay, I'll modify `_TracingMemoryReranker` to use standard properties and synchronize the latest value to the `delegate` after `rerank` completes.

Kairo-J:
> Just a thought, not a blocker: _context() (which awaits self._provider.get() and opens a DB session) and the wait on _lock(scope_id) both happen before the first stage span opens, so they don't show up in the trace. Since the goal here is locating latency in the read path, and test_same_scope_read_only_searches_do_not_serialize_reranking suggests lock contention is already on your radar, those might be interesting to cover eventually.

@Ethan-Xingyue 
Thanks, that’s indeed a great angle. Currently, the lock wait time during context resolution and the `prepare_context` phase appears in the external application Span merely as "unattributed exclusive time."

I noticed a detail here: the default `RelationalContexts.get()` (`_context()` line 1173) simply invokes the `Provider` to resolve and cache scoped services (`RelationalContexts.get()` line 434) without initiating a database session; the actual database read operation still occurs within `memory.search`. However, the skew caused by delayed attribution is indeed an issue.

Given that context retrieval and lock waiting are runtime concerns spanning multiple stages—whereas the scope of this PR is limited to the memory read phase—I intend to keep the PR focused.

We can add `context.resolve` and `scope.lock.wait` as sibling Spans in future work without altering the semantics of the Span introduced here. Thanks for pointing this out.

Kairo-J:
@Ethan-Xingyue @PsiACE  When adding ready regression coverage, I encountered an additional special case that I felt was worth mentioning.

The previous suppression mechanism relied on attaching an unsampled synthetic parent model during ready testing. This worked for the default `ParentBased` sampler, but injected embedding models using an effective `ALWAYS_ON` sampler ignored the parent model's sampling decisions and still exported the `embeddings<model>` span. Since the synthetic parent model was never exported, this resulted in incomplete/isolated ready tracking.

I fixed this by providing a separate, uninstrumented ready copy of the injected `PydanticAIEmbeddingModel` instance:

- The adapter and its `Embedder` are shallow-copied;

- `instrument=False` is applied only to the ready copy;

- The original adapter still retains the instrumentation used for manipulating the search;

- Subclass behavior and existing providers/settings/state are preserved;

- Since no shared objects are temporarily modified, concurrent ready state calls and search calls do not disable each other's tracing.

Regression tests now use `TracerProvider(sampler=ALWAYS_ON)` via the public `create_server_app()` injection path. It first verifies whether the ready state derives any inference span, then performs a real HTTP vector search and verifies that the operation trace still produces the following result:

```
powercontext search_memory
└── memory.search
    └── embeddings test
```

The entire test suite passed 484 tests, skipped 7 tests, and both `make check` and `make docs-test` passed.

src/powercontext/server/tracing.py
```diff
@@ -70,6 +95,56 @@ def start_span(
...
+        try:
+            parent = trace.NonRecordingSpan(
+                trace.SpanContext(
+                    trace_id=_ID_GENERATOR.generate_trace_id(),
+                    span_id=_ID_GENERATOR.generate_span_id(),
+                    is_remote=False,
+                    trace_flags=trace.TraceFlags(0),
+                )
+            )
+            context_token = otel_context.attach(set_span_in_context(parent))
```
PsiACE: Readiness should stay outside tracing, but attaching a synthetic parent makes tracing setup part of the readiness path and can abort server startup. Please make it best-effort.

Kairo-J: get and fixed

PsiACE (approved):
lgtm, thanks

PR #1253, opened 2026-08-17
fix(memory): prevent stale-head search misses during concurrent append

### LINKED ISSUE #1249: bug: concurrent Memory append silently drops valid FTS search hits
### Describe the bug

A Memory FTS search can silently return a false negative when another request appends a memory entry concurrently.
The search first reads and validates head revision N. If an append then commits revision N+1 before the FTS query executes, the mutable head index projection is replaced with rows belonging to revision N+1. The pending search queries those new rows but still filters them using revision N, causing otherwise valid hits to be discarded.
The response therefore reports revision N with an empty hit list, even though the searched fact exists in both revisions N and N+1. No exception, retry signal, or stale-result indication is returned.
This is a deterministic TOCTOU race between head validation and index lookup, and can cause intermittent memory recall failures in applications that search while appending memories.

### Steps to reproduce

1. Use a PowerContext checkout with its dependencies installed.
2. Save the following as repro_memory_search_race.py:

```python
#!/usr/bin/env python3

import asyncio
import json
from types import MethodType
from typing import Any

from powercontext.builtin.artifacts.memory import MemoryEntryInput
from powercontext.builtin.runtime import (
    BuiltinConfig,
    RememberMemoryRequest,
    SearchMemoryRequest,
    open_builtin_runtime,
)


async def main() -> None:
        database = SQLiteConfig(url=f"sqlite+aiosqlite:///{directory}/repro.db")
        async with open_builtin_runtime(BuiltinConfig(database=database)) as runtime:
            memory = runtime.memory.for_scope("race-repro")
            await memory.remember(
                RememberMemoryRequest(entries=(MemoryEntryInput(kind="fact", text="Stable searchable fact."),))
            )

            provider: Any = runtime._provider
            fts = provider.index.indexes[0]
            original_search = fts.search
            paused, resume = asyncio.Event(), asyncio.Event()

            async def pause_search(self: Any, connection: Any, scope: str, request: Any) -> Any:
                paused.set()
                await resume.wait()
                return await original_search(connection, scope, request)

            fts.search = MethodType(pause_search, fts)
            query = SearchMemoryRequest(query="stable searchable", mode="fts")
            try:
                pending_search = asyncio.create_task(memory.search(query))
                await asyncio.wait_for(paused.wait(), 2)
                new_head = await memory.remember(
                    RememberMemoryRequest(entries=(MemoryEntryInput(kind="fact", text="Unrelated appended fact."),))
                )
                resume.set()
                raced = await asyncio.wait_for(pending_search, 2)
            finally:
                resume.set()
                fts.search = original_search

            normal = await memory.search(query)
            observed = {
                "search_revision": raced.memory_ref.revision if raced.memory_ref else None,
                "new_head_revision": new_head.memory_ref.revision,
                "raced_hits": [hit.text for hit in raced.hits],
                "post_race_hits": [hit.text for hit in normal.hits],
            }

    print(json.dumps(observed, sort_keys=True))
    expected = {
        "search_revision": 1,
        "new_head_revision": 2,
        "raced_hits": [],
        "post_race_hits": ["Stable searchable fact."],
    }
    if observed != expected:
        raise SystemExit(1)
    print("0")


if __name__ == "__main__":
    asyncio.run(main())
```

3. Run the reproducer:

```bash
python repro_memory_search_race.py
```

4. Observe that the search associated with revision 1 returns no hits, while the same query immediately succeeds after the concurrent append.

The method wrapper only introduces a deterministic scheduling barrier. It delegates to the original FTS implementation without modifying the request, indexed data, or returned results.

### Expected behavior

The search should operate against one consistent revision.
Either:
- it completes against revision 1 and returns Stable searchable fact., or
- it detects that the head advanced, retries against revision 2, and returns the same fact.

Because the concurrent append adds an unrelated fact and does not remove the existing one, the searched fact is present in both revisions. The search must not silently return a successful empty result.

### Actual behavior

The reproducer consistently prints:
```
{"new_head_revision": 2, "post_race_hits": ["Stable searchable fact."], "raced_hits": [], "search_revision": 1}
0
```

The raced search reports revision 1 but returns:
```
"raced_hits": []
```

The identical query immediately afterward returns:
```
"post_race_hits": ["Stable searchable fact."]
```

This demonstrates that the empty result is not caused by missing data, query mismatch, or FTS tokenization. It is caused by the search combining the old revision with the new head projection.
The final 0 and process exit code 0 mean that the reproducer observed the known buggy state.

### Environment

- PowerContext version: 0.0.1
- Commit: 72b9417118375c5228d73007f54674f5544a042a
- Python version: 3.14.3
- OS: macOS 26.2, arm64
- aiosqlite: 0.22.1
- SQLAlchemy: 2.0.51
- Storage backend: local SQLite
- External services: none

thunguo:

Closes #1249 .

# Rationale for this change

Memory search had a TOCTOU race during concurrent appends. A search could validate revision N, then query the mutable index projection after another request had advanced the head to revision N+1. The results would still be filtered using revision N, potentially producing an empty result even when an unchanged matching entry existed.

This failure was silent: callers received a successful response associated with revision N but no indication that the result was inconsistent.

- Validate requested memory revisions before and after the index query.
- Detect when the memory head advances during index lookup or reranking.
- Retry the search with the latest head for a bounded number of attempts.
- Raise `RevisionConflictError` instead of returning potentially inconsistent results if concurrent updates continue beyond the retry limit.
- Add a deterministic concurrency regression test using synchronization barriers.
- Cover SQLite and OceanBase across FTS, vector, and hybrid search modes.

Yes. Under sustained concurrent Memory updates, `POST /v1/memory/search` may now return `409 revision_conflict` when a consistent revision cannot be obtained within the bounded retry limit. This response is now declared in the OpenAPI contract. There are no request, success-response, or persisted-format changes.

# How was this change tested?

Added `tests/e2e/test_memory_search_concurrency.py`, which deterministically pauses a search after initial head validation but before the index query, performs a concurrent append, and verifies that:

- the search retries against the new head;
- the unchanged matching entry is returned;
- the returned `memory_ref` points to the new revision;
- the expected FTS, vector, and hybrid channels are used.

Code quality checks:

```bash
make check
```

Result: all formatting, lint, and type checks passed.

Full test suite:

```bash
make test
```

Result:

```text
534 passed, 12 skipped
```

The OceanBase matrix was tested against a temporary real OceanBase CE 4.3.5.6 instance:

```bash
POWERCONTEXT_TEST_OCEANBASE_URL="<dedicated-oceanbase-url>" \
  .venv/bin/python -m pytest \
  tests/e2e/test_memory_search_concurrency.py \
  -k oceanbase -q
```

Result:

```text
3 passed, 3 deselected in 22.71s
```

The three passing OceanBase cases cover FTS, vector, and hybrid search.

# AI usage statement

### DIFF
--- a/src/powercontext/builtin/runtime/application.py
+++ b/src/powercontext/builtin/runtime/application.py
@@ -680,21 +680,12 @@ class ScopedMemoryApplication:
                     attempt += 1
                     continue
 
-                # Reranking may await model I/O, so verify optimistically instead
-                # of serializing the complete search under the scope write lock.
-                latest = await _head_or_none(service, context.artifacts.memory_artifact_id)
-                if latest is not None and latest.as_ref() == current.as_ref():
-                    return MemorySearchPage(
-                        memory_ref=current.as_ref(),
-                        mode=result.mode,
-                        hits=result.hits,
-                        rerank=result.rerank,
-                    )
-                if latest is None:
-                    raise ArtifactNotFoundError(current.as_ref())
-                if attempt == _MEMORY_SEARCH_ATTEMPTS:
-                    raise RevisionConflictError(current, latest)
-                attempt += 1
+                return MemorySearchPage(
+                    memory_ref=current.as_ref(),
+                    mode=result.mode,
+                    hits=result.hits,
+                    rerank=result.rerank,
+                )
 
     async def list(self, *, include_inactive: bool = False) -> MemoryEntriesPage:
         async with self._runtime._context(self.scope_id) as context:
--- a/src/powercontext/builtin/runtime/application.py
+++ b/src/powercontext/builtin/runtime/application.py
@@ -36,7 +36,11 @@ from powercontext.builtin.artifacts.memory import (
     MemoryEntryVersion,
     MemoryService,
 )
-from powercontext.builtin.artifacts.memory.errors import MemoryEntryNotFoundError
+from powercontext.builtin.artifacts.memory.errors import (
+    CapabilityNotSupportedError,
+    InvalidMemoryCitationError,
+    MemoryEntryNotFoundError,
+)
 from powercontext.builtin.artifacts.skill import (
     ExternalSkillRegistryUnavailableError,
     ExternalSkillResolution,
@@ -127,6 +131,7 @@ ExperienceRecall = Callable[[str, str, int], Awaitable[tuple[ExperienceSearchHit
 StatisticsServiceFactory = Callable[[str], RelationalScopedStatistics]
 RecallTokenEstimator = Callable[[str, PreparedContextBuild], Awaitable[RecallTokenMeasurement | None]]
 Clock = Callable[[], datetime]
+_MEMORY_SEARCH_ATTEMPTS = 3
 
 
 class _RuntimeConfigurationError(ValueError):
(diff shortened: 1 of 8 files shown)


src/powercontext/builtin/runtime/application.py
```diff
@@ -654,21 +659,33 @@ async def search(self, request: SearchMemoryRequest, /) -> MemorySearchPage:
...
+                        memories=(current,),
+                        limit=request.limit,
+                        mode=request.mode,
+                    )
+                except (CapabilityNotSupportedError, InvalidMemoryCitationError) as error:
+                    latest = await _head_or_none(service, context.artifacts.memory_artifact_id)
+                    if not _is_stale_memory_search(error) or latest is None or latest.as_ref() == current.as_ref():
+                        raise
+                    if attempt == _MEMORY_SEARCH_ATTEMPTS:
+                        raise RevisionConflictError(current, latest) from error
```
Teingi, line 679: Could we update the API contract for this new failure mode? Once the retries are exhausted, this becomes a `RevisionConflictError`, and the server returns `409 revision_conflict`. `POST /v1/memory/search` still does not declare a 409 response in `openapi/powercontext.yaml`, so generated clients and the docs will not know it can happen. Please add the `Conflict` response, regenerate the API artifacts, and add a server-level test that checks the 409 response. The PR description should call this out as a user-facing change too.

Teingi (changes requested):
Nice fix overall. I verified the race against a real OceanBase instance in FTS, vector, and hybrid modes. There is one API contract update needed before this lands.

src/powercontext/builtin/runtime/application.py
thunguo: Okay, I'll fix it later

thunguo: Done.

Teingi (approved):
LGTM

PR #1261, opened 2026-08-18
feat(hermes): implement powercontext plugin for hermes

### LINKED ISSUE #1260: feat(Hermes): research on Hermes memory mechanism and how to integrate PowerContext as a Hermes Memory Provider
## Background

This issue investigates Hermes Agent's memory mechanism and evaluates the feasibility and implementation path for integrating PowerContext as a standard Hermes Memory Provider.

<details>
<summary>1. Hermes Memory Mechanism</summary>

### 1.1 Memory layers and responsibilities

Hermes' memory mechanism supports the Agent's complete runtime workflow and includes:

- **Built-in memory**: Stored in files such as `MEMORY.md` and `USER.md`, managed by Hermes and injected during session initialization or system prompt construction.
- **External Memory Provider**: An external memory system integrated through the standard Provider interface, responsible for memory retrieval, context injection, session synchronization, memory extraction, and write mirroring.
- **Runtime tool calls**: A Provider can expose memory search, read, and write tools to the Agent, allowing the Agent to retrieve relevant historical memories when needed.
- **Session lifecycle hooks**: A Provider can participate in session start, per-turn processing, pre-compression processing, memory writes, session end, and other lifecycle events.

### 1.2 Core Provider interfaces

The standard Memory Provider should support the following capabilities:

- `name`
- `is_available()`
- `initialize()`
- `get_tool_schemas()`
- `handle_tool_call()`
- `system_prompt_block()`
- `prefetch()`
- `queue_prefetch()`
- `sync_turn()`
- `on_session_end()`
- `on_pre_compress()`
- `on_memory_write()`
- `shutdown()`

Their intended responsibilities are:

- `system_prompt_block()`: Inject general long-term memories that are suitable for direct inclusion in the system prompt.
- `prefetch()` / `queue_prefetch()`: Prefetch memories relevant to the current turn.
- `handle_tool_call()`: Handle memory operations explicitly requested by the Agent at runtime.
- `sync_turn()`: Synchronize the current conversation turn with the external memory system.
- `on_session_end()`: Extract or persist memories after a session ends.
- `on_pre_compress()`: Process important memories before the conversation context is compressed.

```
MEMORY.md / USER.md
    = key facts，inject when create a new session as system prompt

session_search
    = historical sessions，will be searched on demand   <--  this is where powercontext take place
```

</details>

<details>
<summary>2. PowerContext's Position in Hermes</summary>

### 2.1 Overall positioning

PowerContext should serve as Hermes' **external long-term memory storage and retrieval layer**, integrated through Hermes' standard Memory Provider interface.

PowerContext should not replace Hermes' Agent orchestration, system prompt construction, tool scheduling, or session lifecycle management. Hermes remains the host and scheduler of the memory mechanism, while PowerContext handles memory storage, indexing, retrieval, organization/compression, and lifecycle operations.

### 2.2 Relationship with Hermes built-in memory

The recommended approach is additive integration:

- Hermes built-in memory remains responsible for stable, low-latency core user information and Agent rules.
- PowerContext handles larger-scale, searchable, continuously accumulated historical memories and context.
- Hermes injects a small amount of general long-term memory into the system prompt.
- More specific, lengthy, or task-relevant historical memories are provided by PowerContext through per-turn prefetching or explicit Agent tool calls.
- Writes to Hermes built-in memory can be mirrored to PowerContext through Provider hooks to prevent long-term divergence between the two memory stores.

### 2.3 PowerContext responsibilities

- Store long-term memories generated by Hermes sessions, users, and Agents.
- Retrieve relevant memories based on the current user input and session context.
- Organize, compress, and semantically store historical context.
- Support memory writing, searching, reading, updating, retirement, and change synchronization.
- Preserve important information before context compression to reduce information loss in long sessions.
- Expose memory tools that can be called by the Hermes Agent.

### 2.4 Recommended data flow

1. Hermes starts a session and the Provider initializes the PowerContext client.
2. The Provider retrieves general long-term memories and injects them through `system_prompt_block()`.
3. When each user turn arrives, the Provider prefetches relevant memories based on the current context.
4. When the Agent needs more specific historical information, it queries PowerContext through memory tools.
5. After the turn, the Provider synchronizes the conversation or structured memories to PowerContext.
6. At session end, the Provider triggers memory extraction, organization, and persistence.
7. Before context compression, the Provider preserves important information and makes the necessary summary available after compression.

</details>

<details>
<summary>3. Implementation Plan: PowerContext as a Standard Hermes Memory Provider Plugin</summary>

Hermes officially support third party memory tool to integrate in it through Standard Hermes Memory Provider Plugin

### 3.1 Plugin structure

Add an independent Provider plugin to Hermes:

```text
plugins/memory/powercontext/
├── __init__.py
├── provider.py
├── config.py
├── client.py
└── README.md
```

Responsibilities:

- `provider.py`: Implement the Hermes `MemoryProvider` interface.
- `client.py`: Encapsulate the PowerContext HTTP API.
- `config.py`: Define configuration fields, validation, and the configuration schema.
- `README.md`: Document installation, configuration, usage, and troubleshooting.

### 3.2 Mapping Provider capabilities to PowerContext APIs

| Hermes capability | PowerContext API / mechanism |
| --- | --- |
| Relevant memory retrieval | `/v1/memory/search` |
| Memory writing | `/v1/memory/remember` |
| Memory retrieval by entry ID | `/v1/memory/entries/get` |
| Memory retirement | `/v1/memory/entries/retire` |
| Session context preparation | `/v1/context/prepare` |
| Session memory flush | `/v1/memory/flush` |
| Memory change synchronization | `/v1/memory/changes` |
| Raw content ingestion | `/v1/sources/content` |

### 3.3 Configuration design

The Provider configuration should include:

- PowerContext service URL.
- Authentication method and credentials.
- User, Agent, session, and profile isolation identifiers.
- Retrieval count, similarity threshold, and maximum injection token budget.
- Whether to enable long-term memory injection into the system prompt.
- Whether to enable per-turn prefetching.
- Whether to expose Agent memory tools.
- Whether to automatically extract and write memories at session end.
- Network timeout, retry count, and failure fallback strategy.

Authentication credentials must be managed through Hermes configuration or environment variables and must not be committed to the repository or written to logs.

#### Initialization

- Validate the configuration and create a PowerContext client.
- Use `is_available()` for local configuration checks without requiring a network health probe.
- Establish the required session/profile context during `initialize()`.

#### System prompt injection

- Retrieve a small set of general, stable long-term memories from PowerContext.
- Inject only high-confidence, low-redundancy memories.
- Enforce a token limit to avoid consuming the Agent's useful context window.

#### Per-turn prefetch

- Call `/v1/context/prepare` or `/v1/memory/search` based on the current user input and necessary session context.
- Prefer asynchronous or background prefetching to avoid blocking the main conversation path.
- Cache prefetched results for the current turn.

#### Agent tools

The Provider should expose tools such as:

- `powercontext_search_memory`
- `powercontext_get_memory`
- `powercontext_remember`
- `powercontext_retire_memory`

Tool responses should include source, timestamp, relevance, and necessary memory metadata so the Agent can assess reliability and determine whether additional retrieval is needed.

#### Turn synchronization and session end

- `sync_turn()`: Synchronize the current turn asynchronously without blocking the main conversation.
- `on_memory_write()`: Mirror Hermes built-in memory writes to PowerContext.
- `on_session_end()`: Trigger memory extraction, organization, flushing, and persistence.
- Apply observable, retryable, and graceful failure handling so Provider failures do not interrupt the main Hermes workflow.

#### Pre-compression processing

- Use `on_pre_compress()` to preserve important facts, user preferences, task state, and unfinished items before compression.
- Do not write the complete raw conversation to long-term memory without filtering.
- Deduplicate, structure, and filter sensitive information before persistence.

</details>

## References

- [Hermes Memory Provider Plugin Developer Guide](https://hermes-agent.nousresearch.com/docs/zh-Hans/developer-guide/memory-provider-plugin)
- [Hermes Memory Providers User Guide](https://hermes-agent.nousresearch.com/docs/zh-Hans/user-guide/features/memory-providers)
- [PowerContext repository](https://github.com/oceanbase/powercontext)

## Expected outcomes

- Document the mapping between Hermes memory capabilities and PowerContext capabilities.
- Clarify PowerContext's responsibilities and data flow within Hermes.
- Produce an actionable design for a standard Hermes Memory Provider plugin.
- Provide a foundation for subsequent implementation, testing, and documentation.

Alanxtl:

Closes #1260

## Rationale for this change

hermes use demo
动画

hermes cli use demo
动画1

This PR implements the proposed integration between PowerContext and Hermes Agent as a standard Hermes `MemoryProvider`.

Hermes remains responsible for Agent orchestration, memory lifecycle management, session handling, and tool scheduling. PowerContext acts as the external long-term memory backend for searchable historical context, memory storage, context preparation, memory extraction, and memory lifecycle operations.

The integration is designed to be additive:

- Hermes built-in memory remains available for stable core facts and user information.
- PowerContext provides larger-scale, searchable, and continuously accumulated historical memory.
- Relevant historical context can be prefetched for the current turn.
- The Agent can explicitly search, read, save, and retire memories through provider tools.
- Important conversation context is persisted before Hermes performs context compression.

- Add a Hermes integration under `integrations/hermes/`.
- Implement the PowerContext Hermes `MemoryProvider`.
- Add a dependency-free HTTP client using only the Python standard library.
- Map Hermes lifecycle hooks to PowerContext operations:
  - `prefetch()` and `queue_prefetch()` use `/v1/context/prepare`.
  - `sync_turn()` captures completed turns through `/v1/sources/content`.
  - `on_session_end()` waits for pending writes and triggers `/v1/memory/flush`.
  - `on_pre_compress()` persists bounded conversation context before compression.
  - `on_memory_write()` mirrors Hermes built-in memory additions and replacements.
- Add Agent memory tools:
  - `powercontext_search_memory`
  - `powercontext_get_memory`
  - `powercontext_remember`
  - `powercontext_retire_memory`
- Add scope isolation based on Hermes profile, user ID, and Hermes home.
- Add configuration support through:
  - `$HERMES_HOME/powercontext.json`
  - Environment variables
  - Hermes provider configuration
- Add the `hermes powercontext` CLI command group:
  - `status`
  - `search`
  - `remember`
  - `get`
  - `retire`
  - `flush`
- Add provider configuration schema and plugin metadata.
- Add installation, configuration, runtime behavior, and CLI documentation.
- Make backend failures fail open so PowerContext outages do not interrupt the Hermes conversation.
- Add integration tests covering retrieval, turn synchronization, pre-compression persistence, memory tools, failure handling, and CLI registration.

Yes.

Users can install and enable PowerContext as a Hermes memory provider:

```yaml
memory:
  provider: powercontext
```

A running PowerContext server is required. The provider uses:

```text
http://127.0.0.1:8000
```

by default and supports configuration through `$HERMES_HOME/powercontext.json` or environment variables.

The following CLI commands are available after Hermes discovers the provider:

```bash
hermes powercontext --help
hermes powercontext status
hermes powercontext search "Python project management"
hermes powercontext remember preference "The user prefers uv"
hermes powercontext get ...
hermes powercontext retire ...
hermes powercontext flush
```

The integration does not add third-party runtime dependencies to Hermes and does not change Hermes core behavior when the provider is disabled.

```bash
uv run pytest tests/integrations/test_hermes_provider.py -q
uv run ruff format --check integrations/hermes tests/integrations/test_hermes_provider.py
uv run ruff check integrations/hermes tests/integrations/test_hermes_provider.py
uv run ty check
```

## AI usage statement

This PR was developed with assistance from OpenAI Codex.


integrations/hermes/plugins/powercontext/__init__.py
```diff
@@ -0,0 +1,537 @@
...
+        """Keep per-session prefetch state aligned with Hermes session changes."""
+        self._session_id = new_session_id
+        with self._prefetch_lock:
+            self._prefetch_cache.clear()
+        self._last_recall = None
+
+    def on_pre_compress(self, messages: list[dict[str, Any]]) -> str:
+        if not self._client or not self._scope_id or not messages:
+            return ""
+        content = _messages_to_text(messages, limit=_MAX_PRECOMPRESS_CHARS)
```
Teingi: Could we avoid sending the whole pre-compression transcript here? Hermes v0.20.4 passes the original `messages` list, including system and tool text, and this hook still runs when `capture_turns` is false. I reproduced that with system/user secret sentinels: both reached `capture_content()` and were flushed.

There is also a dedupe issue. Adding one message changes the content hash, so the next compression creates another source that contains all of the earlier turns again.

I suggest a separate `capture_pre_compress` option that defaults to false, filtering roles and secrets before the request, and capturing only the new turn range with a stable idempotency key. Please add tests for the disabled setting, secret filtering, and overlapping compression windows.

integrations/hermes/plugins/powercontext/__init__.py
```diff
@@ -0,0 +1,537 @@
...
+        return ""
+
+    def on_memory_write(
+        self,
+        action: str,
+        target: str,
+        content: str,
+        metadata: dict[str, Any] | None = None,
+    ) -> None:
+        if not self._executor or not self._client or action not in {"add", "replace"} or not content.strip():
```
Teingi: `replace` is currently handled as another append-only `remember`, while `remove` is dropped by this guard. That leaves stale PowerContext entries active after the user changes or deletes built-in memory. Hermes already passes `old_text` in `metadata`.

We need a stable mapping from the Hermes memory item to its PowerContext citation, then retire the old entry for both replace and remove. If that mapping is out of scope for this PR, please mirror only `add` for now and document the limitation instead of treating replace as successful.

integrations/hermes/plugins/powercontext/__init__.py
```diff
@@ -0,0 +1,537 @@
...
+            return
+        session_key = session_id or self._session_id
+
+        def prepare() -> None:
+            try:
+                response = self._client.prepare_context(
+                    self._scope_id,
+                    query[:8192],
+                    max_bytes=_as_int(
+                        self._config.get("max_bytes", _DEFAULT_MAX_BYTES),
```
Teingi: This path skips the environment-variable precedence used by `prefetch()`. With `POWERCONTEXT_HERMES_MAX_BYTES=16000`, I observed `queue_prefetch()` still send 8000 because it reads only `self._config`.

Please use `_config_value(self._config, "max_bytes", "POWERCONTEXT_HERMES_MAX_BYTES", _DEFAULT_MAX_BYTES)` here as well and add a small env-override test.

integrations/hermes/plugins/powercontext/__init__.py
```diff
@@ -0,0 +1,537 @@
...
+        try:
+            barrier.result(timeout=_as_float(self._config.get("shutdown_timeout", 10), 10.0))
+        except Exception:
+            logger.debug("PowerContext background work did not finish before the barrier", exc_info=True)
+
+    def shutdown(self) -> None:
+        executor = self._executor
+        self._executor = None
+        if executor is not None:
+            executor.shutdown(wait=True, cancel_futures=False)
```
Teingi: One shutdown issue: Python's standard `ThreadPoolExecutor` workers are non-daemon, and `shutdown(wait=True)` drains the entire queue without a deadline. Hermes v0.20.4 already puts sync and queued prefetch work on its own daemon worker with a five-second drain. A slow PowerContext server can therefore keep the process alive after Hermes's own timeout has expired.

Please reuse the host worker for sync/prefetch where possible. If the provider still needs a worker for `on_memory_write`, make it daemonized, bound the queue/drain time, and log how many writes were dropped at shutdown.

Teingi (changes requested):
I retested this against Hermes v0.20.4 using both user and project plugin discovery. CLI registration, schema loading, a live remember/search flow, and model-visible recall all worked. The compatibility problems I saw on v0.15.1 no longer apply.

I'm requesting changes for the automatic write path. The inline comments cover raw transcript capture and deduplication, replacement/removal semantics, environment override handling, and shutdown behavior. The first two can leave sensitive or stale data in long-term memory, so I don't think we should merge with those open.

Please also document v0.20.4 as the minimum Hermes version. It would be good to wire `get_config_schema()` and `save_config()` into `hermes memory setup`; right now the provider is shown as "no setup needed" and the JSON file has to be configured manually.

The retrieval and CLI paths looked good in my testing.

integrations/hermes/plugins/powercontext/__init__.py
Alanxtl: Implemented.

- Added `capture_pre_compress`, disabled by default.
- Only user and assistant messages are eligible for capture.
- System/tool messages are excluded and common secret patterns are redacted.
- Added overlap-aware message fingerprinting to capture only new turns.
- Added stable idempotency keys for pre-compression sources.

integrations/hermes/plugins/powercontext/__init__.py
Alanxtl: Implemented.

- Added a stable Hermes memory item to PowerContext citation mapping.
- `add` stores the citation returned by PowerContext.
- `replace` retires the citation associated with `metadata.old_text` before writing the replacement.
- `remove` retires the citation associated with `metadata.old_text`.
- The mapping is persisted under `$HERMES_HOME/powercontext-memory-map.json`.
- Added an exact-text search fallback for mappings created before the local map was available.

integrations/hermes/plugins/powercontext/__init__.py
Alanxtl: Implemented.

`queue_prefetch()` now uses the same `_config_value(...)` lookup as `prefetch()`, so `POWERCONTEXT_HERMES_MAX_BYTES` correctly overrides the configured value.

Added a regression test covering the `16000` environment override.

integrations/hermes/plugins/powercontext/__init__.py
Alanxtl: Implemented.

- `sync_turn()` and `queue_prefetch()` now run directly on Hermes' existing daemon background worker.
- The provider keeps a separate daemon worker only for `on_memory_write()`.
- The memory-write queue is bounded to 128 items.
- Shutdown drains the queue for at most five seconds.
- Overflowed or abandoned writes are counted and logged.
- Added tests covering daemon worker behavior, queue bounds, and dropped-write reporting.

integrations/hermes/plugins/powercontext/__init__.py
```diff
@@ -0,0 +1,537 @@
...
+                response = self._client.prepare_context(
+                    self._scope_id,
+                    query[:8192],
+                    max_bytes=_as_int(
+                        self._config.get("max_bytes", _DEFAULT_MAX_BYTES),
+                        _DEFAULT_MAX_BYTES,
+                        minimum=512,
+                        maximum=32768,
+                    ),
+                )
```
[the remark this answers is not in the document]
Alanxtl: done

integrations/hermes/README.md
```diff
@@ -0,0 +1,120 @@
...
+  "timeout": 5,
+  "capture_turns": true,
+  "flush_on_session_end": true
+}
+```
+
+Environment variables override file values:
+
+| Variable | Purpose |
+| --- | --- |
```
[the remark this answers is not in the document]
Alanxtl: done

tests/integrations/test_hermes_provider.py
```diff
@@ -0,0 +1,158 @@
...
+import json
+import sys
+from pathlib import Path
+
+import pytest
+
+HERMES_ROOT = Path(__file__).parents[2] / "integrations" / "hermes"
+if str(HERMES_ROOT) not in sys.path:
+    sys.path.insert(0, str(HERMES_ROOT))
+
```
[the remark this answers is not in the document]
Alanxtl: done

integrations/hermes/plugins/powercontext/__init__.py
```diff
@@ -0,0 +1,537 @@
...
+        revision = int(args["revision"])
+    except (TypeError, ValueError) as error:
+        raise ValueError("revision must be an integer") from error  # noqa: TRY003
+    if revision < 1:
+        raise ValueError("revision must be positive")  # noqa: TRY003
+    return {
+        "memory_ref": {"family": str(args["family"]), "artifact_id": str(args["artifact_id"]), "revision": revision},
+        "entry_id": str(args["entry_id"]),
+        "entry_version_id": str(args["entry_version_id"]),
+    }
```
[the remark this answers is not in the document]
Alanxtl: done

integrations/hermes/plugins/powercontext/__init__.py
Teingi: Thanks, the single-entry path works now. There is still a revision problem once the scope contains more than one write. The map saves the full citation, including `memory_ref.revision`, but every memory mutation advances the artifact revision.

I reproduced this against a real PowerContext server on `a61d249`:

1. Add A, and the map stores revision 1.
2. Add B, and the current Memory head moves to revision 2.
3. Replace A, and `retire_memory_entry()` sends revision 1 and gets HTTP 409.

The worker catches that error, so A stays active and the replacement is never created.

Could we keep only the stable entry identity in the map, then refresh the citation from the current Memory head before replace/remove? Matching `entry_id` and `entry_version_id` against the current search or list result would give us the latest `memory_ref.revision` without confusing entries that have the same text. A regression test that advances revisions across add A, add B, and replace/remove A should catch this.

Alanxtl: Fixed 
- Memory map now stores only:
  - `entry_id`
  - `entry_version_id`
- Before `replace`/`remove`, the plugin searches the current Memory head and refreshes `memory_ref.revision`.
- Matching requires both stable IDs, preventing collisions for identical text.
- Added regression coverage for:
  - Add A → revision 1
  - Add B → revision 2
  - Replace A
  - Remove the replacement

integrations/hermes/plugins/powercontext/__init__.py
```diff
@@ -0,0 +1,1010 @@
...
+                text[:8192],
+                limit=50,
+                mode="fts",
+            )
+        except PowerContextError:
+            logger.debug("PowerContext memory citation lookup failed", exc_info=True)
+            return None
+        hits = response.get("hits", []) if isinstance(response, dict) else []
+        for hit in hits:
+            if not isinstance(hit, dict) or str(hit.get("text", "")).strip() != text.strip():
```
Teingi: There's still a mismatch with Hermes's `old_text` contract here. Hermes accepts a unique substring for replace/remove and forwards that substring unchanged. I tried replacing a full memory entry with `old_text="prefers uv"` against a real PowerContext server; the old entry stayed active and the replacement was never created because this comparison expects the whole text.

Could we resolve candidates with the same substring semantics Hermes uses, then intersect them with the identities in `_memory_map` and proceed only when exactly one entry matches? A regression test using partial `old_text` for both replace and remove would cover the host's actual behavior.

integrations/hermes/plugins/powercontext/__init__.py
```diff
@@ -0,0 +1,1010 @@
...
+            if normalized is not None and (identity is None or _entry_identity(normalized) == identity):
+                return normalized
+        return None
+
+    def _lookup_memory_citation(self, target: str, text: str) -> tuple[str, dict[str, Any] | None]:
+        key = self._memory_item_key(target, text)
+        stored = self._memory_map.get(key)
+        identity = _entry_identity(stored)
+        citation = self._find_memory_citation(text, identity=identity)
+        if citation is None and identity is None:
```
Teingi: The no-identity fallback is risky because it can retire any PowerContext entry with the same text, even if Hermes never created it. I reproduced this by writing a normal `preference` entry directly to PowerContext, starting with an empty map, and then issuing a Hermes remove for that text. The unrelated entry was retired.

I think this path should fail closed when no mapping exists. If backward compatibility is required, fetch the candidate with `get_memory_entry()` and verify its `kind` is `hermes-memory` or `hermes-user-memory`, and require a single match before retiring it. A same-text collision test would make this behavior explicit.

integrations/hermes/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Configuration
+
+The provider uses `http://127.0.0.1:8000` by default. Run the Hermes memory setup
+wizard to configure and activate it interactively; the wizard writes
+non-sensitive values to `$HERMES_HOME/powercontext/config.json`, stores the
+authorization header in Hermes' `.env` file, and sets `memory.provider`:
+
+```bash
+hermes memory setup powercontext
```
Teingi: This command doesn't run the provider schema wizard on Hermes v0.20.4. I tested it with a fresh `HERMES_HOME`; it only saved `memory.provider` and did not prompt for PowerContext settings or create `powercontext/config.json`.

For v0.20.4, the working path is `hermes memory setup` followed by selecting PowerContext. Could we document that command here and in the plugin README? The other option is to raise the minimum Hermes version once the direct `memory setup <provider>` path delegates to the generic wizard.

Teingi (commented):
A few follow-ups from the latest live test.

integrations/hermes/plugins/powercontext/__init__.py
Alanxtl: `old_text` now follows Hermes’ substring semantics. The provider filters search results against the stable identities in `_memory_map` and proceeds only when exactly one mapped entry matches. The current citation is used, so `memory_ref.revision` is refreshed before replace/remove.

integrations/hermes/plugins/powercontext/__init__.py
Alanxtl: Replace/remove now fail closed when no stable Hermes mapping exists, so the provider no longer retires an unrelated PowerContext entry based only on matching text.

integrations/hermes/README.md
Alanxtl: `hermes memory setup` is supported actually
I've updated the READMEs to guide users to use `hermes memory setup`

Teingi (approved):
LGTM

PR #1262, opened 2026-08-18
feat(tracing): cover context setup and scope lock waits

### LINKED ISSUE #1257: feat(tracing): cover context setup and scope lock waits
Just a thought, not a blocker: _context() (which awaits self._provider.get() and opens a DB session) and the wait on _lock(scope_id) both happen before the first stage span opens, so they don't show up in the trace. Since the goal here is locating latency in the read path, and test_same_scope_read_only_searches_do_not_serialize_reranking suggests lock contention is already on your radar, those might be interesting to cover eventually.

_Originally posted by @Ethan-Xingyue in https://github.com/oceanbase/powercontext/issues/1252#issuecomment-5314729951_

tracked under #1214.

Ethan-Xingyue:

Closes #1257.

# Rationale for this change

Application spans were not fully explained by their children: the provider resolution and the scope write lock wait
before the first stage span were invisible. A slow request gave no way to tell Memory recall latency apart from time
spent queued behind another write on the same scope.

Two new Runtime stage spans, reusing the existing `stage` unit:

```
powercontext search_memory              powercontext prepare_context
├── scope.context      <- new           ├── scope.context      <- new
└── memory.search                       ├── scope.lock         <- new
    ├── embeddings <model>               ├── memory.search
    └── memory.rerank                    ├── experience.search
                                         └── context.build
```

- **`scope.context`** wraps `await self._provider.get(...)` in `_context()` and nothing else, closing before the context
  is yielded. No attributes. Near zero with the built-in provider (`RelationalContexts.get()` is a cached in-memory
  lookup), kept because a third-party `PowerContextProvider` may do I/O in `get()` — this span is the only place that
  becomes visible.
- **`scope.lock`** measures the lock wait and ends at acquisition, so the critical section stays outside the span. Adds
  `_locked()`, since `_lock()` only returned the `asyncio.Lock` and the wait happened in each caller's `async with`.
  One bounded boolean, `powercontext.scope.lock.contended`.

All 16 `_lock(self.scope_id)` call sites now use `_locked(...)`; `_lock()` had one caller left and was inlined.
Read-only searches still take no lock, so they emit no `scope.lock`. Docs: one row per span in the stage span tables
(en + zh). RFC 0046 already lists the `Runtime stage` unit and is unchanged.

No — API, OpenAPI, CLI, persisted formats, and dependencies are untouched. Neither span records queries, content,
vectors, `scope_id`, artifact identifiers, or exception messages; both are registered in the test attribute allow-list.

# How was this change tested?

`make check` and `make test` pass (547 passed, 12 skipped).

- Extended `test_memory_read_stage_spans_are_bounded_and_nested`: both spans present with expected attributes and
  parents, no `scope.lock` under `search_memory`. The recall stages staying direct children of the application span is
  what proves neither setup span covers the operation body.
- Added `test_scope_lock_stage_span_reports_contention_and_closes_at_acquisition`: real contention via an externally
  held lock, and a failure inside the critical section still releases it. The conflicting write's wait span stays
  `success`, which is the direct evidence the span closes before the critical section.

Outcome semantics are already covered for every stage span in `tests/test_server_tracing.py`, so they are not re-tested
per span.

### DIFF
--- a/src/powercontext/builtin/runtime/application.py
+++ b/src/powercontext/builtin/runtime/application.py
@@ -277,7 +277,7 @@ class ScopedContextApplication:
         builder = PreparedContextBuilder()
         async with (
             self._runtime._context(self.scope_id, embedding_purpose=ModelUsagePurpose.MEMORY_RECALL) as context,
-            self._runtime._lock(self.scope_id),
+            self._runtime._locked(self.scope_id),
         ):
             with self._runtime._stage(
                 _MEMORY_SEARCH_STAGE,
@@ -384,7 +384,7 @@ class ScopedExperienceApplication:
         self.scope_id = validate_scope_id(scope_id)
 
     async def propose(self, request: ProposeExperienceRequest, /) -> ExperienceCandidate:
-        async with self._runtime._scoped_operation(self.scope_id), self._runtime._lock(self.scope_id):
+        async with self._runtime._scoped_operation(self.scope_id), self._runtime._locked(self.scope_id):
             service = self._runtime._review(self.scope_id)
             return await service.propose_experience(
                 request.proposal,
@@ -400,7 +400,7 @@ class ScopedExperienceApplication:
                 self.scope_id,
                 generation_purpose=ModelUsagePurpose.EXPERIENCE_GENERATION,
             ),
-            self._runtime._lock(self.scope_id),
+            self._runtime._locked(self.scope_id),
         ):
             return await self._runtime._generation(self.scope_id).experience(
                 sources=request.sources,
@@ -427,7 +427,7 @@ class ScopedExperienceApplication:
                 self.scope_id,
                 generation_purpose=ModelUsagePurpose.EXPERIENCE_GENERATION,
             ),
-            self._runtime._lock(self.scope_id),
+            self._runtime._locked(self.scope_id),
         ):
             return await incubator(self.scope_id, window_limit)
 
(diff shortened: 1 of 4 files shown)


src/powercontext/builtin/runtime/application.py
```diff
@@ -1198,10 +1198,22 @@ async def _context(
...
+    async def _locked(self, scope_id: str) -> AsyncIterator[None]:
+        """Serialize writes for one scope and trace only the wait, not the critical section."""
+
+        lock = self._locks.setdefault(validate_scope_id(scope_id), asyncio.Lock())
+        with self._stage("scope.lock", attributes={"powercontext.scope.lock.contended": lock.locked()}):
+            await lock.acquire()
+        try:
+            yield
+        finally:
+            lock.release()
```
Kairo-J: Is the lock cleanup mechanism installed too late? Once `lock.acquire()` executes successfully, Python will exit `self._stage()` before entering subsequent try/finally statements. If the injected `RuntimeTracing.stage().__exit__` throws an exception, `_locked().__aenter__` will abort while the lock is still in the acquired state. This could lead to all subsequent write operations to that scope waiting indefinitely, and runtime shutdown might be suspended while waiting for active operations.

Ethan-Xingyue: Confirmed — real leak. Fixed  the release is now armed inside the `with`, so a raising teardown can't leave the lock held.

Added `test_scope_lock_is_released_when_stage_teardown_fails`; verified it fails on the previous commit.

```python
acquired = False
try:
    with self._stage("scope.lock", attributes={...}):
        await lock.acquire()
        acquired = True
    yield
finally:
    if acquired:
        lock.release()
```

PsiACE (approved):
totally lgtm, thanks

PR #1267, opened 2026-08-18
fix(dsh): treat missing session cwd as unscoped

### LINKED ISSUE #1266: bug(dsh): treat session.header.cwd as optional for latest DeepSeek Harness
### Describe the bug

Latest DeepSeek Harness ( [dsh-0.1.0-rc.7](https://github.com/deepseek-ai/deepseek-harness/commit/99f6f02fecdb7dff40c3fbc9470f5907c29f74ca) `@deepseek-ai/dsh-session`) made `SessionHeader.cwd` optional. The PowerContext DSH plugin still treated it as required and, in tools, fell back to `process.cwd()`.

DSH evidence (`packages/core/session/src/types.ts`):

```ts
/** Absolute working directory the session was created in (if any). */
readonly cwd?: string
```

`sessions.create()` without `meta.cwd` leaves `header.cwd` unset. JSONL persistence omits the field rather than writing a path. Ordinary `dsh web` sessions that open a local workspace usually have cwd, so this is easy to miss.

### Steps to reproduce

1. Run a PowerContext Server and load `powercontext-dsh`.
2. Create a session with no workspace (`sessions.create()` without `meta`, or any host path that does not set `header.cwd`).
3. Send a user turn (recall), run `/pc search ...`, or call `pc_search`.

### Expected behavior

- Missing or blank cwd is treated as "no workspace", not as the Harness process directory.
- If `POWERCONTEXT_DSH_SCOPE_ID` / plugin `scopeId` is set, recall, `/pc`, and tools work without cwd.
- If neither a workspace nor `scopeId` is available:
  - recall fail-opens: skip prepare, log `context_prepare / skipped / missing_session_cwd`, do not throw;
  - `/pc` and `pc_*` tools return an explicit unscoped error, do not throw;
  - Source capture does not write a fabricated `cwd`.

### Actual behavior

On current `master` (plugin from #1239):

1. `spawnGit(undefined)` uses the Node process cwd (often the Harness install dir, not the project).
2. After git fails, `path.resolve(undefined)` throws `TypeError`.
3. Recall wraps that in `try/catch`, so the whole recall+capture turn is skipped silently.
4. `/pc` has no such catch and fails the command.
5. Tools use `header.cwd || process.cwd()`, which can remember/search under the Harness process directory.

### Why this has to change

This is a compatibility fix against current DSH, not an optional enhancement. Programmatic sessions, sessions without a workspace, and any host that omits `meta.cwd` are valid DSH sessions. Treating cwd as required either crashes `/pc` or silently scopes Memory to the wrong directory.

### Proposed adjustment

1. Normalize cwd with `sessionCwd()`: blank/whitespace is missing; never fall back to `process.cwd()`.
2. `deriveScopeId(cwd?: string)`: configured `scopeId` wins and never touches cwd; both missing → `undefined`.
3. Recall: skip + warn `missing_session_cwd`; capture metadata omits cwd when absent.
4. `/pc` and tools: return a clear unscoped error instead of throwing.
5. Align shims/types to `cwd?: string`.
6. Cover unit tests plus a live-server e2e that constructs a plugin runtime with `header.cwd === undefined`.

### Environment

- PowerContext: `master` after #1239
- DeepSeek Harness: current `SessionHeader.cwd?: string`
- Reproduced on Windows with a local plugin checkout

knqiufan:

Closes #1266.

# Rationale for this change

Latest DeepSeek Harness([dsh-0.1.0-rc.7](https://github.com/deepseek-ai/deepseek-harness/commit/99f6f02fecdb7dff40c3fbc9470f5907c29f74ca)) made `SessionHeader.cwd` optional. Sessions created without a workspace (`sessions.create()` with no `meta.cwd`) leave the field unset. The plugin treated cwd as required, so `/pc` could throw, recall was swallowed by fail-open, and tools could scope Memory to the Harness process directory.

- Treat missing or blank `session.header.cwd` as no workspace. Never fall back to `process.cwd()`.
- If `scopeId` / `POWERCONTEXT_DSH_SCOPE_ID` is set, recall, `/pc`, and tools work without cwd.
- If neither workspace nor `scopeId` is available: recall skips with `missing_session_cwd`; `/pc` and tools return an explicit unscoped error; capture metadata omits a fabricated cwd.
- Align plugin types/shims to `cwd?: string`.
- Add unit coverage and a live-server e2e that drives the plugin runtime with `header.cwd === undefined`.
- Update en/zh DSH setup docs to match.

Yes, for sessions without a workspace:

- Recall no longer throws or silently uses the Harness process directory.
- `/pc` and `pc_*` tools return a clear unscoped message unless `POWERCONTEXT_DSH_SCOPE_ID` is set.
- Docs now say to set `scopeId` when the session has no workspace cwd.

No Server API or storage format changes.

# How was this change tested?

- `npx vitest run --exclude tests/e2e/**` in `integrations/dsh/plugins/powercontext` (60 passed)
- `npx vitest run tests/e2e` (4 passed: existing HTTP call-through plus the new unscoped-session runtime e2e)

# AI usage statement

Assisted by an AI coding model for the compatibility change, tests, and this write-up.


integrations/dsh/plugins/powercontext/tests/e2e/unscoped-session.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+    wrapContent: (text) => ({ role: 'user', content: [{ type: 'text', text }] }),
+    log: runtime.log,
+  })
+}
+
+function processDirectoryScope(): string {
+  return `local:${createHash('sha256').update(resolve(process.cwd())).digest('hex')}`
+}
+
+describe('plugin runtime with header.cwd === undefined', () => {
```
Teingi, line 133: This second e2e file exposes a shell glob issue in the existing `test` script. On Linux, `vitest run --exclude tests/e2e/**` expands to both e2e paths. Vitest uses the first path as the exclude value and treats this file as a positional filter, so `pnpm test` runs this e2e file and skips all 60 unit tests. I reproduced this at `ade1124`; `vitest run --exclude=tests/e2e/**` runs all 12 unit test files. Can we keep the glob in one argument and make sure `make js-test` reports 60 passing tests?

Teingi (changes requested):
The cwd handling looks good. One test runner issue needs fixing before this merges.

integrations/dsh/plugins/powercontext/tests/e2e/unscoped-session.spec.ts
knqiufan: Thanks @Teingi catching this and for the clear reproduction. Fixed in `71b0e8b` by keeping the exclude glob in one argument (`--exclude=tests/e2e/**`). I verified `pnpm --dir integrations/dsh/plugins/powercontext test` (the command used by `make js-test`) now reports 12 test files / 60 passing tests. The separate E2E suite still reports 2 files / 4 passing tests. I really appreciate the careful review and suggested fix.

Teingi (approved):
LGTM

PR #1271, opened 2026-08-19
Track benchmark coverage expansion

### LINKED ISSUE #1263: Tracking Issue: expand benchmark coverage
## Goal

Track how PowerContext expands its benchmark coverage without creating a separate execution stack for every benchmark.

Existing E2E infrastructure, including Harbor, ACP, and containerized execution, should be the starting point where it fits. A benchmark may reuse upstream Harbor content, be represented as a Harbor task, or use another approach when its native structure calls for one. The integration should preserve the benchmark's original intent, inputs, and evaluation semantics.

Our current priority is to add more benchmark and evaluation workloads through the Harbor-backed E2E harness, using it as a shared basis for community evaluation.

## Principles

We should prefer benchmarks that are recognized or used by comparable memory, context, or agent projects.

Each proposal should explain why the benchmark is relevant to PowerContext, whether the comparison can be fair, and whether the result is clear and measurable. The exact execution and evaluation shape may differ between benchmarks.

Community contributions should focus on a small, runnable validation rather than a complete evaluation. Contributors are not expected to cover the full runtime or inference cost.

After a benchmark is accepted, maintainers may run the complete evaluation or a useful subset. Published results should state what was run and should not present a partial run as a complete benchmark result.

## Tracking

- [ ] Review candidate benchmarks and their use across comparable projects.
- [ ] Identify which parts of the existing E2E and Harbor infrastructure can be reused.
- [ ] Define a lightweight contribution path for adding a benchmark.
- [ ] Review fairness and quantitative evaluation for each proposal.
- [ ] Preserve the source benchmark's semantics when adapting it to PowerContext.
- [ ] Decide the appropriate full or partial evaluation scope after adoption.
- [ ] Add and evaluate accepted benchmarks.

This issue tracks benchmark adoption and shared infrastructure. Benchmark-specific design and execution details belong with each benchmark.

Migrating the existing full LoCoMo and SWE-bench Pro suites is out of scope for now. They should keep their current workflows so this work does not disrupt existing evaluation paths. This does not decide whether they should be consolidated under E2E later.

mercael91:
Closes #1263. Added a benchmark coverage tracking section and integration plan to the README. Preserved existing behavior and documentation.

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -1,7 +1,21 @@
-# PowerContext benchmarks
+# Benchmark Coverage
 
-Benchmarks in this directory exercise public PowerContext behavior against fixed datasets. They are kept outside
-`tests/` because they use real databases and model providers, create durable benchmark namespaces, and may take a
-long time or incur inference cost.
+PowerContext expands its benchmark coverage by integrating with existing E2E infrastructure, including Harbor, ACP, and containerized execution. Benchmarks may reuse upstream Harbor content, be represented as a Harbor task, or use another approach when their native structure calls for one. The integration preserves the benchmark's original intent, inputs, and evaluation semantics.
 
-- [`locomo/`](locomo/README.md): conversation-memory retrieval and end-to-end question-answer accuracy.
+## Current Status
+
+- [ ] Harbor integration
+- [ ] ACP integration
+- [ ] Containerized execution
+
+## Next Steps
+
+1. Define benchmark requirements and specifications.
+2. Implement benchmark integration with Harbor and ACP.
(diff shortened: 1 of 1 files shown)


mercael91:
Pushed an update: Add benchmark coverage tracking and integration plan. Ready for re-review when you have a moment.

PsiACE:
Thanks for the contribution. I've updated the issue description to clarify that changes to the existing benchmark suite are out of scope, so I'll close this PR.

mercael91:
Understood, thanks for reviewing! I'll keep this feedback in mind for future contributions.

PR #1274, opened 2026-08-19
Fix: normalizes cursor race into PowerContext's `GenerationConflictError`.

### LINKED ISSUE #1273: [Bug] Initial Source cursor creation is not concurrency-safe
## Summary

When two independent Runtime instances or processes share the same database and concurrently create the same Source cursor for the first time, the operation is not normalized to PowerContext's generation-conflict error.

## Reproduction

For the same `(scope_id, binding_name)`:

1. Runtime A calls `SourceCursorRepository.save(..., expected_generation=None)`.
2. Runtime B calls the same method at the same time.
3. Both initial `load()` calls can observe no row.
4. Both attempt the `INSERT`.
5. One succeeds; the other receives a raw database exception such as:

```
sqlite3.IntegrityError: UNIQUE constraint failed:
pc_source_cursors.scope_id, pc_source_cursors.binding_name
```

## Root cause

`SourceCursorRepository.save()` currently performs a non-atomic read-then-insert sequence when `expected_generation is None`. The in-process activation lock does not protect independent Runtime instances or processes.

## Expected behavior

The first writer should create generation 1. The losing writer should receive `GenerationConflictError`, consistent with the existing generation CAS behavior for updates, instead of a raw `IntegrityError`.

Alanxtl:

Closes #1273

# Rationale for this change

`SourceCursorRepository.save()` used a check-then-insert flow for initial cursor creation. When independent Runtime instances created the same cursor concurrently, both could observe that the row did not exist, causing one writer to receive a raw database `IntegrityError`.

This change normalizes that race into PowerContext's `GenerationConflictError`.

- Wrap initial cursor insertion in a savepoint.
- Catch concurrent unique-key conflicts.
- Reload the existing cursor using a locking read where supported.
- Raise `GenerationConflictError` for the losing writer.
- Add regression tests covering the concurrent initial-creation path.
- No schema or migration changes.

No

# How was this change tested?

- `uv run pytest tests/builtin/persistence -q`
  - `39 passed, 1 skipped`
- `uv run ruff check src/powercontext/builtin/persistence/cursors.py 
# AI usage statement

Developed with Codex

### DIFF
--- a/tests/builtin/persistence/test_cursors.py
+++ b/tests/builtin/persistence/test_cursors.py
@@ -49,76 +49,54 @@ def test_source_cursor_uses_generation_compare_and_swap() -> None:
     asyncio.run(scenario())
 
 
-def test_source_cursor_initial_creation_translates_a_racing_insert() -> None:
-    class StaleInitialReadRepository(SourceCursorRepository):
-        def __init__(self) -> None:
-            self._first_read = True
-
-        async def load(
-            self,
-            connection: AsyncConnection,
-            scope_id: str,
-            binding_name: str,
-            /,
-            *,
-            for_update: bool = False,
-        ) -> StoredSourceCursor | None:
-            if self._first_read and not for_update:
-                self._first_read = False
-                return None
-            return await super().load(connection, scope_id, binding_name, for_update=for_update)
-
+def test_source_cursor_initial_creation_is_safe_across_connections(tmp_path: Path) -> None:
     async def scenario() -> None:
-        async with repository_profile() as (profile, _):  # noqa: SIM117
-            async with profile.database.transaction() as connection:
-                await SourceCursorRepository().save(
-                    connection,
-                    "scope-a",
-                    "memory-source-window",
-                    SourceCursor(sequence=1),
-                    expected_generation=None,
-                )
+        barrier = asyncio.Barrier(2)
 
-                repository = StaleInitialReadRepository()
-                with pytest.raises(GenerationConflictError) as error:
-                    await repository.save(
+        class SynchronizedRepository(SourceCursorRepository):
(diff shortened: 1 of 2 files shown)


tests/builtin/persistence/test_cursors.py
```diff
@@ -42,3 +47,78 @@ async def scenario() -> None:
...
+            async with first_profile.database.transaction() as first_connection:
+                await repository.save(
+                    first_connection,
+                    "scope-a",
+                    "memory-source-window",
+                    SourceCursor(sequence=1),
+                    expected_generation=None,
+                )
+
+        async with SQLiteProfile.open(config, tables=SHARED_TABLES) as second_profile:  # noqa: SIM117
```
Teingi: I do not think the second test is exercising the race this PR is fixing. The first profile commits before the second one starts, so the second writer simply sees an existing cursor and takes the old conflict path.

The mocked stale read covers the `IntegrityError` translation, but it still runs on the same connection. It would not catch a problem with the savepoint or the locking read across transactions.

Could we make this a real two-connection test against the same database? Both writers should observe the missing cursor before inserting. We can then assert that one creates generation 1 and the other gets `GenerationConflictError(actual=1)`. That would protect the behavior we care about here, especially the current read required by MySQL/OceanBase.

Alanxtl: I replaced the previous tests with a real two-connection sharing the same database file.

Both writers perform a real initial `load()` and are synchronized after observing that the cursor is missing, then attempt the insert concurrently. The test asserts that one writer creates generation 1 and the other receives `GenerationConflictError(actual=1)`.

Teingi (approved):
LGTM

PR #1285, opened 2026-08-20
fix(server): preserve SQLite persistence for partial settings

PsiACE:

N/A

# Rationale for this change

Partial SQLite settings such as `POWERCONTEXT_SERVER_DATABASE_VEC1_EXTENSION` could replace the Server's file database default with `:memory:`, causing data loss after restart.

Partial SQLite settings now retain the default file URL. Explicit in-memory databases still work, with a startup warning. The Vec1 documentation now includes a persistent database URL.

SQLite data remains persistent when Vec1 is configured without an explicit database URL. No breaking API or storage format changes.

# How was this change tested?

- `make check`
- `make docs-test`
- `make test` (580 passed, 12 skipped)

# AI usage statement

OpenAI Codex assisted with investigation, implementation, tests, and documentation.


Teingi (approved):
LGTM

PR #1287, opened 2026-08-20
feat(hermes): add Hermes memory provider setup and diagnostics of powercontext cli

Alanxtl:

# Rationale for this change

add support for Hermes CLI:
```
powercontext setup hermes --source oceanbase/powercontext --ref v0.0.2
powercontext doctor hermes
```

including 

install plugin of powercontext hermes
cp to $HERMES_HOME/plugins/powercontext
check Hermes CLI installed or not
check PowerContext Hermes pligin exists or not
suppport --json

yes, powercontext cli support for hermes

# AI usage statement

coworked with codex

### DIFF
--- a/src/powercontext/cli/hermes.py
+++ b/src/powercontext/cli/hermes.py
@@ -16,9 +16,13 @@
 
 from __future__ import annotations
 
+import hashlib
 import os
+import re
 import shutil
 import subprocess
+import tempfile
+import uuid
 from dataclasses import dataclass
 from pathlib import Path
 from shutil import which
@@ -29,6 +33,8 @@ from powercontext.paths import powercontext_data_dir
 HERMES_HOME_ENV = "HERMES_HOME"
 HERMES_PLUGIN_RELATIVE = Path("integrations") / "hermes" / "plugins" / "powercontext"
 HERMES_PLUGIN_NAME = "powercontext"
+HERMES_MIN_VERSION = (0, 20, 4)
+_HERMES_VERSION_PATTERN = re.compile(r"Hermes Agent v?(\d+)\.(\d+)\.(\d+)")
 
 
 @dataclass(frozen=True, slots=True)
@@ -42,7 +48,8 @@ class HermesSetupResult:
 def install_hermes_plugin(*, source: str, ref: str) -> HermesSetupResult:
     """Install the PowerContext provider into Hermes' user plugin directory."""
 
-    hermes_executable()
+    executable = hermes_executable()
+    get_hermes_version(executable)
     data_dir = powercontext_data_dir()
     try:
         data_dir.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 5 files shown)


Alanxtl:
the ci fails due to timeout, maybe it is github's problem

src/powercontext/cli/hermes.py
```diff
@@ -0,0 +1,215 @@
...
+                detail="Hermes CLI is not installed or is not on PATH",
+            ),
+            "plugin": Diagnostic(
+                status=DiagnosticStatus.SKIPPED,
+                detail="not checked because Hermes CLI is unavailable",
+            ),
+        }
+
+    plugin = hermes_home() / "plugins" / HERMES_PLUGIN_NAME
+    installed = _is_hermes_plugin(plugin)
```
Teingi: This can report a broken install as healthy. I reproduced it with a provider missing client.py: powercontext doctor hermes returned exit 0 and ok: true, while hermes plugins doctor <path> --ci failed to import the provider. Could we validate the installed directory through the real Hermes plugin doctor, and verify the supported Hermes version, before reporting success?

src/powercontext/cli/hermes.py
```diff
@@ -0,0 +1,215 @@
...
+    return (path / "__init__.py").is_file() and (path / "plugin.yaml").is_file()
+
+
+def _usable_checkout(target: Path) -> bool:
+    return _is_hermes_plugin(target) or _is_hermes_plugin(target / HERMES_PLUGIN_RELATIVE)
+
+
+def _materialize_remote_checkout(source: str, ref: str) -> Path:
+    target = checkout_target(ref)
+    if _usable_checkout(target):
```
Teingi: This cache is keyed only by ref, so any usable checkout wins regardless of source, and mutable refs such as the default master are never refreshed. After one setup, a later run against an updated master—or a different repository with the same ref—silently installs the cached provider. Please include the normalized source/resolved commit in the cache identity or refresh mutable refs before returning.

src/powercontext/cli/hermes.py
```diff
@@ -0,0 +1,215 @@
...
+        data_dir.mkdir(parents=True, exist_ok=True)
+    except OSError as error:
+        raise SetupError.data_directory(data_dir, error) from error
+
+    plugin_dir = resolve_hermes_plugin_dir(source=source, ref=ref)
+    home = hermes_home()
+    target = home / "plugins" / HERMES_PLUGIN_NAME
+    try:
+        target.parent.mkdir(parents=True, exist_ok=True)
+        shutil.copytree(plugin_dir, target, dirs_exist_ok=True)
```
Teingi: dirs_exist_ok=True overlays the new release but never removes files that disappeared from it. I reproduced this by installing a provider with an extra module, then a new source without it; the removed module remained under $HERMES_HOME/plugins/powercontext and could still be discovered. Please stage the new directory and replace the old one atomically, or remove the old target before copying.

src/powercontext/cli/hermes.py
Alanxtl: done

src/powercontext/cli/hermes.py
Alanxtl: done

src/powercontext/cli/hermes.py
Alanxtl: done

Teingi (approved):
LGTM

PR #1289, opened 2026-08-20
build: derive package version from vcs

PsiACE:

N/A.

# Rationale for this change

Use Git tags as the package version source.

Adds `hatch-vcs` and uses the resolved version in release, Docker, and artifact builds.

No API changes.

# How was this change tested?

`uv lock --check`, pre-commit hooks, and the CLI version test.

# AI usage statement

OpenAI Codex (GPT-5) was used.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -18,7 +18,7 @@ WORKDIR /app
 COPY pyproject.toml uv.lock README.md LICENSE ./
 COPY src/ ./src/
 
-ARG POWERCONTEXT_VERSION=0.0.0
+ARG POWERCONTEXT_VERSION
 
 RUN --mount=type=cache,id=powercontext-uv,target=/root/.cache/uv \
     test -n "${POWERCONTEXT_VERSION}" \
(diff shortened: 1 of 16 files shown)


pyproject.toml
```diff
@@ -14,7 +14,7 @@
 
 [project]
 name = "powercontext"
-version = "0.0.2"
+dynamic = ["version"]
```
Teingi, line 17: Could we regenerate and commit `e2e/bub/uv.lock` as part of this change? It still records PowerContext as `0.0.2`. On this head, `uvx --from uv==0.10.12 uv lock --project e2e/bub --locked` exits because the lockfile needs to be updated, while the same command succeeds on the base commit. The E2E check stays green because `make harness-sync` rewrites the lockfile without checking the diff. Adding a locked check there would help keep the nested lockfile from drifting again.

docker/Dockerfile
```diff
@@ -18,8 +18,11 @@ WORKDIR /app
 COPY pyproject.toml uv.lock README.md LICENSE ./
 COPY src/ ./src/
 
+ARG POWERCONTEXT_VERSION=0.0.0
```
Teingi: This default makes the documented local Docker build install PowerContext as `0.0.0`: `.git` is excluded from the build context, and the README command does not pass `POWERCONTEXT_VERSION`. I reproduced the documented build and `importlib.metadata.version("powercontext")` returned `0.0.0` inside the image. Could we pass the resolved VCS version from the README and Compose paths as well, or make the build argument required? The GitHub Docker workflow already does the right thing.

Teingi (approved):
LGTM

PR #1292, opened 2026-08-20
feat: enable Dashboard by default without blocking Server startup

### LINKED ISSUE #1290: feat: enable Dashboard by default without blocking Server startup
## Summary

Enable the PowerContext Dashboard by default when starting the Server.

## Expected behavior

- Dashboard is enabled by default.
- Dashboard is available at `/`, for example:
  `http://127.0.0.1:8000/`
- The startup command prints the Dashboard URL.
- Dashboard can start without configured scopes and should show an empty state.
- Dashboard does not require Server bearer authentication when authentication is disabled.
- Dashboard can be explicitly disabled with:

  ```bash
  POWERCONTEXT_SERVER_DASHBOARD_ENABLED=false
  ```

## Failure handling

If Dashboard initialization fails because of missing templates, static assets, or another startup error:

- print the direct failure reason;
- log a warning;
- continue starting the PowerContext Server;
- keep the HTTP API, MCP, and health endpoints available.

## Notes

Dashboard, HTTP API, and MCP share the same Server listener and port. Automatic port fallback is not part of this change.

Zxf-xufeng:
Closes #1290

# Rationale for this change

PowerContext Server did not enable Dashboard by default, and Dashboard initialization could make startup failures affect core Server availability. Users also lacked a clear startup URL.

- Enable Dashboard by default when Server starts.
- Print the Dashboard URL (/) on startup.
- Allow Dashboard to run without configured scopes and without bearer auth when Server auth is disabled.
- Make Dashboard initialization fail open: print/log the direct reason and continue HTTP API, MCP, and health endpoints.
- Add an explicit environment switch: `POWERCONTEXT_SERVER_DASHBOARD_ENABLED=false`.
- Update English and Chinese documentation and regression tests.
- Dashboard, HTTP API, and MCP share the same listener and port; automatic port fallback is out of scope.

Yes. Dashboard is available by default at `http://127.0.0.1:8000/` (assuming the default host and port), and startup prints the URL. Dashboard can be disabled with the environment variable above. A Dashboard startup failure no longer prevents the core Server from starting. No breaking API changes.

# How was this change tested?

- `env POWERCONTEXT_SERVER_AUTH_ENABLED=false POWERCONTEXT_SERVER_DASHBOARD_ENABLED=true .venv/bin/pytest tests/test_dashboard.py tests/test_server.py tests/test_cli.py -q` (50 passed)
- `.venv/bin/ty check src/powercontext/server src/powercontext/cli` passed
- `.venv/bin/zensical build -s` passed
- Ruff lint and format checks passed.
- JavaScript syntax checks passed.
- `git diff --check` passed.

# AI usage statement

Built with OpenAI Codex (GPT-5-based coding agent), used for implementation, tests, and documentation updates.

### DIFF
--- a/docs/en/docs/how-to/install-and-run.md
+++ b/docs/en/docs/how-to/install-and-run.md
@@ -37,8 +37,14 @@ With no environment variables, the Server:
 
 - binds to `127.0.0.1:8000`;
 - enables Streamable HTTP MCP at `/mcp`;
+- enables the Dashboard at `/`; when no scopes are configured, the page shows an explicit empty state;
 - creates a persistent SQLite database in the operating system's user data directory;
 - supports explicit Memory operations without an inference provider.
+
+After startup, the terminal prints the Dashboard URL, such as `http://127.0.0.1:8000/`. The Dashboard shares the
+Server listener and port with the HTTP API and MCP. If Dashboard initialization fails, the Server logs a warning with
+the direct cause and continues serving the other interfaces. Set `POWERCONTEXT_SERVER_DASHBOARD_ENABLED=false` to
+disable the Dashboard explicitly.
 
 `Ctrl-C` performs a clean shutdown. Restarting the command reopens the same database.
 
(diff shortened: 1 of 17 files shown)


Teingi (approved):
LGTM

PR #1293, opened 2026-08-20
feat(persistence): add embedded seekDB support

Evenss:

None

# Rationale for this change

PowerContext currently supports SQLite and externally managed OceanBase deployments. This PR adds an optional embedded seekDB backend for users who need OceanBase-compatible persistence without running a separate database server.

The integration uses `pylibseekdb` only to manage the embedded database lifecycle, while reusing the existing OceanBase-compatible persistence, full-text search, and vector search paths.

- Add an optional `powercontext[seekdb]` extra with `pylibseekdb`.
- Introduce `SeekDBConfig` and `SeekDBProfile`.
- Start and stop the embedded seekDB instance together with the PowerContext runtime.
- Connect to the local seekDB instance through its asynchronous MySQL Unix socket.
- Reuse the existing OceanBase-compatible full-text and vector index implementations.
- Add `seekdb` to the runtime database discriminator and persistence composition.
- Disable session autocommit to preserve SQLAlchemy transaction behavior.
- Add a seekDB-specific SQLAlchemy close path to avoid socket reset errors during graceful shutdown.
- Add cancellation-safe startup cleanup and close the SQLAlchemy connection pool before stopping the embedded instance.
- Support `POWERCONTEXT_SERVER_DATABASE_KIND=seekdb`.
- Use the Server data directory's `seekdb` subdirectory when `DATABASE_PATH` is omitted or blank.
- Support custom seekDB paths and database names.
- Extend the real-experience E2E harness to discover, count, and purge seekDB-backed scopes.
- Add Server settings coverage for default, blank, and custom seekDB paths.
- Update `.env.example` with an opt-in embedded seekDB configuration example.
- Update `uv.lock` with the optional seekDB dependency.

Yes. This PR adds a new opt-in persistence backend.

Install PowerContext with embedded seekDB support:

```bash
pip install "powercontext[server,seekdb]"
```

Configure the Server:

```env

# Optional. When omitted, the Server uses the seekdb subdirectory
# under its data directory, such as $POWERCONTEXT_HOME/seekdb.
POWERCONTEXT_SERVER_DATABASE_PATH=.powercontext/seekdb
```

The seekDB database name defaults to `test`.

This is not a breaking change:

- SQLite remains the default database.
- Existing SQLite and OceanBase behavior is unchanged.
- `pylibseekdb` is not installed by the default Server extra.
- The native seekDB binding is loaded only when `kind=seekdb` is selected.
- Settings parsing does not create the seekDB directory or start an embedded instance.

Embedded seekDB currently requires a supported Linux or macOS `pylibseekdb` wheel. Windows users can continue using SQLite or OceanBase, but embedded seekDB is not currently available on Windows.

# How was this change tested?

- `uv lock --check`: passed.
- Ruff checks on the changed Python sources and focused tests: passed.
- Type checking: passed.
- `git diff --check`: passed.
- Targeted Server database settings tests: 7 passed.
- Manually started PowerContext with an embedded seekDB data directory.
- Verified `/health/live` returned a healthy status.
- Verified `/health/ready` reported both the runtime and database as ready.
- Verified database schema initialization through the existing persistence path.
- Verified graceful `Ctrl+C` shutdown exited with code 0.
- Verified shutdown produced no `ConnectionResetError`, SQLAlchemy close warning, or traceback.

# AI usage statement

OpenAI Codex (GPT-5.6) was used to assist with implementation, debugging, code review, test execution, and drafting this PR.


src/powercontext/builtin/persistence/seekdb/profile.py
```diff
@@ -0,0 +1,178 @@
...
+        try:
+            engine = _create_engine(config, instance.connection_options())
+            database = AsyncDatabase.own(engine)
+            profile = cls(database=database, tables=tables)
+            try:
+                async with database.transaction() as connection:
+                    await create_tables(connection, tables)
+                yield profile
+            finally:
+                await database.close()
```
Teingi: Could we make shutdown cancellation-safe here? If `database.close()` is cancelled while it is waiting for an active transaction, its close state is reset, but this outer `finally` still calls `instance.close()`. I reproduced this with `pylibseekdb==1.3.0.post4`: hold a transaction open, start `__aexit__`, and cancel it once the database enters `_closing`. The active connection then fails with `OperationalError 2013 (Lost connection to MySQL server)`. This can abort an in-flight write during shutdown. Please shield and await the database close before closing the seekDB instance, then re-raise the cancellation. A regression test with a held transaction would be useful.

pyproject.toml
```diff
@@ -44,6 +44,10 @@ builtin = [
     "pyobvector>=0.2.28,<0.3",
     "sqlalchemy[asyncio]>=2,<3",
 ]
+seekdb = [
```
Teingi, line 49: This extra also changes the root package metadata recorded in `e2e/bub/uv.lock`, but that lockfile was not regenerated. On this head, `uvx --from uv==0.10.12 uv lock --project e2e/bub --locked` exits with `The lockfile needs to be updated`; the same command passes on the base commit. Please regenerate and commit `e2e/bub/uv.lock`. It may also be worth making the E2E validation run the locked check, since the current sync step can update the file silently.

src/powercontext/builtin/persistence/seekdb/profile.py
```diff
@@ -0,0 +1,178 @@
...
+    # aiomysql otherwise mistakes the default-on session for an explicit
+    # transaction and rollback becomes ineffective.
+    options["init_command"] = "SET autocommit = 0"
+    url = URL.create(
+        _DIALECT_DRIVER,
+        username=username,
+        password=None if password_value is None else str(password_value),
+        host=host,
+        port=None if port_value is None else int(cast(int | str, port_value)),
+        database=config.database,
```
Teingi, line 176: How is a non-default database supposed to be created? With a fresh embedded path and `database="custom"`, this connection fails with `OperationalError 1049: Unknown database 'custom'`. The new settings test only checks parsing, so it does not catch this. Please either create the requested database before opening the main engine, or document and restrict this setting to an existing database and cover that behavior with a real open test.

.env.example
```diff
@@ -16,6 +16,11 @@ POWERCONTEXT_SERVER_DATABASE_VEC1_EXTENSION=.powercontext/vec1.so
 # POWERCONTEXT_SERVER_DATABASE_KIND=oceanbase
 # POWERCONTEXT_SERVER_DATABASE_URL=mysql+aoceanbase://user:password@host:2881/powercontext?charset=utf8mb4
 
+# To use embedded seekDB instead, install powercontext[server,seekdb] and replace the SQLite database values.
```
Teingi, line 18: Could we add a short section to the user-facing docs showing the complete setup for embedded seekDB? This is an optional backend, and right now the only guidance is this `.env.example` comment. Please include the install command, supported platforms, the required environment variables, and a start/verification example. Without that, users can switch `DATABASE_KIND` and only discover at runtime that the seekDB extra is missing.

Evenss:
Thanks for the review @Teingi  . I have addressed all of the feedback:

- Made seekDB shutdown cancellation-safe. If cancellation occurs during shutdown, PowerContext now waits for active transactions to finish and for the engine to be disposed before closing the seekDB instance, then re-raises the cancellation.
- Added a regression test that cancels shutdown while an active transaction is held.
- Regenerated `e2e/bub/uv.lock` with uv 0.10.12 and the official PyPI index. The locked validation now passes.
- Restricted embedded seekDB to its built-in `test` database. The storage path remains configurable, while other database names are rejected during configuration validation.
- Added embedded seekDB setup sections to the English and Chinese user documentation, covering installation, supported platforms, environment variables, startup, and verification.

Validation completed:

- seekDB profile tests: 6 passed
- Server seekDB configuration tests: 5 passed
- Ruff, ty, prek, and `git diff --check`: passed
- Strict documentation build: passed
- Real temporary seekDB startup and connection smoke test: `SELECT DATABASE()` returned `test`

I kept the GitHub workflows unchanged to keep this PR focused on the minimal seekDB integration.

src/powercontext/builtin/persistence/seekdb/profile.py
```diff
@@ -0,0 +1,175 @@
...
+            try:
+                async with database.transaction() as connection:
+                    await create_tables(connection, tables)
+                yield profile
+            finally:
+                close_task = asyncio.create_task(database.close())
+                try:
+                    await asyncio.shield(close_task)
+                except asyncio.CancelledError:
+                    await close_task
```
Teingi: Could we keep the cleanup task shielded after the first cancellation? The initial `asyncio.shield(close_task)` protects `database.close()`, but the `except` branch awaits `close_task` directly. A second cancellation therefore cancels the close, and the outer `finally` still calls `instance.close()` while an active transaction is running.

I reproduced this on the current head with `pylibseekdb==1.3.0.post4`: hold an insert transaction, start `__aexit__`, then cancel the shutdown task twice. The transaction fails with `OperationalError 2013 (Connection reset by peer)`, and the inserted row is missing after reopening the database. With one cancellation, the row is preserved.

Please wait for the cleanup task through repeated shielded awaits, defer and re-raise cancellation only after cleanup completes, and add a regression test covering multiple cancellations. `_open_instance()` has the same unshielded follow-up await.

Evenss: Thanks for catching this. This has been fixed in the latest update.

I added a shared `_finish_task()` helper that repeatedly awaits the cleanup task through `asyncio.shield()`. Both database shutdown and `_open_instance()` are now protected from subsequent cancellations, allowing cleanup to complete before the original cancellation is re-raised.

I also updated the regression tests to cancel twice and verify that:

- the active transaction completes before the engine and seekDB instance are closed;
- an instance opened during cancelled startup is properly closed;
- the first cancellation is propagated after cleanup completes.

All 6 seekDB profile tests pass.

Teingi (approved):
LGTM

PR #1295, opened 2026-08-20
feat(integrations): support pydantic-ai integrate

### LINKED ISSUE #1269: feat: Pydantic AI framework integration
### Feature description

Add `powercontext-pydantic-ai`, an adapter distributed from `integrations/`, connecting a Pydantic AI agent to a running PowerContext Server through the public Python client.

The adapter provides two components:

1. `PowerContextToolset`, an `AbstractToolset` implementation exposing three Memory tools and their usage instructions.
2. `PowerContext`, an `AbstractCapability` implementation that composes the toolset and adds bounded context preparation before each model request, optional trajectory capture, and Memory flush.

Minimal usage:

```python
from pydantic_ai import Agent
from powercontext_pydantic_ai import PowerContext

agent = Agent("openai:gpt-5", capabilities=[PowerContext()])

async with agent:
    result = await agent.run("Why was the Redis plan dropped?")
```

Users who already operate their own retrieval stage can adopt the toolset alone, via `Agent(toolsets=[PowerContextToolset()])`, without the adapter modifying their message history.

In scope: Memory read and write, bounded context preparation, optional trajectory capture.

Out of scope for the initial release: Handoff, Artifact Candidate review, and Experience or Skill generation. These target coding-agent workflows and are deferred until a Pydantic AI use case is demonstrated.

Tracked under #1213.

### Problem

PowerContext supports four hosts today: Codex, Claude Code, DeepSeek Harness, and Bub. Each is a command-line coding agent, and each integration is implemented against a host-specific hook or plugin protocol. No integration exists for a general-purpose agent framework.

A developer building an agent with Pydantic AI has two options, neither of which is sufficient.

1. Call the HTTP API or the Python client directly from application code. This requires reimplementing scope derivation, context injection, injection idempotency, trajectory capture, flush scheduling, and failure isolation in every project.
2. Connect to the Server MCP endpoint. This exposes Memory reads and writes, but not `prepare_context`. `_MCP_OPERATION_IDS` in `src/powercontext/server/mcp.py` is an explicit allow list of 18 operation ids, and `_select_mcp_type` returns `MCPType.EXCLUDE` for every route outside it. `flush_memory` is excluded on the same basis.

The consequence is that the two behaviors distinguishing PowerContext from a generic memory tool, namely bounded context preparation before each model request and durable capture of the agent trajectory, are unavailable to Pydantic AI users without per-project glue code.

### Proposed solution

The design maps PowerContext behavior onto Pydantic AI V2 extension points that are public API. All references were verified against tag `v2.31.1`.

#### Component structure

| Component | Base class | Responsibility |
| --- | --- | --- |
| `PowerContextToolset` | `AbstractToolset` | Tool definitions, tool dispatch, usage instructions, client lifecycle |
| `PowerContext` | `AbstractCapability` | Composes the toolset, performs context preparation, capture, and flush |

Separating the two allows the toolset to be adopted independently, and keeps tool usage instructions attached to the tools rather than to the capability. `AbstractToolset.get_instructions` returns text that is merged into agent instructions, so the instructions remain present when only the toolset is registered.

#### Interface mapping

| Tool | HTTP operation |
| --- | --- |
| `powercontext_search` | `POST /v1/memory/search` |
| `powercontext_remember` | `POST /v1/memory/remember` |
| `powercontext_context` | `POST /v1/context/prepare` |

Tool names use underscores rather than the dotted names used by the Bub plugin. `pydantic_ai._utils.TOOL_NAME_SANITIZER` matches `[^a-zA-Z0-9_-]` and is documented as covering characters not accepted by most providers, but it is applied only by `models/bedrock.py`. Dotted names therefore reach other providers unmodified. Name collisions with other toolsets are resolved by the caller using `AbstractToolset.prefixed`, so the adapter does not need a prefix parameter.

#### Installation and authentication

```bash
uv pip install powercontext-pydantic-ai
powercontext server run
```

Configuration is read through pydantic-settings with the prefix `POWERCONTEXT_PYDANTIC_AI_`.

| Variable | Default | Purpose |
| --- | --- | --- |
| `BASE_URL` | `http://127.0.0.1:8000` | Server URL |
| `TOKEN` | unset | Bearer token forwarded to `PowerContextClient` |
| `SCOPE_ID` | derived | Durable scope shared across sessions |
| `TIMEOUT` | `10` | Client timeout in seconds |
| `MAX_BYTES` | `8000` | Prepared context size limit |
| `CAPTURE_EVENTS` | `false` | Enable trajectory capture |
| `CAPTURE_CHECKPOINT_EVERY` | `5` | Flush interval measured in captured events |
| `CAPTURE_MAX_BYTES` | `8192` | Per-event capture limit |

`TOKEN` carries a bare token rather than a complete `Authorization` header value. This differs from the `POWERCONTEXT_*_AUTHORIZATION` convention used by the Codex, Claude Code, and DeepSeek Harness plugins, which compose the header themselves. `PowerContextClient.__init__` accepts a bare token and composes `Authorization: Bearer {token}` internally, so an adapter built on the client should not compose and then decompose the header. The difference must be stated in the setup documentation.

The adapter installs no files outside the Python environment, requires no elevated permissions, and does not modify user configuration. The fourth constraint in #1213, covering installer disclosure, therefore does not apply.

#### Scope mapping

Scope resolution reuses `derive_scope_id` from `integrations/codex/plugins/powercontext/scripts/project_scope.py`, in the order: explicit configuration, then `git:<host>/<path>` normalized from `remote.origin.url` with credentials and the `.git` suffix removed, then `local:<sha256>` of the resolved project root. Values are bounded by `MAX_SCOPE_ID_LENGTH`, defined as 256 in `src/powercontext/limits.py`.

One deviation from the Codex plugin is required. A Pydantic AI agent is frequently deployed as a long-running service rather than executed inside a repository checkout, so explicit configuration is a primary path rather than an override. When `SCOPE_ID` is set, Git detection is skipped and no warning is emitted.

To support multi-tenant deployments, `PowerContext(scope_id=...)` accepts either a fixed string or a `Callable[[RunContext], str]` resolving from `ctx.deps`. Pydantic AI applies the same value-or-callable pattern to `AgentModelSettings`, so the shape is consistent with existing framework conventions.

#### Recall and capture behavior

| Behavior | Extension point |
| --- | --- |
| Context preparation | `AbstractCapability.before_model_request` |
| Model-step capture | `AbstractCapability.after_model_request` |
| Tool-result capture | `AbstractCapability.after_tool_execute` |
| Terminal flush | `AbstractCapability.after_run` |

`before_model_request` receives a `ModelRequestContext` whose `messages` field is a mutable `list[ModelMessage]`. The adapter extracts query text from the most recent `UserPromptPart`, calls `prepare_context`, and prepends the result as a `ModelRequest` containing a `SystemPromptPart`. When `PreparedContext.status` is `empty`, no message is added.

Two properties are required of this behavior.

Injection must be idempotent within a run. `before_model_request` fires once per model request, including after each tool round. The injected part carries a fixed prefix, and injection is skipped when that prefix is already present, matching the marker-scanning approach used by the Bub plugin.

Injected content must be labelled as untrusted historical evidence, following the wording already used by the Bub plugin. Memory content originates from prior model output and user input. Presenting it as authoritative system instruction would extend the prompt injection surface to historical data. RFC 0016 states the same position in its trust boundary section, and this proposal does not weaken it.

Capture is disabled by default, consistent with the Bub plugin, because it transmits tool output to the Server and constitutes a data flow requiring explicit consent. When enabled, captured events are written as Content Sources, and `flush_memory` is invoked every `CAPTURE_CHECKPOINT_EVERY` events so that later model steps within the same run can recall earlier findings. The credential redaction behavior implemented in the Bub plugin must be preserved rather than reimplemented; see open question 1.

#### Failure and recovery behavior

Server unavailability must not block an agent run. `TransportError`, `ServerResponseError`, and `InvalidResponseError`, all defined in `src/powercontext/client/errors.py` as subclasses of `ClientError`, are handled as follows.

| Operation | Behavior on failure |
| --- | --- |
| Context preparation | Skip injection and continue the run |
| Capture | Discard the event without retry |
| Flush | Skip. `flush_memory` advances a cursor, so pending Sources are processed by the next successful flush |
| Tool call | Raise `ModelRetry` with the failure reason |

Tool calls are treated differently because a model that invoked `powercontext_search` and received an empty result will conclude that no relevant memory exists. Returning the failure through `ModelRetry` allows the model to retry or select a different strategy.

`ServerResponseError` carries `status_code`. Responses of 401 and 403 indicate configuration errors rather than transient faults and must produce a distinct log entry on first occurrence, identifying a missing or invalid token.

#### Tests

The adapter does not modify `openapi/powercontext.yaml`, so `make contract-test` is unaffected.

Behavior tests under `tests/pydantic_ai_adapter/` drive an agent with `TestModel` and `FunctionModel` against `TestClient(create_server_app(...))`, following the pattern in `tests/e2e/test_runtime_server.py`. The assertions cover externally observable behavior:

- context is injected exactly once per run, including runs containing multiple tool rounds;
- injected content carries the untrusted-evidence label;
- a run completes successfully while the Server is unreachable;
- no `POST /v1/sources/content` request is issued while capture is disabled.

An end-to-end test at `tests/e2e/test_pydantic_ai_chain.py` starts a Server and exercises the capture, flush, and search sequence, matching `tests/e2e/test_codex_service_chain.py` and `tests/e2e/test_dsh_http_chain.py`.

Per `AGENTS.md`, the tests assert observable behavior only and do not freeze call counts, call ordering, or internal structure.

#### Documentation

Setup and troubleshooting documentation is added under `docs/en/docs/how-to/` with a Chinese counterpart under `docs/zh/`, matching the existing per-host configuration guides. `docs/en/docs/reference/interfaces.md` gains a row for the adapter. The documentation must state the token format difference described above, and must present the MCP endpoint as a supported lower-capability alternative with its limitations.

#### Distribution

A separate distribution under `integrations/pydantic-ai/`, following the precedent set by `integrations/bub`:

- distribution name `powercontext-pydantic-ai`, module `powercontext_pydantic_ai`;
- dependencies `powercontext[client]`, `pydantic-ai-slim>=2.29,<3`, `pydantic-settings`;
- build backend hatchling;
- `requires-python = ">=3.11"`, matching the root package, since pydantic-ai-slim requires only 3.10;
- the directory added to the `[tool.ty] exclude` list in the root `pyproject.toml`, currently `["e2e/bub", "integrations/bub"]`.

No entry-point registration is provided. Bub discovers plugins through `[project.entry-points."bub"]`; Pydantic AI has no equivalent discovery mechanism, and implicit activation would be inappropriate for a library that users compose explicitly.

### Use the MCP endpoint without an adapter

Pydantic AI V2 can connect to the Server with no additional package:

```python
from pydantic_ai.capabilities import MCP

agent = Agent("openai:gpt-5", capabilities=[
    MCP(url="http://127.0.0.1:8000/mcp", authorization_token="..."),
])
```

Assessment: sufficient for users who only require model-initiated Memory lookup, and it should be documented as such. It does not replace the adapter for three reasons. `prepare_context` is not projected as an MCP tool, so automatic context preparation is unavailable. The 18 operations in the allow list include Handoff and Artifact Candidate review, which are not relevant to a general-purpose agent; `MCP(allowed_tools=[...])` can filter them, but requires the user to know which to retain. Capture is initiated by the host rather than by the model, so it has no representation in a model-invoked protocol.

Decision: document both paths. MCP is presented as the zero-dependency option, the adapter as the complete one.

### Implement context preparation with the `ProcessHistory` capability

Assessment: `ProcessHistory` is a wrapper that rewrites `request_context.messages` inside `before_model_request`. Since the design already requires a capability, using the underlying hook removes one level of indirection and retains access to `model_settings` and `model_request_parameters` for future context-budget work.

Decision: use `before_model_request` directly.

### Reuse the dotted tool names defined by the Bub plugin

Assessment: `powercontext.search` and its siblings would reach providers unmodified, because the tool name sanitizer is applied only by the Bedrock model implementation. Provider acceptance of dotted names is not guaranteed and is outside PowerContext's control.

Decision: use underscore-separated names.

### Register the adapter through a packaging entry point

Assessment: this would match the Bub integration, but Pydantic AI provides no plugin discovery mechanism, and implicit registration conflicts with the explicit composition model the framework uses.

Decision: require explicit registration by the user.

### Version baseline

Verified against pydantic-ai tag `v2.31.1`. V2 relocated the extension parameters that previously appeared on `Agent.__init__` onto `capabilities`, and replaced the `MCPServerStdio`, `MCPServerSSE`, and `MCPServerStreamableHTTP` classes with `MCPToolset`. This design applies to V2 only. The proposed floor is `pydantic-ai-slim>=2.29,<3`; see open question 3.

### Relationship to RFC 0016

`docs/en/rfcs/0016_pydantic_ai_inference_integration.md` covers PowerContext using Pydantic AI internally as its inference layer, implemented under `src/powercontext/builtin/inference/` and declared in the `builtin` extra as `pydantic-ai-slim[anthropic,openai]>=2.27.1,<3`. This proposal addresses the opposite direction, in which user-authored Pydantic AI agents consume PowerContext. The two share an upstream dependency but no code path.

### Constraint compliance

| Constraint from #1213 | How the design satisfies it |
| --- | --- |
| Reuse existing PowerContext interfaces | All server interaction is through `PowerContextClient` and the published HTTP operations |
| Do not duplicate Runtime or Memory behavior | The adapter holds no memory state; entry identity, versioning, and search ranking remain server-side |
| Do not depend on unsupported upstream extension mechanisms | `AbstractToolset` and `AbstractCapability` are public API in V2 and are documented as extension points |
| Report installer effects before modifying the environment | Not applicable; the adapter is a library and modifies nothing outside the Python environment |

### Open questions

1. The credential redaction helpers in the Bub plugin, `_sanitize`, `_redact_known_secrets`, and `_codex_auth_secrets`, are private to `powercontext_bub`. A decision on extracting them into a shared location is required before capture is implemented, since duplication would fork the two implementations immediately.
2. Scope derivation has already diverged. The Codex plugin derives from the Git remote, while the Bub plugin uses `bub:<sha256[:20]>` of the workspace path. Extraction to a shared location is worth considering, and would need to decide where the shared code lives.
3. Whether to lower the dependency floor to `2.27.1` for consistency with the root `builtin` extra. The APIs used by this design were verified on 2.29 and 2.31.1; verifying 2.27.1 is a prerequisite for lowering it.
4. Behavior of capability hooks under `durable_exec` (Temporal, DBOS, Prefect) has not been verified. `AbstractToolset.id` must be stable for durable execution and is straightforward to provide, but the first release should declare durable execution unverified rather than implying support.

### Unrelated defect observed during design research

`integrations/bub` exposes no authentication setting. `PowerContextSettings` has no token field, and all six `PowerContextClient(...)` call sites in `plugin.py` and `tools.py` pass only `base_url` and `timeout`. The Bub plugin therefore cannot reach a Server with authentication enabled. This is unrelated to the present proposal and warrants a separate issue.

thunguo:

Closes #1269 .

# Rationale for this change

Pydantic AI users currently lack a native PowerContext integration that follows Pydantic AI’s toolset and capability lifecycles.

This PR adds an independent adapter that provides Memory tools, automatic context preparation, optional event capture, and checkpoint flushing through the public PowerContext Python Client.

- Add the independently distributable `powercontext-pydantic-ai` package.
- Export the following public APIs:
  - `PowerContextSettings`
  - `PowerContextToolset`
  - `PowerContext`
- Add per-run `PowerContextClient` lifecycle management and the following tools:
  - `powercontext_search`
  - `powercontext_remember`
  - `powercontext_context`
- Add automatic context preparation from the latest user prompt.
  - Context is injected once per run as untrusted historical evidence.
  - Reusing history in a new run prepares context again.
  - Empty context is not injected.
- Add deterministic scope resolution using:
  1. Constructor value or callback
  2. `POWERCONTEXT_PYDANTIC_AI_SCOPE_ID`
  3. Normalized Git origin
  4. A deterministic local hash
- Convert tool-side client errors into Pydantic AI `ModelRetry` errors.
- Keep context recall, capture, and flush operations fail-open when the PowerContext server is unavailable.
- Add optional event capture for user prompts, visible model output, tool calls, arguments, and results.
  - Thinking content is excluded.
  - Capture is disabled by default.
  - Captured JSON is bounded and credentials are redacted.
  - Source IDs are deterministic and idempotent.
  - Checkpoint and final-run flushing are supported.
- Extract shared capture sanitization and bounded JSON rendering into `powercontext.client.capture`, and reuse it in the Bub integration.
- Add English and Chinese configuration documentation and update the documentation navigation and interface reference.
- Add adapter unit tests and an end-to-end test using `FunctionModel`, the real server lifecycle, and SQLite.
- Update root type-checking and test import paths for the new integration.

This PR does not modify the server API, OpenAPI schema, or MCP allow-list.

Yes. Users can install the new package and attach PowerContext directly to a Pydantic AI agent.

The integration introduces new `POWERCONTEXT_PYDANTIC_AI_*` configuration variables. The token setting expects a raw token; users should not include the `Bearer` prefix.

Capture is opt-in and disabled by default because enabling it may persist prompts, visible model output, tool calls, and tool results. Sensitive values are redacted before capture, but users should still review the associated data-retention risk.

There are no breaking changes to existing public APIs or persisted formats, and no migration is required. Durable execution frameworks such as Temporal, DBOS, and Prefect have not been validated in this initial release.

# How was this change tested?

The following validation was completed:

- Adapter unit and end-to-end tests: `25 passed`
- Full test suite via `make test`: `631 passed, 12 skipped`
- Bub capture sanitization regression test: `1 passed`
- `make check`
- `make docs-test`
- `uv build --project integrations/pydantic-ai`
- `tox` across Python 3.11, 3.12, 3.13, and 3.14
- Explicit type checking of `integrations/pydantic-ai/src`
- Manual inspection of the generated wheel metadata and contents, including verification that no entry point is registered

The end-to-end test verifies the complete flow:

1. A tool result is captured.
2. The capture checkpoint is flushed.
3. The next model step recalls the prepared context.
4. A final PowerContext search returns the stored result with its citation.

An initial Python 3.12 tox run encountered an existing readiness-test timeout; both the isolated test and the complete Python 3.12 environment passed on rerun.

# AI usage statement


src/powercontext/client/capture.py
```diff
@@ -0,0 +1,130 @@
...
+    return any(part in folded for part in SENSITIVE_KEY_PARTS)
+
+
+def _codex_auth_secrets() -> set[str]:
+    codex_home = Path(os.getenv("CODEX_HOME", str(Path.home() / ".codex"))).expanduser()
+    try:
+        auth = json.loads((codex_home / "auth.json").read_text(encoding="utf-8"))
+    except (OSError, json.JSONDecodeError):
+        return set()
+    return _sensitive_values(auth)
```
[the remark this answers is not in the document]
thunguo: @copilot Fix the code for all comments in [this review comment](https://github.com/oceanbase/powercontext/pull/1295#discussion_r3822390627).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review comment.

pyproject.toml
```diff
@@ -112,13 +112,15 @@ python-version = "3.11"
 extra-paths = [
     "./integrations/claude-code/plugins/powercontext",
     "./integrations/codex/plugins/powercontext",
+    "./integrations/pydantic-ai/src",
 ]
 
 [tool.ty.src]
-exclude = ["e2e/bub", "integrations/bub", "evaluation"]
+exclude = ["e2e/bub", "integrations/bub", "integrations/pydantic-ai", "evaluation"]
```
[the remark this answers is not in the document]
thunguo: @copilot Fix the code for all comments in [this review comment](https://github.com/oceanbase/powercontext/pull/1295#discussion_r3822390705).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review comment.

integrations/pydantic-ai/src/powercontext_pydantic_ai/capability.py
```diff
@@ -0,0 +1,336 @@
...
+                        metadata=metadata,
+                    )
+                )
+            except ClientError as exc:
+                self._auth_reporter.report(exc, "event capture")
+                logger.debug("PowerContext event capture failed open: %s", type(exc).__name__)
+                return
+            except Exception as exc:  # Capture must never change an agent result.
+                logger.debug("PowerContext event capture failed open: %s", type(exc).__name__)
+                return
```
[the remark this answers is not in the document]
thunguo: @copilot Fix the code for all comments in [this review comment](https://github.com/oceanbase/powercontext/pull/1295#discussion_r3822390750).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review comment.

integrations/pydantic-ai/src/powercontext_pydantic_ai/capability.py
```diff
@@ -0,0 +1,336 @@
...
+            scope_id=state.scope_id,
+            query=query[:8192],
+            max_bytes=self.settings.max_bytes,
+        )
+        try:
+            response = await self._toolset._require_client().prepare_context(request)
+        except ClientError as exc:
+            self._auth_reporter.report(exc, "context preparation")
+            logger.debug("PowerContext context preparation failed open: %s", type(exc).__name__)
+            return None
```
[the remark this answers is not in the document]
thunguo: @copilot Fix the code for all comments in [this review comment](https://github.com/oceanbase/powercontext/pull/1295#discussion_r3822390779).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review comment.

integrations/pydantic-ai/src/powercontext_pydantic_ai/capability.py
```diff
@@ -0,0 +1,336 @@
...
+                "final" if final else "checkpoint",
+                type(exc).__name__,
+            )
+            return
+        except Exception as exc:  # Flush must never change an agent result.
+            logger.debug(
+                "PowerContext %s capture flush failed open: %s",
+                "final" if final else "checkpoint",
+                type(exc).__name__,
+            )
```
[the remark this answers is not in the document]
thunguo: @copilot Fix the code for all comments in [this review comment](https://github.com/oceanbase/powercontext/pull/1295#discussion_r3822390817).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review comment.

thunguo:
@PsiACE This PR is ready for review, plz review it when you are free, thx.

integrations/pydantic-ai/src/powercontext_pydantic_ai/capability.py
```diff
@@ -0,0 +1,345 @@
...
+            state.captured_position = max(state.captured_position, response.position)
+            if state.captured_events % self.settings.capture_checkpoint_every == 0:
+                await self._flush_locked(final=False)
+
+    async def _flush_locked(self, *, final: bool) -> None:
+        state = self._require_state()
+        if state.captured_position <= state.flushed_position:
+            return
+        try:
+            response = await self._toolset._require_client().flush_memory(FlushMemoryRequest(scope_id=state.scope_id))
```
AlexStocks: [P1] checkpoint/final flush 不能保证追到本次捕获位置

这里每次只调用一次 `flush_memory`，但 Server 每次最多处理 `source_window_limit` 个 Source；当同一 scope 已有 backlog 时，返回的 `current_cursor` 仍可能小于 `state.captured_position`。使用真实 SQLite Server、`source_window_limit=1` 和两条 backlog 的探针中，本次 tool result 位于 position 5，checkpoint 与 final flush 后 cursor 只到 1、2，下一次 model request 看不到该结果；继续手工 flush 到 5 后才可召回。这会破坏 checkpoint 的 read-your-write 语义。建议在 deadline/max-calls/no-progress 约束下循环，直到 cursor 到达本次捕获位置，并补一个返回 partial cursor 的回归测试。

src/powercontext/client/capture.py
```diff
@@ -0,0 +1,147 @@
...
+            upper_bound = candidate_length - 1
+    return rendered
+
+
+def sanitize_capture_value(value: Any) -> Any:
+    """Recursively replace values belonging to credential-like keys."""
+
+    if isinstance(value, Mapping):
+        return {
+            str(key): REDACTED if is_sensitive_key(str(key)) else sanitize_capture_value(item)
```
AlexStocks: [P1] 当前 sanitizer 会原样持久化常见凭据形态

这里的 key 归一化无法识别 camelCase `apiKey`（会变成 `apikey`，但规则只含 `api_key`）；同时非 Mapping 的字符串会原样返回，而 locked Pydantic AI 2.29 的 `ToolCallPart.args` 是 `str | dict | None`，OpenAI adapter 会直接传入 JSON 字符串。探针确认 `{"apiKey":"..."}` 和字符串形式的 `{"api_key":"..."}` 都保留了 synthetic secret。capture 会持久化 tool arguments/results，这会绕过文档承诺的 credential redaction。建议覆盖 camelCase/compact aliases，并对结构化 JSON 参数使用 `args_as_dict()` 或安全解析后递归脱敏，补两种形态的回归测试。

integrations/pydantic-ai/src/powercontext_pydantic_ai/capability.py
```diff
@@ -0,0 +1,345 @@
...
+        if not prepared_content:
+            return request_context
+
+        state.context_injected = True
+        context_request = ModelRequest(
+            parts=[SystemPromptPart(f"{CONTEXT_PREFIX}\n\n{prepared_content}")],
+            run_id=ctx.run_id,
+            conversation_id=ctx.conversation_id,
+        )
+        return replace(request_context, messages=[context_request, *request_context.messages])
```
AlexStocks, line 149: [P1] 新 run 应替换旧的 PowerContext context

这里直接把新 context 前置到完整 `message_history`，旧 run 注入的 `CONTEXT_MARKER` 消息不会被移除。连续三次 `agent.run(..., message_history=previous.all_messages())` 的公开路径探针得到 marker 数 `[1, 2, 3]`；第三轮同时携带 context-1、context-2、context-3，既让陈旧 system context 继续影响模型，也会按每轮最多 8KB 线性膨胀历史。建议在插入当前 run context 前过滤或替换旧的 PowerContext-marked request，保证复用历史时始终只有一个当前 context，并增加多 run 回归测试。

Teingi:
Please resolve conflicts

thunguo:
> Please resolve conflicts

Done.

integrations/pydantic-ai/src/powercontext_pydantic_ai/capability.py
AlexStocks: [P1] 当前 Head 只完成了部分修复。现在虽然加入了 `asyncio.timeout()`、no-progress 检查和循环，但固定 `_CAPTURE_FLUSH_MAX_CALLS=10` 会在 cursor 每次都正常推进、总 timeout 仍未耗尽时提前停止。exact Head 的真实 SQLite 探针使用 `source_window_limit=1` 和 20 条 backlog，本次 run 的 high watermark 为 24；checkpoint 与 `after_run` 结束后都检索不到刚捕获的 tool result，还要继续手工 flush 21→22→23→24 四次才可召回。当前新增测试只覆盖很小的 cursor 序列，没有穿过 10 次上限。建议在 no-progress/timeout 边界内持续追到 `captured_position`，或按实际 cursor gap 计算足够的调用上限，并增加超过 10 个窗口的公开 read-your-write 回归。

src/powercontext/client/capture.py
AlexStocks: [P1] 当前 Head 仍有同根因残留：对象型工具结果会绕过凭据清洗。`sanitize_capture_value()` 只递归处理 Mapping、list/tuple 和 JSON 字符串；Pydantic model、dataclass 等对象在 `capture.py:102` 原样返回，随后被 `json.dumps(default=str)` 压成含字段名和值的 repr。公开 Agent 探针让工具返回 `CredentialResult(apiKey="object-field-secret-sentinel")` 后，持久化的 `tool_result` 仍包含明文 `apiKey='object-field-secret-sentinel'`，没有 `[REDACTED]`。建议先把支持的 Pydantic/dataclass/value object 转为 JSON-safe 结构再递归清洗；无法安全结构化的对象不要直接 `str`，并补真实 Agent tool-result 回归。

integrations/pydantic-ai/pyproject.toml
```diff
@@ -0,0 +1,32 @@
...
+# limitations under the License.
+
+[project]
+name = "powercontext-pydantic-ai"
+version = "0.0.1"
+description = "Pydantic AI integration for PowerContext durable memory."
+readme = "README.md"
+requires-python = ">=3.11,<4"
+dependencies = [
+    "powercontext[client]>=0.0.2",
```
AlexStocks: [P1] 最低 core 版本不包含这里导入的新模块。该包声明 `powercontext[client]>=0.0.2`，但 `powercontext.client.capture` 是本 PR 才新增的模块。当前 PyPI 最高 core 仍是 0.0.2；我校验了官方 wheel，它不含 `powercontext/client/capture.py`。在干净 Python 3.11 环境安装 exact Head 构建的适配器 wheel 时，解析器合法选择 0.0.2，安装成功后 `import powercontext_pydantic_ai` 立即报 `ModuleNotFoundError`；`integrations/bub` 的新共享导入也使用同一下限。请把两个独立包的 core 下限提高到首个包含该模块的发布版本，并增加只使用发布依赖的 wheel 安装/导入 smoke test。

integrations/pydantic-ai/README.md
```diff
@@ -0,0 +1,74 @@
+# PowerContext for Pydantic AI
+
+`powercontext-pydantic-ai` connects a Pydantic AI agent to a running PowerContext Server through the public async
+Python Client. It provides three tools, prepares relevant context before model requests, and can optionally capture
+bounded agent events and flush them into Memory.
+
+## Install and use
+
+```bash
+uv add powercontext-pydantic-ai
```
AlexStocks: [P1] 这条安装命令不足以运行下面的 OpenAI 示例。当前包只依赖基础 `pydantic-ai-slim`，不会安装 OpenAI provider。干净环境仅执行这里的 `uv add powercontext-pydantic-ai` 后，构造下方 `Agent("openai:gpt-5.2")` 会直接报 `ImportError: Please install the openai package`，还未进入 PowerContext。请同时安装 `pydantic-ai-slim[openai]`，或把首个示例改为 provider-neutral 并单独说明各 provider extra；英文、中文 how-to 需要同步，并用干净环境执行文档 smoke test。

Teingi (approved):
LGTM

PR #1296, opened 2026-08-20
fix(persistence): report concurrent artifact create as revision conflict

### LINKED ISSUE #1294: bug: concurrent Memory revise races surface as HTTP 500 instead of 409 revision_conflict
### Describe the bug

`ArtifactRepository.revise` compare-and-swap locks `pc_artifact_heads`, then inserts revision N+1 into `pc_artifacts`. Two independent Server instances can both pass `rowcount` for revision N and both insert N+1. The loser raises `sqlalchemy.exc.IntegrityError`, which the HTTP layer maps to `500 internal_error` instead of `409 revision_conflict`.

The in-process `asyncio.Lock` does not serialize two processes. 

### Steps to reproduce

From a clean checkout of this repository. No Docker.

```bash
uv sync
uv run python repro_concurrent_remember.py
```

Save as `repro_concurrent_remember.py`:

```python
import asyncio
from collections import Counter
from contextlib import AsyncExitStack
from pathlib import Path

import httpx

from powercontext.builtin.runtime.config import SQLiteConfig
from powercontext.client import PowerContextClient
from powercontext.client.errors import ServerResponseError
from powercontext.http import RememberMemoryRequest
from powercontext.server.factory import create_server_app
from powercontext.server.settings import McpConfig, ServerSettings


async def main() -> None:
        database = SQLiteConfig(url=f"sqlite+aiosqlite:///{Path(directory) / 'race.db'}")
        async with AsyncExitStack() as stack:
            clients: list[PowerContextClient] = []
            for _ in range(2):
                app = create_server_app(
                    settings=ServerSettings(database=database, mcp=McpConfig(enabled=False))
                )
                await stack.enter_async_context(app.router.lifespan_context(app))
                http = await stack.enter_async_context(
                    httpx.AsyncClient(
                        transport=httpx.ASGITransport(app=app, raise_app_exceptions=False),
                        base_url="http://x",
                    )
                )
                clients.append(PowerContextClient("http://x", http_client=http))

            async def write(replica: int, index: int) -> str:
                try:
                    await clients[replica].remember_memory(
                        RememberMemoryRequest(scope_id="race", kind="fact", text=f"r{replica}-w{index}")
                    )
                except ServerResponseError as error:
                    return f"HTTP {error.status_code} {error.code}"
                return "ok"

            outcomes = await asyncio.gather(*(write(r, i) for r in range(2) for i in range(8)))

    tally = Counter(outcomes)
    print(dict(tally))
    raise SystemExit(1 if any(key.startswith("HTTP 500") for key in tally) else 0)


if __name__ == "__main__":
    asyncio.run(main())
```

`raise_app_exceptions=False` is required so the client sees the HTTP status a real deployment would return. The Server may still print an `IntegrityError` traceback; that is the bug, not a problem with the script.

### Expected behavior

Exit code 0. Losing writes are only `HTTP 409 revision_conflict`, for example:

```
{'ok': 8, 'HTTP 409 revision_conflict': 8}
```

### Actual behavior

Exit code 1. Typical result (three consecutive runs on this machine):

```
{'ok': 8, 'HTTP 409 revision_conflict': 7, 'HTTP 500 internal_error': 1}
```

```
sqlalchemy.exc.IntegrityError: UNIQUE constraint failed: pc_artifacts.scope_id, pc_artifacts.family, pc_artifacts.artifact_id, pc_artifacts.revision
INSERT INTO pc_artifacts ...
```

Acknowledged writes still persist. The defect is the error code, not silent data loss.

### Environment

```
- PowerContext version: 0.0.1 (git 2186dbe0b2d39364803be0f80f960dc669d78452)
- Python version: 3.14
- OS: macOS 15.2
```

Ethan-Xingyue:

Closes #1294.

# Rationale for this change

`revise` is not the culprit. Its compare-and-swap detects conflicts through
`UPDATE ... WHERE revision = N` and a rowcount — a write, which cannot be raced.

`create` decides "this lifecycle already exists" from a `SELECT`, and a read cannot
exclude a writer that commits before the insert. Both instances see no head, both
insert revision 1, and the loser's `IntegrityError` matches no domain error, falling
through to the HTTP 500 default.

Only Memory reaches this, because its Artifact id is a per-scope constant. Candidate
and approved Artifact ids are uuid4 and never collide.

`ArtifactRepository.create` now takes the unique-constraint violation as the
authoritative conflict signal. The inserts are wrapped in `try/except IntegrityError`;
a committed head turns the failure into `RevisionConflictError`. With no committed
head the error is re-raised, so foreign key violations and other genuine integrity
failures are unchanged.

The losing concurrent create returns `409 revision_conflict` instead of
`500 internal_error`. No API, schema, or persisted format changes.

# How was this change tested?

- Added `test_lifecycle_committed_after_the_head_read_conflicts`, covering the window
  where a lifecycle is committed between the head read and the insert.
- The reproduction script from #1294, five runs each way. Before:
  `{'ok': 8, 'HTTP 409 revision_conflict': 7, 'HTTP 500 internal_error': 1}`, exit 1.
  After: `{'ok': 8, 'HTTP 409 revision_conflict': 8}`, exit 0.
- `pytest --doctest-modules --ignore=tests/e2e`: 563 passed, 1 skipped. `ty check` and
  ruff clean. E2E not run.

### DIFF
--- a/src/powercontext/builtin/persistence/artifacts.py
+++ b/src/powercontext/builtin/persistence/artifacts.py
@@ -21,6 +21,7 @@ from typing import Any
 
 from pydantic import BaseModel
 from sqlalchemy import insert, select, update
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncConnection
 
 from powercontext.artifacts import (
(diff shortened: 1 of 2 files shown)


Teingi (approved):
LGTM

Issue #1298, opened 2026-08-21
Tracking Issue: local Server availability and service installation


knqiufan:
## Status

RFC #1299 has been accepted and merged. This issue now tracks implementation of the accepted design. The merged RFC is the source of truth for the public contract, responsibility boundaries, security rules, and platform qualification requirements.

Implementation may be delivered through independently reviewed pull requests. A platform must continue to report `unsupported` until its adapter ships and passes the required base and native lifecycle tests on a matching runner.

## Problem

`powercontext server run` is intentionally a foreground process. Agent integrations depend on a reachable Server and fail open when it is unavailable, so closing the terminal or restarting a personal machine can leave recall and capture unavailable during later agent sessions.

PowerContext needs a clear boundary among:

- host-visible Server-unavailability diagnostics in agent integrations;
- correlated installed-environment diagnostics in `powercontext doctor`;
- opt-in native per-user service installation for personal use through a distribution/service-install layer;
- container or administrator-managed services for managed deployments.

The Server role continues to own foreground process construction and `powercontext server run`. Integrations remain fail-open and do not install, start, or supervise the Server. Native service adapters belong to the distribution/service-install layer rather than `src/powercontext/server`.

## Accepted design

- RFC: #1299
- RFC state: accepted and merged
- Personal lifecycle commands:

  ```text
  powercontext service install
  powercontext service uninstall
  powercontext service status
  ```

- The commands manage one opt-in service for the current operating-system user. There are no named profiles or privileged/system-wide installation modes.
- The intended endpoint comes from local `ServerSettings` and must use a loopback bind. Client or integration `server_url` values do not select the local service registration.
- `service install` performs endpoint preflight, installs or reconciles a verified PowerContext-owned native artifact, enables it for future logins, and starts it immediately unless a valid PowerContext Server is already live.
- `service uninstall` stops a manager-owned process before removing the verified registration. It never kills a process only because it occupies the intended port.
- `service status` reports independent support, registration, definition, manager, liveness, and log-location facts in human-readable and JSON forms.
- Service artifacts contain no complete environment, API keys, bearer tokens, or provider credentials. File-backed settings and secrets remain a separate concern under the existing `pydantic-settings` direction and are not a general-availability condition for this work.

### Design

- [x] Accept RFC #1299
- [x] Fix the public lifecycle interface at `powercontext service install|uninstall|status`
- [x] Keep configuration and credential handoff separate from the service lifecycle contract
[...]

PR #1299, opened 2026-08-21
docs(rfc): define local Server service-installation boundaries

knqiufan:

Tracking issue: #1298

# Rationale for this change

The local Server currently has an intentional foreground entry point, while agent integrations depend on a reachable Server and fail open when it is unavailable. Maintainer feedback on #1298 asked that persistent startup be designed in a distribution/service-install layer rather than the Server CLI, with visible integration diagnostics, clear `doctor` output, and a separate boundary for managed deployments. This RFC establishes that design before implementation.

- Defines ownership boundaries among the Server role, agent integrations, CLI diagnostics, the personal distribution/service-install layer, and managed deployments.
- Proposes an opt-in per-user service contract for Linux, macOS, and Windows without privilege elevation or a PowerContext supervisor.
- Defines service state, diagnostics, security, upgrade, testing, delivery-splitting, alternatives, and unresolved questions.
- Contains no implementation code or released API change.

No released behavior changes. The PR is a design proposal for review. If accepted and implemented later, the foreground `powercontext server run` behavior remains unchanged and persistent personal service installation remains opt-in.

# How was this change tested?

- `uv run prek run --files docs/en/rfcs/0000_local_server_availability_and_service_installation.md docs/zh/rfcs/0000_local_server_availability_and_service_installation.md`
- `uv run zensical build -s`
- `git diff --check`

# AI usage statement

OpenAI Codex (GPT-5) was used to analyze the issue discussion and help draft and cross-check the synchronized English and Chinese RFC documents. The proposal is submitted by the author for maintainer review and revision through the RFC process.

### DIFF
--- a/docs/en/rfcs/1299_local_server_availability_and_service_installation.md
+++ b/docs/en/rfcs/1299_local_server_availability_and_service_installation.md
@@ -1,6 +1,6 @@
 - Proposal Name: local_server_availability_and_service_installation
 - Start Date: 2026-08-21
-- RFC PR: [oceanbase/powercontext#0000](https://github.com/oceanbase/powercontext/pull/0000)
+- RFC PR: [oceanbase/powercontext#1299](https://github.com/oceanbase/powercontext/pull/1299)
 - Tracking Issue: [oceanbase/powercontext#1298](https://github.com/oceanbase/powercontext/issues/1298)
 
 # Summary
--- a/docs/en/rfcs/1299_local_server_availability_and_service_installation.md
+++ b/docs/en/rfcs/1299_local_server_availability_and_service_installation.md
@@ -61,18 +61,25 @@ integration does not create persistent operating-system state.
 
 ### Persistent personal use
 
-A user who wants the local Server after login explicitly installs a per-user service through the distribution layer:
+A user who installed both the CLI and the ready-to-run Server role can explicitly install a per-user service:
 
 ```text
 powercontext service install
 powercontext service status
 ```
 
-The public lifecycle command group is `powercontext service`. It is not part of `powercontext server` and manages only
-PowerContext-owned registration artifacts for the current user. Installation is off by default and does not require
-`root`, `SYSTEM`, or an administrator account.
+The public lifecycle command group is `powercontext service`. It is not part of `powercontext server` and manages
+exactly one PowerContext-owned personal Server registration for the current operating-system user. The initial
+contract has no named service profiles. Installation is off by default and does not require `root`, `SYSTEM`, or an
+administrator account.
 
-The installed service still executes the same Server entry point. It does not introduce a daemon inside PowerContext:
+The personal service accepts only a loopback Server bind. Its endpoint comes from the local Server settings used by the
+registration, not from a Client `server_url` or a remote `powercontext doctor --server-url` target. Additional local
+instances remain foreground processes; shared, remotely addressed, or publicly bound Servers are managed deployments.
+
+The native registration invokes a distribution-owned internal launcher. After a duplicate-prevention preflight, that
+launcher transfers control in the same manager-owned process to the same Server runner used by the foreground entry
+point; it does not remain as a second daemon or supervisor:
 
 ```text
 powercontext server run
@@ -90,8 +97,11 @@ by PowerContext. It does not terminate an unrelated foreground Server.
 ### Managed deployment
 
 The personal service installer is not a production deployment manager. A managed installation uses the project's
-container image or an administrator-managed systemd system unit, launch daemon, Windows service, or equivalent
+container image or an administrator-managed `systemd` system unit, launch daemon, Windows service, or equivalent
 orchestrator. PowerContext does not install those privileged resources through `powercontext service`.
+
+Client and integration settings may continue to target such a remote Server. The personal-service commands neither
+register nor mutate remote endpoints, and diagnostics must not recommend local service installation for them.
 
 ## What users see when the Server is unavailable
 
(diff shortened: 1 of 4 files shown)


docs/en/rfcs/1299_local_server_availability_and_service_installation.md
```diff
@@ -0,0 +1,431 @@
...
+
+The initial personal-service profile therefore relies on configuration that is available to the native user-service
+environment. Status and doctor must detect common configuration divergence where it can be observed without reading
+secrets. Deployments that require injected credentials or environment management beyond the native user-service
+contract remain operator-managed.
+
+A portable credential store, environment snapshot, or cross-platform secret-file format is not introduced by this
+RFC. If personal service installation cannot support the project's documented inference configuration without one of
+those contracts, that configuration handoff must be resolved before the service-install implementation is declared
+generally available.
```
PsiACE: Configuration should be handled separately. The existing `pydantic-settings` dependency provides a mature solution for file-backed settings and secrets; this RFC should not make configuration handoff a GA condition.

docs/en/rfcs/1299_local_server_availability_and_service_installation.md
```diff
@@ -0,0 +1,431 @@
...
+command while offering a separate install or service command for persistent personal use.
+
+Containers and administrator-managed services are established deployment boundaries for managed workloads because
+they make identity, configuration, credentials, restart policy, and observability explicit. This RFC applies that
+separation to PowerContext instead of treating every local or managed Server as the same autostart problem.
+
+# Unresolved questions
+
+- Should the public distribution command remain powercontext service, or fit under an existing setup-oriented command
+  without implying that the Server role owns installation?
```
PsiACE: The public command should be `powercontext service`. This is an ongoing lifecycle interface with `install`, `status`, and `uninstall`; it does not belong under a setup-oriented command.

docs/en/rfcs/1299_local_server_availability_and_service_installation.md
```diff
@@ -0,0 +1,431 @@
...
+- What host-native presentation satisfies visible but non-blocking Server-unavailable diagnostics for Codex, Claude
+  Code, and DeepSeek Harness without producing warning fatigue?
+- Which documented personal configurations are guaranteed to be available inside each native user-service
+  environment, and does that require a separate non-secret configuration contract before general availability?
+- Should install start the service immediately by default, or only register it for the next login unless an explicit
+  start option is provided?
```
PsiACE: `service install` should start the service immediately by default. The command is already explicit opt-in, and line 176 already defines this behavior.

docs/en/rfcs/1299_local_server_availability_and_service_installation.md
```diff
@@ -0,0 +1,431 @@
...
+Linux, macOS, and Windows while allowing the accepted implementation to be split into independently reviewed changes.
+
+# Motivation
+
+Agent integrations depend on a reachable PowerContext Server but do not own its process lifecycle. The normal local
+entry point is intentionally foreground:
+
+~~~text
+powercontext server run
+~~~
```
PsiACE: Please replace `~~~` fences with triple backticks throughout, and use inline backticks for commands, flags, paths, and identifiers.

docs/en/rfcs/1299_local_server_availability_and_service_installation.md
knqiufan: Thank you. I updated both RFC documents to keep file-backed settings and secrets as a separate concern under the existing `pydantic-settings` direction, and removed configuration handoff as a GA condition.

docs/en/rfcs/1299_local_server_availability_and_service_installation.md
knqiufan: Thank you. The RFC now fixes the public lifecycle interface at `powercontext service` and removes setup placement from the unresolved questions.

docs/en/rfcs/1299_local_server_availability_and_service_installation.md
knqiufan: Thank you. I adjusted this section that `powercontext service install` starts the service immediately by default unless the intended endpoint is already live, and removes the contradictory unresolved question.

docs/en/rfcs/1299_local_server_availability_and_service_installation.md
knqiufan: Thank you. I replaced the tilde fences with triple backticks and applied inline code formatting for commands, flags, paths, and identifiers across both the English and Chinese RFC documents.

knqiufan:
@PsiACE I have pushed a second RFC-only update (`9313fd9`) after addressing the four inline review comments in the preceding commit.

The additional changes clarify how the personal service should work:
- Each OS user can have one PowerContext personal service. It is available only when the local Server components are installed, listens only on localhost, and does not manage remote or production deployments.
- Before starting the Server, a small internal launcher checks whether one is already running. This prevents duplicate processes and reports a clear conflict if another program is using the address.
- Installation avoids leaving incomplete service files. If registration succeeds but startup fails, the valid registration is kept so the user can inspect the logs and retry. Running powercontext service install again also refreshes outdated paths, versions, and service definitions.
- Uninstall stops the process owned by the service manager before removing the registration, avoiding an untracked process that continues running.
- powercontext service status reports whether the service is installed, current, running, reachable, and where its logs are stored. However, because the personal service is optional, not installing it—or using a platform where it is not yet supported—does not by itself make powercontext doctor fail.
- Linux, macOS, and Windows each have defined logging, restart, and ownership rules. A platform is advertised as supported only after the service lifecycle has been tested on that operating system.
- Integration warnings must appear somewhere the user can actually see. They must not expose sensitive content, and repeated Server-unavailable warnings must be limited or deduplicated to avoid notification spam.

The English and Chinese RFCs are synchronized; the focused pre-commit checks and strict documentation build both pass.

Could you please take another look and let me know whether these contracts match the intended project direction, or whether any part should be narrowed or adjusted? Thank you.

PsiACE (approved):
totally lgtm

PR #1303, opened 2026-08-21
fix(memory): enforce authoritative revision integrity

### LINKED ISSUE #1297: bug: Memory commit accepts malformed revisions and search can return text that entries() rejects
### Describe the bug

RFC 0014 and the Memory protocol name the backend as the final trust boundary: `commit()` must validate and atomically persist one complete Revision in the same transaction, and search must filter every candidate through the authoritative head manifest. The current implementation does not honor that contract.

This is not a default HTTP/MCP remember, revise, retire, or flush incident. Those paths go through `MemoryService`, which currently fills predecessor, hash, and projections itself. The gap is the public SPI: `MemoryCommit`, `MemoryWritePlan`, and `MemoryBackend` allow a writer to bypass the service and call `backend.commit()` directly. Tests and a later writer can already take that seam.

**What the contract says, and what the code does**

1. The protocol requires `MemoryUnitOfWork.commit()` to validate and atomically commit one complete Revision. `_validate_commit()` only checks family, the whole-document `content_hash`, revision +1, base identity, and equality of the active/projection version-id sets.
2. The RFC sequence diagram requires the backend to validate identity, predecessor, manifest, hash, and vectors. The implementation does not recompute the entry body hash, validate the direct predecessor, or require projections to match the authoritative entry.
3. The RFC schema draws a self-referencing foreign key on `previous_version_id`. `pc_memory_entry_versions.previous_version_id` is a plain string column.
4. The RFC requires search to revalidate active membership, exact version, and canonical bytes against the authoritative manifest. The implementation only confirms that the target is still the current head, so the index can act as a second content authority.
5. The RFC describes projections and indexes as rebuildable, non-authoritative copies. Hit text can come from an unrevalidated index/body and no longer line up with the manifest hash.

**What has already been reproduced**

1. A revision-1 entry with `previous_version_id="does-not-exist"` commits and reads back.
2. A body replaced with `tampered searchable body.` while keeping the hash of `User prefers black tea.` also commits.
3. In that corrupted state, `search()` returns the tampered body, while `entries()` raises `InvalidMemoryCitationError` for the same revision.
4. `entries()` also omits the `version.entry_content_hash` comparison that `_validate_anchor()` already performs.

Closing only the write path or only the read path leaves the loop open. Both sides should be fixed in one PR.

**What this gap can cause later**

Official product paths almost never construct a bad envelope today, so normal use usually does not show the damage. If the seam is taken, or if the service or a second writer later emits a bad commit, the likely consequences are:

1. Illegal history becomes durable authority. A hash is then a label written with the envelope, not a commitment.
2. A fake predecessor has almost no current read consumer, but later lineage, offline validation, or citation code that walks `previous_version_id` will not be able to rebuild history reliably.
3. After a body / declared / manifest hash mismatch, remember, revise, forget, and organize fail with `hash-mismatch`. The Memory can still be recalled but can no longer be governed.
4. If a projection or `searchable_text` diverges from the canonical row, search and automatic `POST /v1/context/prepare` injection can hand a different body to the model.
5. Context Pack and Handoff can consume text the authoritative manifest never committed.
6. A new adapter, migration script, or homemade `MemoryWritePlan` will copy the current thin checks as the contract, and the gap will spread.
7. Fixing only one side still lets dirty data in, or still lets search return it after it is in.

### Steps to reproduce

1. Use a PowerContext checkout with its dependencies installed.
2. Save the following as `repro_memory_integrity.py`.
3. Run:

```bash
python repro_memory_integrity.py
```

```python
#!/usr/bin/env python3
"""Reproduce Memory commit/search integrity gap."""

from __future__ import annotations

import asyncio
import json

from powercontext.builtin.artifacts.memory import (
    Memory,
    MemoryCommit,
    MemoryContent,
    MemoryEntryVersion,
    MemoryManifest,
    MemoryManifestEntry,
    MemoryProjection,
    MemoryService,
)
from powercontext.builtin.artifacts.memory.canonical import entry_content_hash, memory_content_hash
from powercontext.builtin.artifacts.memory.errors import InvalidMemoryCitationError
from powercontext.builtin.artifacts.search import analyze_text
from powercontext.builtin.persistence.memory import RelationalMemoryBackend
from powercontext.builtin.runtime import BuiltinConfig, open_builtin_contexts

ORIGINAL = "User prefers black tea."
TAMPERED = "tampered searchable body."


def _version(*, text: str, content_hash: str, previous_version_id: str | None) -> MemoryEntryVersion:
    return MemoryEntryVersion(
        memory_artifact_id="memory",
        entry_id="pref-tea",
        entry_version_id="pref-tea-v1",
        version=1,
        previous_version_id=previous_version_id,
        kind="fact",
        text=text,
        entry_content_hash=content_hash,
        created_in_revision=1,
    )


def _commit(version: MemoryEntryVersion) -> MemoryCommit:
    content = MemoryContent(
        manifest=MemoryManifest(
            entries=(
                MemoryManifestEntry(
                    entry_id=version.entry_id,
                    entry_version_id=version.entry_version_id,
                    entry_content_hash=version.entry_content_hash,
                    state="active",
                ),
            )
        )
    )
    memory = Memory(artifact_id="memory", revision=1, content=content)
    return MemoryCommit(
        base=None,
        memory=memory,
        content_hash=memory_content_hash(content),
        entry_versions=(version,),
        projections=(
            MemoryProjection(
                entry_version=version,
                searchable_text=analyze_text(version.text),
            ),
        ),
    )


def _backend(contexts: object, scope_id: str) -> RelationalMemoryBackend:
    return RelationalMemoryBackend(
        database=contexts.database,
        scope_id=scope_id,
        artifacts=contexts.repositories.artifacts,
        index=contexts.index,
    )


async def main() -> None:
    original_hash = entry_content_hash(
        kind="fact",
        text=ORIGINAL,
        source_refs=(),
        artifact_refs=(),
    )
        config = BuiltinConfig(
            database=SQLiteConfig(url=f"sqlite+aiosqlite:///{directory}/repro.db")
        )
        async with open_builtin_contexts(config) as contexts:
            fake_backend = _backend(contexts, "fake-predecessor")
            fake_version = _version(
                text=ORIGINAL,
                content_hash=original_hash,
                previous_version_id="does-not-exist",
            )
            async with fake_backend.begin() as unit_of_work:
                committed = await unit_of_work.commit(_commit(fake_version))
            stored = (await fake_backend.entries(committed.as_ref()))[0]

            mismatch_backend = _backend(contexts, "hash-mismatch")
            mismatch_version = _version(
                text=TAMPERED,
                content_hash=original_hash,
                previous_version_id=None,
            )
            async with mismatch_backend.begin() as unit_of_work:
                mismatch_memory = await unit_of_work.commit(_commit(mismatch_version))
            service = MemoryService(backend=mismatch_backend)
            search = await service.search(
                "tampered searchable",
                memories=(mismatch_memory,),
                mode="fts",
            )
            entries_error = None
            try:
                await service.entries(mismatch_memory)
            except InvalidMemoryCitationError as error:
                entries_error = type(error).__name__

            observed = {
                "entries_error": entries_error,
                "fake_predecessor": stored.previous_version_id,
                "search_hits": [hit.text for hit in search.hits],
            }
            print(json.dumps(observed, sort_keys=True))
            expected = {
                "entries_error": "InvalidMemoryCitationError",
                "fake_predecessor": "does-not-exist",
                "search_hits": [TAMPERED],
            }
            if observed != expected:
                raise SystemExit(1)
            print("0")


if __name__ == "__main__":
    asyncio.run(main())
```

4. Observe that the process prints the JSON below and exits 0. That means it observed the known buggy state.

Do not use HTTP remember/flush as the minimal reproduction. That path currently fills the envelope correctly and does not prove the contract gap.

### Expected behavior

The backend is the final trust boundary. `commit()` must reject an illegal Revision in the same transaction, including:

1. Nonexistent, cross-entry, cross-Memory, or skipped predecessors, and a first version with a non-null predecessor.
2. Body / declared-hash / manifest-hash mismatches.
3. Duplicate, missing, or extra entry versions, and projections that do not match the active manifest entry or canonical `searchable_text`.
4. A failed validation must roll back the whole transaction.

Search must treat the index as a rebuildable projection, not a second authority:

1. Before returning, each hit must be revalidated in the same read transaction against the exact revision, manifest, and entry: active membership, identity, version, and canonical bytes.
2. For a given corrupted state, `search()`, `entries()`, and `expand()` must agree: they either succeed together and return the same authoritative body, or fail together with an integrity error.
3. They must never return the suspect body.

The RFC-legal degradation “authoritative rows present, vector absent” must remain accepted. Existing stale-head concurrency semantics should stay unchanged.

### Actual behavior

The reproducer consistently prints:

```json
{"entries_error": "InvalidMemoryCitationError", "fake_predecessor": "does-not-exist", "search_hits": ["tampered searchable body."]}
0
```

That matches the reproduced items above:

1. The fake predecessor was stored as an authoritative row.
2. The body/hash mismatch was accepted.
3. `search()` returned the tampered body while `entries()` rejected the same revision.

The final `0` and process exit code 0 mean the reproducer observed the known buggy state.

### Environment

- PowerContext version: current mainline
- Commit: `7bddf6abde482b6081b9984d1ca0dbae183e3946`
- Storage backend: local SQLite
- External services: none

Related but not a duplicate: #1249 fixed a concurrent stale-head FTS miss and did not add authoritative candidate validation. PR #1293 added seekDB and did not change these validators.

knqiufan:

Closes #1297.

# Rationale for this change

`MemoryUnitOfWork.commit()` is the final persistence trust boundary, but it previously accepted incomplete or internally inconsistent revisions. A malformed direct SPI writer could persist illegal predecessor links or body/hash mismatches, and search could then return text that authoritative entry reads rejected.

- Validate the stored base, every history row reachable from its manifest heads, and the complete base-to-revision transition inside the commit transaction.
- Require the manifest, changes, new entry versions, direct predecessors, and active projections to agree exactly.
- Recompute entry hashes and validate declared/manifest hashes, canonical searchable text, and configured embedding bindings before any write.
- Enforce scope-global `entry_version_id` identity with a database unique index, including startup migration and duplicate-data detection for existing SQLite and OceanBase schemas; translate insert races into `MemoryBackendConfigurationError` with full rollback.
- Reject semantic revisions of inactive entries while retaining the narrow `organize(normalize)` path when canonical content bytes are unchanged.
- Validate dense embeddings by dimension, finite values, and unit-norm tolerance without a second exact floating-point normalization comparison.
- Revalidate `entries()`, `expand()`, projection reads, and every FTS/vector candidate against the exact authoritative revision, manifest, entry row, and current head projection.
- Add regression coverage for malformed/cross-entry/cross-Memory/skipped predecessors, pre-existing corrupted history, missing/extra/duplicate rows, hash and projection corruption, inactive revisions, dense vectors, migrations, rollback behavior, and read/search/expand agreement.

Malformed commits made through the public Memory backend SPI now fail with `MemoryBackendConfigurationError` instead of becoming durable state. Existing SQLite and OceanBase schemas receive the scope-global entry-version unique index during startup; startup reports `MemoryBackendConfigurationError` if legacy duplicate identities must be repaired first. Valid service workflows and the documented no-vector degradation remain unchanged. There are no HTTP API changes.

# How was this change tested?

- `uv lock --locked`
- `uv run --no-sync prek run -a`
- `uv run --no-sync ty check`
- `uv run --no-sync python -m pytest tests/builtin/artifacts/memory tests/builtin/persistence/test_memory.py tests/builtin/persistence/test_memory_integrity.py tests/builtin/persistence/test_sqlite_profile.py tests/builtin/persistence/test_oceanbase_profile.py tests/e2e/test_memory_search_concurrency.py tests/e2e/test_builtin_runtime.py -q` (`60 passed, 7 skipped`)
- Full Windows run: `620 passed, 15 skipped`; 8 unrelated pre-existing Windows failures remain in symlink privileges, the unavailable `python3` command, POSIX mode bits, encoding, and path-separator assertions.
- The live two-connection OceanBase race regression is opt-in through `POWERCONTEXT_TEST_OCEANBASE_URL`; it is skipped when no dedicated OceanBase test database is configured.

# AI usage statement

OpenAI Codex (GPT-5) was used to inspect the issue and review feedback, implement the fixes, design regression tests, and review the final change.


src/powercontext/builtin/persistence/memory.py
```diff
@@ -491,6 +606,157 @@ async def _commit(self, connection: AsyncConnection, value: MemoryCommit) -> Mem
...
+        rows = (
+            await connection.execute(
+                select(MEMORY_ENTRY_VERSIONS_TABLE).where(
+                    MEMORY_ENTRY_VERSIONS_TABLE.c.scope_id == self._scope_id,
+                    MEMORY_ENTRY_VERSIONS_TABLE.c.entry_version_id.in_(tuple(manifest)),
+                )
+            )
+        ).mappings()
+        stored_rows = tuple(rows)
+        if any(str(row["entry_version_id"]) in new_by_id for row in stored_rows):
```
Teingi, line 698: **[P1] Make version-ID uniqueness database-enforced**

Could we make this uniqueness check database-enforced? The SELECT and later INSERT are not atomic. On OceanBase, two transactions can both observe no row and insert the same `entry_version_id` for different `memory_artifact_id` values, because the current primary key and unique constraints include the Memory ID. Both commits can therefore succeed, leaving an ID that this code treats as scope-global duplicated. Please add a database-level `UNIQUE (scope_id, entry_version_id)` with an explicit migration for existing tables, translate the resulting integrity error, and cover it with a coordinated two-connection race where both reads complete before either insert.

src/powercontext/builtin/persistence/memory.py
```diff
@@ -491,6 +606,157 @@ async def _commit(self, connection: AsyncConnection, value: MemoryCommit) -> Mem
...
+        return versions
+
+    async def _validate_new_version_history(
+        self,
+        connection: AsyncConnection,
+        value: MemoryCommit,
+        canonical_base: Memory | None,
+        new_by_entry: Mapping[str, MemoryEntryVersion],
+    ) -> None:
+        previous_ids = tuple(
```
Teingi, line 720: **[P1] Validate history already referenced by the base**

This only collects predecessors for `value.entry_versions`, so an invalid version already referenced by `canonical_base` is never rechecked. I reproduced this by changing a persisted v1 row to `previous_version_id="does-not-exist"` and then appending an unrelated entry: the commit succeeds and the new Revision still references the broken history. Because older releases could already persist this state, please either validate the base-referenced chain here or provide a one-time audit/migration plus database constraints that make this assumption safe.

src/powercontext/builtin/persistence/memory.py
```diff
@@ -530,6 +802,112 @@ def _validate_commit(value: MemoryCommit) -> None:
...
+            and current.state == "active"
+        )
+        target = current.entry_version_id
+    elif change.op == "revise":
+        valid = (
+            previous is not None
+            and change.from_entry_version_id == previous.entry_version_id
+            and change.to_entry_version_id == current.entry_version_id
+            and current.entry_version_id != previous.entry_version_id
+            and current.state == previous.state
```
Teingi, line 947: **[P2] Reject semantic revisions of inactive entries**

This accepts a `revise` whenever the state is unchanged, including `inactive -> inactive`. A direct SPI writer can therefore deactivate an entry containing “black tea” and commit an inactive revision containing “green tea”, although the Memory contract requires reactivation before a semantic revision. `organize(normalize)` still needs a narrow exception, so please allow an inactive revision only when the new and predecessor canonical content bytes are identical and the change is normalization; otherwise require reactivation first.

Teingi (changes requested):
Thanks for tightening the Memory integrity boundary. The original issue is real and the main reproducer is addressed, but the three cases below still allow revisions outside the documented contract. Requesting changes until these integrity gaps are closed.

src/powercontext/builtin/persistence/memory.py
```diff
@@ -491,6 +606,157 @@ async def _commit(self, connection: AsyncConnection, value: MemoryCommit) -> Mem
...
+                model=profile.model,
+                dimension=profile.dimension,
+                distance=profile.distance,
+                normalization=profile.normalization,
+                entry_content_hash=projection.entry_version.entry_content_hash,
+            )
+        except (TypeError, ValueError) as error:
+            raise _InvalidMemoryCommitError("vector") from error
+        if canonical != projection.embedding or expected_hash != projection.embedding_content_hash:
+            raise _InvalidMemoryCommitError("vector")
```
AlexStocks, line 820: [P1] 不要用精确浮点相等判断单位向量已经规范化

`MemoryService._embed_texts()` 已先执行一次 `canonical_embedding()`；这里再次归一化后再做 tuple 精确比较。单位归一化不是位级幂等的：公开 `MemoryService.remember()` 探针让模型返回 `(0.2407121489724894, -0.9705965492093231)`，服务层得到 `(0.24071214897248938, -0.9705965492093231)`，这里二次归一化又把首项改成 `0.24071214897248935`，于是合法的有限非零向量被抛为 `_InvalidMemoryCommitError("vector")`，整个 Memory 写入失败。现有向量测试只使用 `(1, 0, 0)` 这类轴向量，无法覆盖该舍入路径。请改为校验维度、有限性和带容差的单位范数（或确保只规范化一次并复用结果），并补一个非轴向量的公开 `MemoryService.remember()` 回归测试。

src/powercontext/builtin/persistence/memory.py
knqiufan: Thank you for catching this. I reproduced the two-transaction gap and confirmed that the SELECT/INSERT check was not sufficient. This is now fixed with a database-enforced unique index on (scope_id, entry_version_id), startup migrations for existing SQLite and OceanBase tables with duplicate-data detection, IntegrityError translation with rollback, and a coordinated opt-in OceanBase two-connection regression test.

src/powercontext/builtin/persistence/memory.py
knqiufan: Thank you for the detailed reproducer. I verified that an unrelated append could carry forward a broken predecessor already referenced by the canonical base. This is now fixed: the commit boundary validates every base-referenced chain back to v1, including identity, version sequence, predecessor links, revision ordering, and canonical hashes. A regression test covers corruption in a deeper ancestor.

src/powercontext/builtin/persistence/memory.py
knqiufan: Thank you for pointing this out. I reproduced an inactive semantic revision and confirmed that it violated RFC 0014. This is now fixed: inactive revisions are rejected unless they are the narrow organize(normalize) case and the predecessor/new canonical content bytes are identical. Both the rejection and the allowed normalization path are covered by regression tests.

src/powercontext/builtin/persistence/memory.py
knqiufan: 感谢指正。我使用你给出的非轴向量复现并确认了这个问题：二次归一化后的精确浮点比较确实会误拒绝合法写入。现已修复为校验维度、有限性和带容差的单位范数，同时保留 profile/content hash 校验，并补充了公开 MemoryService.remember() 稠密向量回归测试及非单位向量拒绝测试。

src/powercontext/builtin/persistence/memory.py
Teingi: Thanks for adding the unique index. I tested this exact head against a real OceanBase tenant, and the collision path still does not roll back the artifact write. The coordinated race produced one successful commit and one `MemoryBackendConfigurationError`, but a fresh-connection audit found `pc_artifacts=2` and `pc_artifact_heads=2` while only one entry version/projection existed. The failed Memory therefore remained visible with a manifest that referenced a missing version. Could we make the whole Memory commit atomic on OceanBase and extend this regression to verify artifact/head/version/projection rows from a fresh connection after the losing writer fails?

src/powercontext/builtin/persistence/memory.py
Teingi: The in-memory rounding issue is improved, but `1e-12` is still too strict after an OceanBase vector round trip. With the configured real 1024-dimensional embedding model, the first write succeeded; after reload, the stored vector had norm `1.0000000727221907` (dimension, finiteness, and embedding hash were all valid). Reusing that projection on the next append failed with `_InvalidMemoryCommitError("vector")` because the `7.27e-8` storage error exceeds the tolerance. Could we use a storage-compatible tolerance or renormalize after hydration, and add an OceanBase write/read/append regression?

knqiufan: done. Please take a moment to review again to see if there are any other issues.

PR #1306, opened 2026-08-21
feat(sqlite)!: replace Vec1 with bundled sqlite-vec

PsiACE:
# Rationale for this change

Replace the external Vec1 dependency with bundled sqlite-vec so SQLite vector search works without manual extension installation.

SQLite now uses sqlite-vec for vector and hybrid search. `SQLiteConfig.vec1_extension` has been removed without backward compatibility.

# How was this change tested?

- `make check`
- `make test`: 629 passed, 9 skipped
- `make docs-test`

# AI usage statement

OpenAI Codex was used to implement and test this change.

### DIFF
--- a/docs/en/development/memory-layer.md
+++ b/docs/en/development/memory-layer.md
@@ -90,20 +90,14 @@ the same citation fields through list and exact-read operations.
 unavailable. Explicit `vector` and `hybrid` requests fail when the configured profile does not provide that
 capability.
 
-## Enable SQLite Vec1
+## Enable SQLite vector search
 
-SQLite vector search is enabled only when both a Vec1 0.7 or newer loadable extension and an embedding model are
-supplied. PowerContext does not install or build the native extension; provide a compatible library for the target
-operating system and architecture:
+SQLite vector search is enabled when an embedding model is supplied. The `powercontext[builtin]` extra bundles
+`sqlite-vec`, so no extension path or separate native-library installation is required:
 
 ```python
-from pathlib import Path
-
 config = BuiltinConfig(
-    database=SQLiteConfig(
-        url="sqlite+aiosqlite:///powercontext.db",
-        vec1_extension=Path("/opt/sqlite-extensions/vec1"),
-    )
+    database=SQLiteConfig(url="sqlite+aiosqlite:///powercontext.db")
 )
 async with open_builtin_runtime(
     config,
@@ -112,7 +106,7 @@ async with open_builtin_runtime(
     ...
 ```
 
-The SQLite profile composes FTS5 and Vec1 strategies. It reports `fts`, `vector`, and `hybrid` through Memory
+The SQLite profile composes FTS5 and sqlite-vec strategies. It reports `fts`, `vector`, and `hybrid` through Memory
 capabilities.
 Stored projections and query vectors must use the same `EmbeddingProfile`, including model name, dimension, distance,
 and normalization. Changing that profile requires rebuilding projections before vector search resumes.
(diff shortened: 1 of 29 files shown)


PsiACE:
This is a breaking change because sqlite-vec is not compatible with Vec1

src/powercontext/builtin/persistence/sqlite/profile.py
```diff
@@ -118,10 +118,8 @@ def _create_database_directory(value: str) -> None:
 def _configure_sqlite(engine: AsyncEngine, config: SQLiteConfig) -> None:
     @event.listens_for(engine.sync_engine, "connect")
     def set_pragmas(dbapi_connection: DBAPIConnection, _connection_record: object) -> None:
-        extension = config.vec1_extension
-        if extension is not None:
-            run_async = dbapi_connection.run_async
-            run_async(lambda connection: _load_extension(connection, extension))
+        run_async = dbapi_connection.run_async
+        run_async(_load_sqlite_vec)
```
Teingi: [P1] Load sqlite-vec only for vector-enabled profiles

`SQLiteProfile.open(SQLiteConfig(), tables=SHARED_TABLES)` now executes this for every SQLite connection, even when no embedding model is configured and `open_builtin_contexts()` only constructs the FTS index. I reproduced this by making `sqlite_vec.loadable_path()` point to an unavailable library: the FTS-only profile fails during `_warm_sqlite()` with `sqlite3.OperationalError`, before FTS can be used. This also contradicts the updated configuration docs, which say FTS remains available without an embedding model. Please only register/load sqlite-vec when the vector index is enabled, keeping the FTS-only path independent of extension loading.

docs/en/development/memory-layer.md
```diff
@@ -90,20 +90,14 @@ the same citation fields through list and exact-read operations.
...
 capability.
 
-## Enable SQLite Vec1
+## Enable SQLite vector search
 
-SQLite vector search is enabled only when both a Vec1 0.7 or newer loadable extension and an embedding model are
-supplied. PowerContext does not install or build the native extension; provide a compatible library for the target
-operating system and architecture:
+SQLite vector search is enabled when an embedding model is supplied. The `powercontext[builtin]` extra bundles
+`sqlite-vec`, so no extension path or separate native-library installation is required:
```
Teingi, line 96: [P2] Update the RFC contracts for sqlite-vec

The runtime contract now replaces Vec1 and removes `vec1_extension`, but the English and Chinese copies of RFCs 0014, 0019, and 0020 still prescribe Vec1, its extension path, and `vec1(...)` table/query semantics. Readers now get incompatible setup and schema descriptions from the checked-in documentation. Please update both locale copies as part of this breaking change.

Teingi (changes requested):
The bundled sqlite-vec migration is clear overall, but the FTS-only startup regression and stale RFC contracts should be addressed before merging.

src/powercontext/builtin/persistence/sqlite/profile.py
PsiACE: Fixed in 80494f17. sqlite-vec is now loaded only when vector search is enabled, with FTS-only regression coverage.

docs/en/development/memory-layer.md
PsiACE: Added notes to RFCs 0014, 0019, and 0020 in both locales to mark the Vec1 sections as historical and no longer applicable.

Teingi (changes requested):
[P1] Remove the obsolete Vec1 setting from `.env.example`

`.env.example` still tells users to export `POWERCONTEXT_SERVER_DATABASE_VEC1_EXTENSION`, but this PR removes `vec1_extension` and `SQLiteConfig` forbids extra fields. Following the documented copy-and-source flow now makes `ServerSettings()` fail with:

`database.sqlite.vec1_extension: Extra inputs are not permitted`

This prevents the server from starting. Please remove the stale variable and Vec1 comment, and add coverage that the example can be loaded by `ServerSettings`.

Teingi (approved):
LGTM

PR #1309, opened 2026-08-21
docs: announce Pi Coding Agent support

pi-dal:
# Which issue or RFC does this PR close?

Related to #1259.

# Rationale for this change

Pi Coding Agent is now a supported PowerContext integration. The root READMEs need an immediately visible, accurate installation path so users can discover and start using it.

# What changes are included in this PR?

- Add a Pi Coding Agent quickstart to the English, Chinese, and Japanese READMEs.
- Explain how to install the CLI and native Pi package from the same PowerContext ref, using `master` as the current working example.
- Add Pi Coding Agent to every official-integration table and link to its detailed configuration guide.
- Clarify that Pi uses a native package rather than a host plugin.

# Are there any user-facing changes?

Yes. The READMEs now advertise Pi Coding Agent support and provide working installation and verification commands.

# How was this change tested?

- `mise exec -- make docs-test`
- `mise exec -- make check`

# AI usage statement

Built with Codex (GPT-5).

PR #1310, opened 2026-08-21
fix(ci): refresh lockfile package sources

pi-dal:

Related to #1309.

# Rationale for this change

The OceanBase acceptance workflow failed before running its scenarios because the frozen lockfile referenced a `pydantic-core` wheel on the Tsinghua PyPI mirror that returns HTTP 403.

- Refresh `uv.lock` against the official PyPI default index.
- Replace the mirror-specific registry and distribution URLs with official PyPI URLs.
- Preserve the resolved package versions and dependency graph.

No API or runtime behavior changes. Builds and installations no longer depend on the unavailable mirror URL recorded in the lockfile.

# How was this change tested?

- `mise exec -- make check`
- `mise exec -- uv run python -m pytest --doctest-modules -q` — 626 passed, 13 skipped
- Verified the previously failing Linux x86_64 `pydantic-core` wheel URL from the refreshed lockfile returns HTTP 200.

Docker is unavailable in the local environment, so the container acceptance build will be verified by CI.

# AI usage statement

Built with Codex (GPT-5).


pi-dal:
Context for review: this is a reliability hardening, not a deterministic behavioral or dependency-version regression.

The trigger was #1309: its OceanBase acceptance job ran in `westus3` and received HTTP 403 for the locked `pydantic-core` wheel from the Tsinghua PyPI mirror before scenarios began. A later #1311 job, still using the pre-change lockfile, freshly downloaded the same package successfully from `centralus`. The mirror is therefore region/IP-dependent.

This PR removes those hard mirror URLs in favor of official PyPI URLs so builds do not depend on that availability difference. All checks are green, including SQLite and OceanBase acceptance.

Evidence: failed job https://github.com/oceanbase/powercontext/actions/runs/32486415491/job/96784021364; successful later download https://github.com/oceanbase/powercontext/actions/runs/32507898779/job/96852286657.

Issue #1213, opened 2026-08-11
Tracking Issue: Agent Integrations for PowerContext


PsiACE:
## Goal

Track agent and framework integrations built on PowerContext's HTTP, MCP, Client, CLI, and plugin interfaces.

This issue replaces #922, #942, and #1055.

## Current support

- [x] HTTP API
- [x] MCP tools
- [x] Python Client
- [x] Codex plugin and CLI setup

## Agents

- [x] Codex
- [x] Claude Code
- [ ] Cursor
- [x] OpenCode
- [x] Pi
- [ ] Kimi Code
- [x] Hermes

## Frameworks

- [x] Pydantic AI
- [x] LangGraph

## Requirements

Each integration must define:

- [ ] Installation and authentication
- [ ] Scope mapping
- [ ] Recall and capture behavior
- [ ] Failure and recovery behavior
- [ ] Contract or end-to-end tests
- [ ] Setup and troubleshooting documentation
- [ ] Distribution, when applicable

Create or link a dedicated issue for each integration.

## Constraints

- Reuse existing PowerContext interfaces.
- Do not duplicate Runtime or Memory behavior in adapters.
- Do not depend on unsupported upstream extension mechanisms.
- Report files, permissions, storage, and rollback steps before an installer changes the user's environment.

thunguo:
Hi, if there are no other plans for this issue at the moment, could I handle the Frameworks part? I am relatively familiar with these two frameworks.

PsiACE:
@thunguo Thanks! Please open a separate issue for each integration with a brief proposal. We can discuss the details there, and I’ll assign them to you.

thunguo:
> [@thunguo](https://github.com/thunguo) Thanks! Please open a separate issue for each integration with a brief proposal. We can discuss the details there, and I’ll assign them to you.

Okay, I'll create issues for pydantic-ai and langgraph later.

curryxjh:
@PsiACE , if the OpenCode integration has not been claimed or assigned to anyone yet, please assign it to me. I will first create a dedicated issue for the OpenCode integration and submit a design proposal.

My initial plan is to start with a narrow MCP-first scope, reuse the existing PowerContext HTTP/MCP/CLI interfaces, and avoid duplicating Runtime or Memory logic in the adapter. If OpenCode's plugin hooks are stable enough later, I can consider automatic recall and prompt capture as a follow-up.

PsiACE:
> start with a narrow MCP-first scope

[...]

PR #1329, opened 2026-08-24
feat(opencode): add PowerContext integration

### LINKED ISSUE #1330: feat: OpenCode integration
### Feature description

Add an official OpenCode integration that gives each user turn automatic, project-scoped recall from PowerContext and exposes curated Memory and work-continuity tools.

The integration should align with the core behavior of the DeepSeek Harness plugin while using OpenCode's native plugin, tool, permission, and Skill surfaces. Runtime, Memory selection, ranking, citations, rendering, and persistence remain owned by the PowerContext Server.

Expected user flow:

```bash
powercontext setup opencode --source oceanbase/powercontext --ref master
powercontext server run
opencode
```

Installation can be checked independently from Server health:

```bash
powercontext doctor opencode
powercontext doctor
```

### Problem and proposed solution

PowerContext supports several coding-agent hosts, but OpenCode users currently need to call HTTP/MCP manually and cannot receive relevant project context automatically before a model request. MCP tools alone are insufficient because an agent cannot reliably search for context it does not yet know exists.

The proposed solution is a self-contained TypeScript package under `integrations/opencode/plugins/powercontext`, targeting OpenCode `>=1.18.21 <2`.

#### Automatic recall and prompt capture

Map one OpenCode turn to the existing HTTP contract:

```text
chat.message
  -> extract non-synthetic user text
  -> derive the project scope
  -> POST /v1/context/prepare
  -> validate powercontext.prepared-context.v1
  -> independently POST /v1/sources/content

experimental.chat.messages.transform
  -> inject the prepared value before model dispatch
  -> label it as untrusted historical evidence
  -> deduplicate it across tool-loop model steps
```

Recall, capture, and optional flush must fail independently. An unavailable or incompatible Server must never block normal OpenCode work. Prepared content is transient model context and must not be persisted into the OpenCode transcript.

#### Scope mapping

Resolve scope in this order:

1. `POWERCONTEXT_OPENCODE_SCOPE_ID`;
2. normalized `git:<host>/<path>` from `remote.origin.url`;
3. `local:<sha256>` of the resolved worktree path.

This keeps OpenCode compatible with project Memory created through Codex, Claude Code, DeepSeek Harness, and Pi.

#### Tools and permissions

Expose curated `pc_*` tools for:

- Memory search and lifecycle operations;
- bounded context preparation and explicit Source capture;
- Handoff preparation and continuation;
- Experience and Skill generation/read operations;
- read-only Artifact Candidate review.

Generate operation routing from `openapi/powercontext.yaml`. Durable mutations must use OpenCode's permission prompt. Candidate approval/rejection and destructive or administrative operations remain explicit human CLI or Dashboard actions.

#### Skill, configuration, and security

Install an OpenCode-native `project-context` Skill in the resolved global config directory. Setup may replace only an existing Skill carrying a PowerContext ownership manifest; a user-owned same-name path must remain untouched.

Use the `POWERCONTEXT_OPENCODE_` prefix for:

```text
BASE_URL
AUTHORIZATION
SCOPE_ID
CAPTURE_PROMPTS
FLUSH_ON_CAPTURE
REQUEST_TIMEOUT_MS
HTTP_BUDGET_MS
MAX_BYTES
FLUSH_MAX_CALLS
```

The integration must reject unsafe Server URLs, redirects, oversized or invalid responses, and secret-shaped captured or durable content. Authorization comes only from the environment and must not appear in metadata or diagnostics.

#### Installation and diagnosis

Add `powercontext setup opencode` and `powercontext doctor opencode`.

Setup validates the OpenCode version and checked-in bundle, registers the plugin globally, and installs the owned Skill with backup-and-replace recovery. Doctor reports the OpenCode executable, configured plugin, and installed Skill independently.

The plugin does not start or embed the Server and does not modify the PowerContext HTTP API or persisted formats.

### Alternatives considered

1. **Use MCP only.** This exposes model-initiated tools but cannot guarantee automatic recall before the first model request.
2. **Target the OpenCode v2 plugin API.** Its context hook is a closer semantic match, but v2 is currently beta and would narrow compatibility. The initial release uses the OpenCode 1.x plugin API.
3. **Persist recalled context through `chat.message`.** This is simpler but would pollute the transcript. The selected design caches recall per turn and injects it transiently through `experimental.chat.messages.transform`.
4. **Implement a DSH-style direct `/pc` command.** OpenCode 1.x server plugins cannot reliably return a local command result without invoking the model, so diagnosis and review remain CLI/Dashboard operations.
5. **Extract a shared JavaScript integration runtime first.** DSH, Pi, and OpenCode have different host contracts. Self-contained packages with shared contract tests avoid coupling installation to another integration's source tree.

### Additional context

- Tracking issue: #1213
- Implementation PR: #1329
- The OpenCode v1 pre-dispatch path depends on `experimental.chat.messages.transform`; compatibility must be protected by tests against the supported OpenCode version.

happy-v587:

Closes #1330.

Tracking issue: #1213.

# Rationale for this change

PowerContext supports several coding-agent hosts but did not have a native OpenCode integration. This adds project-scoped recall, prompt capture, and curated durable-context tools while keeping the PowerContext Server as the single runtime and storage boundary.

- Add a native OpenCode 1.x plugin with bounded automatic recall, independent prompt capture, fail-open HTTP behavior, Git-based scope mapping, secret filtering, and curated `pc_*` tools.
- Require confirmation for durable tool mutations while keeping Candidate approval and rejection as explicit human operations.
- Add `powercontext setup opencode` and `powercontext doctor opencode`, including OpenCode version validation and an ownership-protected global `project-context` Skill installation.
- Generate the OpenCode operation table from `openapi/powercontext.yaml` and add TypeScript, CLI, contract, packaging, and isolated real-OpenCode validation.
- Add English and Chinese setup documentation and navigation entries.

Yes. Users can install the integration with:

```bash
powercontext setup opencode --source oceanbase/powercontext --ref master
```

The plugin requires OpenCode 1.18.21 or newer in the 1.x line. Configuration uses the `POWERCONTEXT_OPENCODE_` environment prefix. This change does not modify the HTTP API, persisted formats, or existing integration behavior.

# How was this change tested?

- `make check`
- `make test` — 644 passed, 13 skipped
- `make contract-test` — 29 passed
- `make opencode-test` — 10 Vitest tests, TypeScript check, and bundle build
- `make docs-test`
- `make build`
- Isolated real OpenCode 1.18.21 installation with `powercontext setup opencode` followed by a successful three-check `powercontext doctor opencode`

# AI usage statement

OpenAI Codex (GPT-5) was used to inspect the repository and OpenCode plugin contract, implement the integration, write tests and documentation, resolve the latest-base conflicts, and run verification.

### DIFF
--- a/integrations/opencode/plugins/powercontext/src/index.ts
+++ b/integrations/opencode/plugins/powercontext/src/index.ts
@@ -50,12 +50,14 @@ type MessageBundle = {
 }
 
 type CachedTurn = { messageID: string; content?: string }
+type SessionContext = { cwd: string; scopeId: string }
 
 interface Runtime {
   client: PowerContextClient
   config: ResolvedConfig
-  cwd: string
-  resolveScope: () => Promise<string>
+  cacheSessionContext: (sessionID: string, cwd: string) => void
+  resolveSessionContext: (sessionID: string) => Promise<SessionContext>
+  sessionContexts: Map<string, Promise<SessionContext>>
   turns: Map<string, CachedTurn>
   log: (event: Record<string, unknown>) => Promise<void>
 }
@@ -128,7 +130,7 @@ async function flushThrough(runtime: Runtime, scopeId: string, position: number,
 
 async function capturePrompt(
   runtime: Runtime,
-  input: { scopeId: string; sessionID: string; messageID: string; prompt: string; signal: AbortSignal },
+  input: { cwd: string; scopeId: string; sessionID: string; messageID: string; prompt: string; signal: AbortSignal },
 ): Promise<void> {
   if (
     !runtime.config.capturePrompts
@@ -143,7 +145,7 @@ async function capturePrompt(
       metadata: {
         origin: 'opencode',
         event: 'user_prompt_submit',
-        cwd: runtime.cwd,
+        cwd: input.cwd,
         session_id: input.sessionID,
         message_id: input.messageID,
       },
@@ -164,11 +166,11 @@ async function prepareTurn(
   setTurn(runtime, input.sessionID, { messageID: input.messageID })
   const signal = createTimeoutSignal(runtime.config.httpBudgetMs)
   try {
-    const scopeId = await runtime.resolveScope()
+    const context = await runtime.resolveSessionContext(input.sessionID)
     let content: string | undefined
     try {
       const result = await runtime.client.request('prepare_context', {
-        scope_id: scopeId,
+        scope_id: context.scopeId,
         query: input.prompt,
         max_bytes: runtime.config.maxBytes,
       }, signal)
@@ -183,29 +185,54 @@ async function prepareTurn(
       await runtime.log({ event: 'context_prepare', outcome: 'failed' })
     }
     setTurn(runtime, input.sessionID, { messageID: input.messageID, content })
-    await capturePrompt(runtime, { ...input, scopeId, signal })
+    await capturePrompt(runtime, { ...input, ...context, signal })
   } catch {
     await runtime.log({ event: 'turn_prepare', outcome: 'failed' })
   }
 }
 
+async function sessionContextFromDirectory(cwd: string, config: ResolvedConfig): Promise<SessionContext> {
+  const directory = cwd.trim()
+  if (!directory) throw new Error('OpenCode session has no directory')
+  return { cwd: directory, scopeId: await deriveScopeId(directory, { configuredScopeId: config.scopeId }) }
+}
+
+async function loadSessionContext(input: PluginInput, config: ResolvedConfig, sessionID: string): Promise<SessionContext> {
+  const result = await input.client.session.get({ path: { id: sessionID } })
+  const cwd = result.data?.directory
+  if (!cwd) throw new Error(`OpenCode session ${sessionID} has no directory`)
+  return sessionContextFromDirectory(cwd, config)
+}
+
 function createRuntime(input: PluginInput, config: ResolvedConfig): Runtime {
-  const cwd = input.worktree?.trim() || input.directory
-  let scope: Promise<string> | undefined
+  const sessionContexts = new Map<string, Promise<SessionContext>>()
   return {
     config,
-    cwd,
     client: new PowerContextClient({
       baseUrl: config.baseUrl,
       authorization: config.authorization,
       requestTimeoutMs: config.requestTimeoutMs,
     }),
-    turns: new Map(),
-    resolveScope() {
-      scope ??= deriveScopeId(cwd, { configuredScopeId: config.scopeId })
-      void scope.catch(() => { scope = undefined })
-      return scope
+    sessionContexts,
+    cacheSessionContext(sessionID, cwd) {
+      const context = sessionContextFromDirectory(cwd, config)
+      sessionContexts.set(sessionID, context)
+      void context.catch(() => {
+        if (sessionContexts.get(sessionID) === context) sessionContexts.delete(sessionID)
+      })
     },
+    resolveSessionContext(sessionID) {
+      let context = sessionContexts.get(sessionID)
+      if (!context) {
+        context = loadSessionContext(input, config, sessionID)
+        sessionContexts.set(sessionID, context)
+        void context.catch(() => {
+          if (sessionContexts.get(sessionID) === context) sessionContexts.delete(sessionID)
+        })
+      }
+      return context
+    },
+    turns: new Map(),
     async log(event) {
       try {
         await input.client.app.log({
@@ -456,10 +483,21 @@ export const PowerContextPlugin: Plugin = async (input) => {
       output.system.push(GUIDANCE)
     },
     event: async ({ event }) => {
-      const value = event as unknown as { type?: string; properties?: { info?: { id?: string }; sessionID?: string } }
+      const value = event as unknown as {
+        type?: string
+        properties?: { info?: { id?: string; directory?: string }; sessionID?: string }
+      }
+      const info = value.properties?.info
+      if ((value.type === 'session.created' || value.type === 'session.updated') && info?.id && info.directory) {
+        runtime.cacheSessionContext(info.id, info.directory)
+        return
+      }
       if (value.type !== 'session.deleted') return
-      const sessionID = value.properties?.info?.id ?? value.properties?.sessionID
-      if (sessionID) runtime.turns.delete(sessionID)
+      const sessionID = info?.id ?? value.properties?.sessionID
+      if (sessionID) {
+        runtime.sessionContexts.delete(sessionID)
+        runtime.turns.delete(sessionID)
+      }
     },
   }
   await signalActivationProbe(runtime)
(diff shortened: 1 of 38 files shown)


happy-v587:
@PsiACE PTAL

src/powercontext/cli/opencode.py
```diff
@@ -0,0 +1,304 @@
...
+    try:
+        payload = json.loads((path / "package.json").read_text(encoding="utf-8"))
+    except (OSError, ValueError):
+        return False
+    return payload.get("name") == OPENCODE_PLUGIN_NAME
+
+
+def _materialize_remote_checkout(source: str, ref: str) -> Path:
+    target = checkout_target(ref)
+    if _is_opencode_plugin(target) or _is_opencode_plugin(target / OPENCODE_PLUGIN_RELATIVE):
```
Teingi: **[P1] Cache checkouts by source and resolved commit**

This cache hit never verifies `source` and never refreshes a mutable `ref`. Because `checkout_target()` is derived only from `ref`, installing `owner-a/repo@master` and then `owner-b/repo@master` silently reuses the first checkout; rerunning setup after the same `master` advances also keeps stale code. Please key the cache by normalized source plus ref/resolved commit and refresh mutable refs atomically.

src/powercontext/cli/opencode.py
```diff
@@ -0,0 +1,304 @@
...
+        actual = _version()
+    except SetupError as error:
+        return {
+            "opencode": Diagnostic(status=DiagnosticStatus.FAILED, detail=str(error)),
+            "plugin": Diagnostic(status=DiagnosticStatus.SKIPPED, detail="not checked because OpenCode is unavailable"),
+            "skill": Diagnostic(status=DiagnosticStatus.SKIPPED, detail="not checked because OpenCode is unavailable"),
+        }
+    try:
+        config_dir = opencode_config_dir()
+        configured = _configured_plugin(_run_opencode("debug", "config"))
```
Teingi: **[P1] Verify that the plugin actually activated**

`opencode debug config` still lists this file when `PowerContextPlugin` rejects its environment and returns `{}`. In a real OpenCode 1.18.21 run, setting `POWERCONTEXT_OPENCODE_BASE_URL=http://example.com` left the plugin with no hooks or tools, while `powercontext doctor opencode --json` reported all three checks as `OK`. Please probe activation (model-free) instead of treating the configured file path as sufficient.

integrations/opencode/plugins/powercontext/src/index.ts
```diff
@@ -0,0 +1,444 @@
...
+  cwd: string
+  resolveScope: () => Promise<string>
+  turns: Map<string, CachedTurn>
+  log: (event: Record<string, unknown>) => Promise<void>
+}
+
+function promptText(parts: readonly MessagePart[]): string {
+  return parts
+    .filter((part) => part.type === 'text' && !part.synthetic && typeof part.text === 'string')
+    .map((part) => part.text?.trim())
```
Teingi: **[P2] Normalize prompts from `opencode run`**

With OpenCode 1.18.21, the `opencode run "multi word prompt"` path reaches this hook with literal surrounding quotes. They are then used for both `prepare_context` and Source capture: in my E2E run, the same unquoted query produced a 683-byte ready context, while this path returned empty and stored the quoted prompt. Please normalize the `run` message representation before recall/capture and cover it with a real-host regression test.

src/powercontext/cli/opencode.py
happy-v587: Fixed in 257dd6c. Remote checkouts are now keyed by canonical GitHub repository identity, ref, and the resolved HEAD commit. Each setup clones into a staging directory, resolves the commit, and atomically promotes an immutable commit checkout; a failed refresh leaves the previous checkout intact. Regression tests cover source collisions, commit changes, and failed refresh cleanup. I also dogfooded oceanbase/powercontext@master and happy-v587/powercontext@master: they resolved to different SHAs and cache paths, while repeating the first source reused its resolved-commit path.

src/powercontext/cli/opencode.py
happy-v587: Fixed in 257dd6c. Doctor now creates a one-time nonce and asks model-free opencode debug config to activate the plugin. The plugin writes the nonce only after configuration, runtime, tools, and hooks are constructed; Doctor requires both the configured path and this activation handshake. In a real OpenCode 1.18.21 run, the valid configuration reports configured and active, while POWERCONTEXT_OPENCODE_BASE_URL=http://example.com reports did not activate and exits 1.

integrations/opencode/plugins/powercontext/src/index.ts
happy-v587: Fixed in 257dd6c. The message hook now decodes the single JSON-string layer emitted by opencode run before calling prepare_context or Source capture. Added a focused plugin regression test and a real-host E2E test that launches OpenCode 1.18.21 against a recording HTTP server; both the recall query and captured Source contain multi word prompt without the transport quotes.

integrations/opencode/plugins/powercontext/src/index.ts
```diff
@@ -0,0 +1,469 @@
...
+    .map((part) => normalizePromptPart(part.text!))
+    .filter((value): value is string => Boolean(value))
+    .join('\n\n')
+}
+
+function normalizePromptPart(value: string): string {
+  const text = value.trim()
+  if (!text.startsWith('"') || !text.endsWith('"')) return text
+  try {
+    const decoded: unknown = JSON.parse(text)
```
Teingi, line 78: **[P2] Preserve user-authored outer quotes**

This decodes every text part that happens to be a valid JSON string, not only the transport representation produced by `opencode run`. A direct message-hook input of `"preserve these quotes"` reaches both `prepare_context` and Source capture as `preserve these quotes`, changing the user's query and stored content. Please distinguish the run transport encoding from already-decoded chat text and add a regression test for intentional outer quotes.

integrations/opencode/plugins/powercontext/src/index.ts
```diff
@@ -0,0 +1,470 @@
...
+    config,
+    cwd,
+    client: new PowerContextClient({
+      baseUrl: config.baseUrl,
+      authorization: config.authorization,
+      requestTimeoutMs: config.requestTimeoutMs,
+    }),
+    turns: new Map(),
+    resolveScope() {
+      scope ??= deriveScopeId(cwd, { configuredScopeId: config.scopeId })
```
Teingi: **[P1] Resolve automatic scope from the session project directory**

`createRuntime` caches a single scope derived from the plugin startup cwd, so every `chat.message` handled by one headless `opencode serve` instance shares it. I reproduced this with two sessions created in different project directories: both automatic `prepare_context` and capture requests used the same scope and recorded `cwd: "/"`. This can mix recalled context and captured prompts across projects, and it also disagrees with explicit tools, which derive scope from `context.worktree || context.directory`.

Please use `sessionID` only to resolve and cache the session directory (through `client.session.get` and/or session lifecycle events), then derive the existing project-level scope from that directory and pass the same `{ cwd, scopeId }` through automatic recall and capture. Sessions in the same Git project will still share memory and handoffs, while different projects will be isolated. Keep the explicit configured scope as an override, clear the cache when the session is deleted, and fail open without falling back to the server startup cwd if session lookup fails.

Teingi (approved):
LGTM

PR #1335, opened 2026-08-24
fix(server): display uvicorn lifecycle logs without an error logger name

AlexStocks:

N/A — discovered during local usage; no tracking issue exists.

# Rationale for this change

uvicorn hardcodes `uvicorn.error` as its lifecycle logger name, which mostly
carries INFO startup lines but reads like an error channel. New users routinely
misread startup logs such as `INFO uvicorn.error Started server process` as an
error. The `Logger.name` attribute is read-only, so the fix rewrites the record
name in the display layer only.

- `src/powercontext/server/logging.py`: add `_UvicornDisplayNameFilter`,
  registered in the dictConfig and attached to the `uvicorn.error` logger
  (attached to the logger, not the shared handler, so powercontext's own
  loggers are unaffected). It rewrites `record.name` to `uvicorn` for display;
  the logger tree, level routing, and any filtering keyed on the original name
  are untouched.
- `tests/test_server_logging.py`: add 2 unit tests — the rewrite makes the JSON
  formatter emit `"logger":"uvicorn"`, and other logger names are left untouched.

Display-only: server startup/shutdown logs now render as `INFO uvicorn ...`
instead of `INFO uvicorn.error ...`. No API, config, or persisted format
changes. JSON-formatted logs emit `"logger":"uvicorn"` as well. Real errors
still render as `ERROR uvicorn ...`.

# How was this change tested?

- `uv run python -m pytest tests/test_server_logging.py`: 6 passed
- `uv run ruff check src/powercontext/server/logging.py tests/test_server_logging.py`: passed
- Manual: started the server and verified startup/shutdown lines render as
  `INFO uvicorn ...`; a simulated error still renders as `ERROR uvicorn ...`

# AI usage statement

An AI assistant (WorkBuddy) was used to implement the filter and draft this PR
description; the implementation was validated locally with tests and manual
runs.

### DIFF
--- a/src/powercontext/server/logging.py
+++ b/src/powercontext/server/logging.py
@@ -91,6 +91,22 @@ class _HumanContextFilter(OperationalContextFilter):
         return True
 
 
+class _UvicornDisplayNameFilter(logging.Filter):
+    """Rewrite uvicorn's lifecycle logger name for display.
+
+    uvicorn hardcodes "uvicorn.error" for its lifecycle logger, which mostly
+    carries INFO startup lines but reads like an error channel. Rewriting the
+    record name here is display-only: the logger tree, level routing, and any
+    filtering keyed on the original name are untouched.
+    """
+
+    @override
+    def filter(self, record: logging.LogRecord) -> bool:
+        if record.name == "uvicorn.error":
+            record.name = "uvicorn"
+        return True
+
+
 def configure_server_logging(config: ServerLoggingConfig) -> None:
     """Configure process logging for the foreground Server command."""
 
@@ -106,7 +122,10 @@ def configure_server_logging(config: ServerLoggingConfig) -> None:
     logging.config.dictConfig({
         "version": 1,
         "disable_existing_loggers": False,
-        "filters": {"operational": {"()": _HumanContextFilter}},
+        "filters": {
+            "operational": {"()": _HumanContextFilter},
+            "uvicorn_display_name": {"()": _UvicornDisplayNameFilter},
+        },
         "formatters": {"server": formatter},
         "handlers": {
             "server": {
@@ -126,6 +145,7 @@ def configure_server_logging(config: ServerLoggingConfig) -> None:
                 "handlers": ["server"],
                 "level": config.level,
                 "propagate": False,
+                "filters": ["uvicorn_display_name"],
             },
         },
     })
--- a/src/powercontext/server/logging.py
+++ b/src/powercontext/server/logging.py
@@ -92,12 +92,13 @@ class _HumanContextFilter(OperationalContextFilter):
 
 
 class _UvicornDisplayNameFilter(logging.Filter):
-    """Rewrite uvicorn's lifecycle logger name for display.
+    """Rewrite uvicorn's logger name for downstream display.
 
-    uvicorn hardcodes "uvicorn.error" for its lifecycle logger, which mostly
-    carries INFO startup lines but reads like an error channel. Rewriting the
-    record name here is display-only: the logger tree, level routing, and any
-    filtering keyed on the original name are untouched.
+    Uvicorn emits lifecycle and error records through ``uvicorn.error``.
+    Logger selection, hierarchy, and effective-level checks use that logger.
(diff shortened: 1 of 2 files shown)


tests/test_server_logging.py
```diff
@@ -21,10 +21,38 @@
...
 
+def test_uvicorn_error_records_are_displayed_as_uvicorn() -> None:
+    record = logging.makeLogRecord({
+        "name": "uvicorn.error",
+        "levelno": logging.INFO,
+        "levelname": "INFO",
+        "msg": "Started server process",
+    })
+
+    _UvicornDisplayNameFilter().filter(record)
```
Teingi: Could we exercise this through `configure_server_logging()` rather than calling the private filter directly? These tests would still pass if the filter were never registered or attached to `uvicorn.error`, so they do not protect the actual server behavior. Emitting a record via `logging.getLogger("uvicorn.error")` and asserting the console/JSON output would cover the wiring as well as the rename.

AlexStocks: [P1] 回归测试未覆盖真实 logging 接线

已在固定 Head `5f565411d7a169dd4e684fc4643cea7faeb54baf` 的隔离副本中做变异式反证：仅删除 `dictConfig` 中 filter 的注册与 `uvicorn.error` 挂载，保留 filter 类和这两个新增测试，测试仍然 `2 passed`；但通过 `configure_server_logging()` 后实际 emit `logging.getLogger("uvicorn.error")`，输出已退回 `INFO uvicorn.error ...`。因此当前测试无法检出本 PR 要修复的原始回归。建议按本线程所述，从 `configure_server_logging()` 驱动 console/JSON 两条真实路径并断言最终输出，而不是直接调用私有 filter。

src/powercontext/server/logging.py
```diff
@@ -91,6 +91,22 @@ def filter(self, record: logging.LogRecord) -> bool:
         return True
 
 
+class _UvicornDisplayNameFilter(logging.Filter):
+    """Rewrite uvicorn's lifecycle logger name for display.
+
+    uvicorn hardcodes "uvicorn.error" for its lifecycle logger, which mostly
+    carries INFO startup lines but reads like an error channel. Rewriting the
+    record name here is display-only: the logger tree, level routing, and any
```
AlexStocks: [P2] 收窄“原 logger 名过滤不受影响”的说明

这里会在 logger filter 阶段原地修改同一个 `LogRecord.name`。Python logging 随后才把该 record 交给 handlers，因此当前 `JsonFormatter`、handler filters，以及以后追加在该 filter 之后的 logger filters 都只能看到 `uvicorn`，无法再按 `uvicorn.error` 过滤；JSON 测试本身也证明了这一点。未变化的是 logger 查找、级别判断和改名前已经运行的过滤，不是“any filtering”。建议修正文档说明；如果设计目标确实要求保留原始分类，则应在 formatter 中使用单独的展示字段，而不要覆盖 `record.name`。

tests/test_server_logging.py
AlexStocks: 已在 054f5e9 修复。测试现在通过隔离子进程调用 configure_server_logging()，再使用真实的 logging.getLogger("uvicorn.error") 发出 INFO 和 ERROR 记录，分别精确断言 console 与 JSON 最终输出，同时验证 powercontext.server.factory 名称不变。相同测试在只移除 filter 注册和挂载的 mutation probe 中会因输出 uvicorn.error 而失败，在当前 Head 上通过。最终本地非 E2E 单元测试为 610 passed、1 skipped，Ruff 检查通过。

src/powercontext/server/logging.py
AlexStocks: 已在 054f5e9 修正说明。当前 docstring 明确：logger selection、hierarchy 和 effective-level checks 使用原始 uvicorn.error logger；filter 只改写 record.name；同一 logger 上排在其后的 filters、下游 handler filters 和 formatters 会看到 uvicorn；父 logger filters 不会在传播期间运行。运行时逻辑和 console/JSON 改名行为未改变。

Teingi (approved):
LGTM

Issue #1317, opened 2026-08-23
docs: README integration lists are inconsistent across languages and omit OpenClaw/Pi


AlexStocks:
## Summary

The three localized README files list the official integrations inconsistently, and the OpenClaw and Pi integrations — which are already shipped and wired into the CLI — are missing or miscounted.

## Actual state

The CLI already supports **six** hosts. `src/powercontext/cli/system.py` registers:

```
@setup_app.command("codex")
@setup_app.command("claude-code")
@setup_app.command("dsh")
@setup_app.command("openclaw")
@setup_app.command("pi")
@setup_app.command("hermes")
```

and `integrations/` contains `codex`, `claude-code`, `dsh`, `hermes`, `openclaw`, `pi`.

## What each README says

| File | Text | Table | Missing |
| --- | --- | --- | --- |
| `README.md` (EN) | "...for Codex, Claude Code, DeepSeek Harness, Hermes Agent, and Pi. All five integrations" | 4 cells (Codex, Claude Code, DeepSeek Harness, Hermes) | OpenClaw |
| `README_CN.md` | "为 Codex、Claude Code、DeepSeek Harness 和 Hermes Agent 提供官方插件" (four) | 4 cells | Pi, OpenClaw |
| `README_JP.md` | "Codex、Claude Code、DeepSeek Harness、Hermes Agent 向け" (four) | 4 cells | Pi, OpenClaw |

Notes:

- EN counts "five" integrations but renders four table cells and describes Pi only in a separate paragraph.
- CN and JP count "four" and omit both Pi and OpenClaw entirely.
- The Quick Start section in all three files only shows `powercontext setup` for codex / claude-code / dsh / hermes, even though `setup openclaw` and `setup pi` exist.

## Suggestion

Update all three READMEs to list the same six integrations (Codex, Claude Code, DeepSeek Harness, Hermes Agent, Pi, OpenClaw), include `setup openclaw` / `setup pi` in the Quick Start, and keep the "N integrations" sentence consistent with the rendered table.

XnLemon:
comment in https://github.com/oceanbase/powercontext/pull/1282#issuecomment-5382351516

OpenClaw的文档作为follow-up
如果可以的话我也可以把pi的补上
ask for assign

AlexStocks:
> comment in [#1282 (comment)](https://github.com/oceanbase/powercontext/pull/1282#issuecomment-5382351516)
> 
[...]

PR #1339, opened 2026-08-24
docs(integrations/openclaw): add configuration guides

XnLemon:

Part of #1317.

# Rationale for this change

OpenClaw is already supported by the CLI and integration code, but it lacks dedicated usage documentation. The integration was merged after `v0.0.2`, so the guide currently references `master` rather than that release tag.

- Add an English OpenClaw configuration how-to guide.
- Add a Chinese OpenClaw configuration how-to guide.
- Document plugin setup, Server startup, scope modes, memory tools, authentication, verification, and disabling the integration.
- Explain why `v0.0.2` cannot be used with `setup openclaw`.
- Recommend publishing a new tag containing the OpenClaw integration, after which the guide can use that tag instead of `master`.

Yes. This PR adds documentation for configuring the existing OpenClaw integration.

There are no runtime, API, persistence, or CLI behavior changes.

# How was this change tested?

- `make docs-test`
- `git diff --check`


XnLemon:
> OpenClaw is already supported by the CLI and integration code, but it lacks dedicated usage documentation. The integration was merged after v0.0.2, so the guide currently references master rather than that release tag.

Pls pay attention to this. no README.md change.

Teingi (commented):
I found a few user-facing documentation issues that should be addressed before this guide is published.

docs/en/docs/how-to/configure-openclaw.md
```diff
@@ -0,0 +1,92 @@
...
+Then begin a new OpenClaw session in another terminal:
+
+```bash
+openclaw tui
+```
+
+## Choose the memory scope
+
+The default `agent` scope isolates Memory by OpenClaw agent. Keep it unless agents must deliberately share project
+Memory. To use a trusted project identity when OpenClaw provides exactly one for the turn, reinstall with:
```
Teingi: The project scope still includes the OpenClaw agent ID (`openclaw:agent:<agent>:project:<hash>`), so two agents in the same project do not share Memory. This wording says project mode is for cross-agent sharing, which does not match the implemented isolation boundary. Please describe it as per-agent project partitioning, or change the scope contract and tests if cross-agent sharing is intended.

docs/en/docs/how-to/configure-openclaw.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+A local checkout works as well:
+
+```bash
+powercontext setup openclaw --source . --server-url http://127.0.0.1:8000
+```
+
+Setup builds and links the plugin, selects it as OpenClaw's memory slot, enables automatic recall and capture, adds its
+tools to `tools.alsoAllow`, and restarts the Gateway. Run setup again to refresh an existing installation. It does not
```
Teingi: Re-running this command does not refresh a remote mutable ref. `materialize_remote_checkout()` returns the existing valid checkout immediately, so a cached `master` is neither fetched nor recloned. Users can remain on an old plugin while believing it was updated. Please refresh mutable refs or document a reliable update procedure and cover the second-run case.

docs/en/docs/how-to/configure-openclaw.md
```diff
@@ -0,0 +1,92 @@
+---
+title: Configure OpenClaw
+description: Install and operate the PowerContext memory provider for OpenClaw.
+---
+
+# Configure OpenClaw
```
Teingi, line 6: This new guide is not added to the English or Chinese Integrations navigation in `zensical.toml`. The strict build generates both pages, but users cannot discover them through the site navigation. Please add both locale entries.

docs/en/docs/how-to/configure-openclaw.md
```diff
@@ -0,0 +1,92 @@
...
+  --scope-mode project
+```
+
+Group, channel, and incognito sessions are not captured or searched. Recall and capture fail open, so an unavailable
+PowerContext Server does not block an ordinary OpenClaw turn.
+
+## Use the memory tools
+
+The plugin registers `powercontext_memory_search` and `powercontext_memory_get` for reads. Explicit durable changes use
+`powercontext_memory_store`, `powercontext_memory_revise`, and `powercontext_memory_retire`. Automatic recall and source
```
Teingi, line 74: Automatic source capture does not make captured conversations searchable Memory by itself. The default Server has no generation model, and the plugin defers flush when `memory_extraction` is unavailable, so a fresh setup only stores pending Sources unless the user invokes an explicit store tool. Please document the generation-model prerequisite and add a capabilities or store/search verification step.

docs/en/docs/how-to/configure-openclaw.md
XnLemon: fixed

XnLemon:
Note: Conflict with #1340 
目前以不用tag的方式添加了项目总README
btw OpenCode 没日语ReadMe

XnLemon:
close bcz duplicate #1340

PR #1344, opened 2026-08-25
fix(transport): enforce a shared loopback-only plaintext HTTP policy (#1319)

### LINKED ISSUE #1319: bug: network transport policy permits unsafe remote HTTP configurations
### Describe the bug

The ready-to-run Server and the primary Python Client do not enforce the network safety policy already used by the Agent integrations and documented in the configuration guide.

Two unsafe configurations are accepted:

1. ServerSettings accepts a non-loopback bind such as 0.0.0.0 while bearer authentication remains disabled.
2. ClientSettings and PowerContextClient accept a non-loopback http URL and send a bearer token over that plaintext connection.

The Codex, Claude Code, and Pi integrations already reject non-loopback plaintext HTTP, so the same deployment currently has different security behavior depending on which client surface is used.

### Steps to reproduce

At commit a6e0dfe8b0b20b9a40e905aaa07c62427c1418f4:

#### Server probe

Construct:

~~~python
settings = ServerSettings(
    http={"host": "0.0.0.0"},
    auth={"enabled": False},
    mcp={"enabled": False},
    dashboard={"enabled": False},
)
~~~

Create the app and request /v1/capabilities without Authorization.

Observed:

~~~text
host=0.0.0.0
auth_enabled=False
capabilities_without_authorization=200
~~~

server/cli.py passes this host directly to uvicorn.run.

#### Client probe

Construct ClientSettings(server_url="http://memory.example"), then create PowerContextClient with token="probe-token" and a MockTransport.

Observed:

~~~text
accepted_url=http://memory.example
request_url=http://memory.example/health/live
authorization=Bearer probe-token
~~~

Relevant paths:

- src/powercontext/server/settings.py:40-44 and 129-153
- src/powercontext/server/cli.py:50-80
- src/powercontext/client/settings.py:38-49
- src/powercontext/client/client.py:181-194
- integrations/claude-code/plugins/powercontext/claude_code_settings.py:144-149
- integrations/pi/plugins/powercontext/src/config.ts:71-88
- docs/en/docs/reference/configuration.md:57-59

### Expected behavior

- Plain HTTP clients should be accepted only for 127.0.0.1, localhost, and ::1.
- A non-loopback Server bind should not silently start with authentication disabled.
- Deployments that intentionally use TLS termination or a controlled private network should require an explicit, clearly named opt-in rather than inheriting the unsafe combination by default.
- The Python Client, CLI, and Agent integrations should share one transport-policy contract and matching tests.

### Actual behavior

The Server accepts an unauthenticated public bind, and the Python Client sends bearer credentials to non-loopback HTTP URLs. Documentation warns against these configurations, but the primary runtime surfaces do not reject them.

### Environment

- Python: 3.14.3
- OS: Windows 11
- External network access was not required; both probes used in-process transports.

larry-zy:
## Background / Motivation

Closes #1319

PowerContext has several surfaces that open or configure HTTP connections
(the Python Client, the CLI, the Server, and the Agent plugins), but each one
implemented the "plaintext HTTP is only trusted on loopback" rule on its own,
with inconsistent behaviour that could be bypassed or misjudged. This PR
consolidates that rule into one shared implementation and fixes two bugs found
along the way.

## What changed

**1. Unified transport safety policy (`transport.py`)**
- Plaintext HTTP is trusted only on loopback addresses; bearer tokens are never
  sent over an unencrypted non-loopback connection; an unauthenticated Server
  must not bind to a routable address without an explicit opt-in.
- Client / CLI / Server now all route through the shared `is_loopback_host` /
  `is_plaintext_non_loopback` helpers.

**2. Fix loopback detection that only matched three named hosts**
- The old check matched only `127.0.0.1` / `::1` / `localhost`, so any other
  address in `127.0.0.0/8` (e.g. `127.0.0.2`) was wrongly classified as
  non-loopback and its plaintext requests were rejected.
- Detection now uses `ipaddress.ip_address(host).is_loopback`, covering the
  whole IPv4 loopback block.
- Both Agent plugins (Codex / Claude Code) ship isolated and cannot import
  `powercontext`, so each vendors its own copy of the check; new parametrised
  drift-guard tests pin both copies to the shared contract.

**3. Fix false positive of the token guard for caller-supplied transports**
- The "don't send a bearer token over non-loopback plaintext" guard also fired
  for a caller-supplied `http_client`, which broke the authenticated e2e flow
  that uses an in-process `httpx.ASGITransport` (`base_url=http://testserver`)
  with a token.
- When the caller supplies its own transport, transport security is the
  caller's responsibility (ASGI in-process, a Unix socket, or a TLS-terminating
  proxy all carry an `http://` label, so the scheme is no longer a reliable
  signal). The guard now applies only to the transport the client opens itself.

**4. Friendlier CLI error**
- When `server run` fails to load/validate `ServerSettings`, it now raises a
  `typer.BadParameter` instead of a raw traceback.

## Testing

- `uv run pytest`: 738 passed, 9 skipped
- `uv run ruff check`: clean
- Added/extended: loopback detection (incl. `127.0.0.2`, `127.1.2.3`),
  Client/Server policy, parametrised drift guards for both plugins, and a case
  allowing a token over a caller-supplied transport.

## Notes

- The Agent plugin `pi` (TypeScript) has the same `isLoopback` bug; it is left
  out of this PR and will be handled separately.
- bub / hermes / openclaw currently do not enforce loopback-only plaintext;
  that is existing design and out of scope here.


src/powercontext/client/client.py
```diff
@@ -187,6 +188,13 @@ def __init__(
         http_client: httpx.AsyncClient | None = None,
     ) -> None:
         self._base_url = base_url.rstrip("/")
+        # The plaintext guard only holds meaning for the transport this facade opens itself: there
+        # ``base_url``'s scheme accurately reflects what crosses the wire. A caller-supplied
+        # ``http_client`` owns its own transport (ASGI in-process, a Unix socket, a TLS-terminating
+        # proxy -- all of which carry an ``http://`` label), so the scheme is no longer a reliable
+        # signal and enforcing it would only produce false positives.
+        if token and http_client is None and is_plaintext_non_loopback(self._base_url):
```
Teingi: A caller-supplied http_client is not necessarily an in-process or otherwise secure transport. The LangGraph adapter passes a shared plain httpx.AsyncClient here for connection pooling, so this condition lets a token reach a non-loopback HTTP URL. I reproduced a request to http://memory.example carrying the Bearer header. Please keep the guard for shared clients and make any ASGI/Unix-socket bypass explicit instead of inferring safety from http_client being non-null.

src/powercontext/server/settings.py
```diff
@@ -171,6 +190,16 @@ def default_database_to_sqlite(cls, value: object) -> object:
             return value
         return {"kind": "sqlite", "url": _default_database().url, **value}
 
+    @model_validator(mode="after")
+    def reject_unauthenticated_non_loopback_bind(self) -> ServerSettings:
```
Teingi, line 212: This validator makes the documented Server container fail out of the box: docker/Dockerfile still exports POWERCONTEXT_SERVER_HTTP_HOST=0.0.0.0 while authentication defaults to disabled, and docker/README.md runs the image without authentication or this opt-in. The E2E Compose file now sets the opt-in, so the green harness masks the production-image regression. Please update the real container startup contract and cover the standalone documented invocation with a smoke test.

src/powercontext/server/cli.py
```diff
@@ -47,12 +52,26 @@ def run(
 ) -> None:
     """Run the ASGI service in the foreground."""
 
-    environment = ServerSettings()
+    try:
+        environment = ServerSettings()
```
Teingi: Because ServerSettings is validated before the CLI host override is applied, a safe command-line value cannot repair an unsafe environment value. For example, POWERCONTEXT_SERVER_HTTP_HOST=0.0.0.0 powercontext server run --host 127.0.0.1 fails here even though the final bind would be loopback-only. Please apply the CLI overrides before running the cross-field transport validation.

tests/test_transport.py
```diff
@@ -0,0 +1,196 @@
...
+_ALL_INTERFACES = "0.0.0.0"  # noqa: S104 - a non-loopback bind used to exercise the policy.
+
+# Both agent plugins ship isolated (they do not depend on powercontext) and each vendors its own
+# copy of the loopback policy. Load every copy by path and pin it to the shared contract so drift in
+# any one implementation is caught. Loading is defensive: a moved path or a plugin import error
+# skips that plugin's drift guard rather than failing collection for the whole module.
+_INTEGRATIONS = Path(__file__).resolve().parent.parent / "integrations"
+_VENDORED_PLUGIN_PATHS = {
+    "codex": _INTEGRATIONS / "codex" / "plugins" / "powercontext" / "settings.py",
+    "claude-code": _INTEGRATIONS / "claude-code" / "plugins" / "powercontext" / "claude_code_settings.py",
```
AlexStocks, line 53: [P1] 共享策略测试遗漏了 Pi 集成

#1319 明确把 `integrations/pi/plugins/powercontext/src/config.ts` 列为需要统一的 Agent surface，但这里仅枚举 Codex 和 Claude Code。当前 Pi 的 `isLoopback` 仍只接受 `localhost`、`127.0.0.1` 和 `[::1]`；我通过它的生产 `resolveConfig` 验证 `http://127.0.0.2:8000` 仍会报 `must use HTTPS outside loopback`，而本 PR 的共享实现和另外两个插件已经接受整个 `127.0.0.0/8`。因此新增 drift guard 无法发现已知的不一致，当前状态也不能满足“Agent integrations 共享同一 transport contract”。请同步修复 Pi，并让其 Vitest 消费同一组 loopback 测试向量。

src/powercontext/client/client.py
larry-zy: Fixed in fb4250a. The guard now stays on for caller-supplied transports; the only bypass is an explicit `trust_transport_security=True`. The LangGraph shared-client path (`open_client`) does not set it, so a token over `http://memory.example` now raises `ValueError` instead of being sent. Covered by tests/test_transport.py.

src/powercontext/server/settings.py
larry-zy: Fixed in fb4250a. docker/Dockerfile now declares `POWERCONTEXT_SERVER_ALLOW_UNAUTHENTICATED_NON_LOOPBACK=true` (with a comment explaining the network-namespace boundary the guard targets), and docker/README.md documents it. Added tests/test_docker_contract.py as the requested smoke test: it reads the real `POWERCONTEXT_SERVER_*` env straight from the Dockerfile and drives the documented `server run` invocation with no CLI overrides, so a Dockerfile that drops the opt-in fails here rather than being masked by the Compose harness.

src/powercontext/server/cli.py
larry-zy: Fixed in fb4250a. The CLI `--host` / `--port` overrides are now merged into the settings as a partial `http` mapping *before* `ServerSettings` is constructed, so the cross-field bind check runs against the address actually bound. `POWERCONTEXT_SERVER_HTTP_HOST=0.0.0.0 powercontext server run --host 127.0.0.1` now succeeds. See `run()` in src/powercontext/server/cli.py; covered by tests/test_cli.py.

tests/test_transport.py
larry-zy: 已修复（9d6fb2c）。`integrations/pi/plugins/powercontext/src/config.ts` 的 `isLoopback` 现在接受整个 `127.0.0.0/8`，`http://127.0.0.2:8000` 已通过。其 `transport-policy.spec.ts` 与 Python drift guard 消费同一份测试向量 `tests/fixtures/transport_loopback_vectors.json`，两侧 drift guard 均绿（Python 47 passed / Vitest 9 passed）。

src/powercontext/server/settings.py
AlexStocks: [P0] 不能把安全 opt-in 固化为镜像默认值

这样每个容器在 operator 做出选择之前就已经设置了 `ALLOW_UNAUTHENTICATED_NON_LOOPBACK=true`。README 的默认命令又使用 `--publish 8000:8000`，会把容器端口发布到宿主机接口；端口一旦发布，容器 network namespace 就不是未授权访问的安全边界。结果是镜像仍默认启动 `0.0.0.0 + auth=false`，本 PR 新增的 guard 对容器用户永远不会生效，正好保留了 #1319 要关闭的危险配置。请不要在 Dockerfile 内置 opt-in：本地示例应显式传入 opt-in 并把端口绑定到 `127.0.0.1`，网络部署则要求 authentication/TLS；对应 contract test 也应验证安全的发布方式，而不是只验证进程能启动。

src/powercontext/client/client.py
```diff
@@ -185,8 +186,21 @@ def __init__(
...
+        # transport itself, ``base_url``'s scheme accurately reflects what crosses the wire. A
+        # caller-supplied ``http_client`` *may* instead own a transport whose ``http://`` label is
+        # only a routing token -- an in-process ASGI app, a Unix socket, or a TLS-terminating proxy
+        # -- but a plain pooling ``httpx.AsyncClient`` (e.g. the shared client the LangGraph adapter
+        # installs) is exactly as exposed as one we would open ourselves. Supplying a transport is
+        # therefore not evidence of safety: the guard stays on for caller-supplied transports too,
+        # and a caller that knows its transport is secure must say so explicitly via
+        # ``trust_transport_security`` rather than have safety inferred from the argument being set.
+        transport_trusted = http_client is not None and trust_transport_security
+        if token and not transport_trusted and is_plaintext_non_loopback(self._base_url):
```
AlexStocks: [P1] 无 token 时仍会把业务内容发往远程明文 HTTP。这个条件被 `token` 短路了；公开 `capture_content_source` 探针确认 `PowerContextClient("http://memory.example")` 会把 Memory 内容发到 `/v1/sources/content`。Issue #1319 的验收和本 PR 新增文档都要求普通 Client 的明文 URL 仅限 loopback；没有 bearer token 不代表请求体不敏感。请让普通网络 transport 独立于 token 拒绝非 loopback HTTP，只为显式 `trust_transport_security=True` 的 ASGI/Unix-socket/TLS-proxy transport 保留绕过，并把当前允许该行为的测试反转为公共入口拒绝测试。

docs/en/docs/reference/configuration.md
```diff
@@ -55,8 +56,15 @@ Server settings use the `POWERCONTEXT_SERVER_` prefix.
 | `POWERCONTEXT_SERVER_EXTERNAL_SKILLS` | unset | JSON object containing the host identity and explicit Codex Skill roots |
 
 Static bearer authentication is disabled by default. When enabled, API and MCP requests must include
-`Authorization: Bearer <token>`; the liveness and readiness endpoints remain public. Plain HTTP should remain on a
-loopback address. Use TLS before exposing an authenticated Server over a network.
+`Authorization: Bearer <token>`; the liveness and readiness endpoints remain public. Plain HTTP is trusted only on a
+loopback address (`127.0.0.1`, `localhost`, `::1`). The Server refuses to start when it binds to a non-loopback address
+while authentication is disabled; either enable authentication, keep the bind on loopback, or, when TLS is terminated
+upstream or the network is otherwise controlled, set `POWERCONTEXT_SERVER_ALLOW_UNAUTHENTICATED_NON_LOOPBACK=true` to
```
AlexStocks: [P1] 同步修正仍推荐 `0.0.0.0` 的启动文档。新策略要求 authentication 或显式 opt-in，但 `docs/en/development/remote-access-implementation.md:25` 和对应中文文档仍直接运行 `server run --host 0.0.0.0 --port 8080`。Base/Head 公共 CLI 对照中，该命令从 exit 0、实际绑定 `0.0.0.0:8080` 变为当前 Head 的 exit 2；绿色 `check-docs` 只构建 Markdown，发现不了这个复制粘贴接口回归。请同步更新两份文档，给出 loopback、受控 opt-in 或 authentication/TLS 的完整命令，并增加命令级验证。

Teingi (approved):
LGTM

PR #1353, opened 2026-08-25
fix(opencode): install plugin through global autoload path

### LINKED ISSUE #1354: bug: OpenCode plugin setup reports configured but did not activate
### Describe the bug

The OpenCode integration setup can leave the PowerContext plugin listed in OpenCode's global configuration without actually activating the plugin. This makes a fresh installation appear configured while the OpenCode plugin lifecycle is not running.

### Steps to reproduce

1. Use macOS with OpenCode 1.18.23.
2. Install PowerContext from the GitHub master branch:

```bash
uv tool install --force "powercontext[cli,server] @ git+https://github.com/oceanbase/powercontext.git@master"
```

3. Run the OpenCode setup:

```bash
powercontext setup opencode \
  --source oceanbase/powercontext \
  --ref master
```

4. Run the integration diagnostic:

```bash
powercontext doctor opencode --json
```

### Expected behavior

The setup command should install the native OpenCode Server plugin and the `project-context` Skill. The diagnostic should confirm that the plugin is configured and active, and a new OpenCode session should load the plugin.

### Actual behavior

The plugin path can be written into OpenCode's global configuration, but activation fails with:

```text
PowerContext OpenCode plugin is configured but did not activate
```

The diagnostic reports:

```json
{
  "plugin": {
    "ok": false,
    "status": "failed",
    "detail": "PowerContext OpenCode plugin is configured but did not activate"
  }
}
```

The previous diagnostic relied on `opencode debug config`, which resolves configuration but does not prove that the plugin entered the real OpenCode lifecycle.

### Environment

- OS: macOS
- OpenCode: 1.18.23
- PowerContext source: `oceanbase/powercontext@master`
- Python: 3.12

### Proposed implementation

Install the built Server bundle into OpenCode's supported global local-plugin autoload directory, protect the destination with an ownership manifest, and verify activation through a model-free real `opencode run` lifecycle probe. This keeps the setup usable before a published npm package exists and does not require starting a PowerContext Server for the activation check.

happy-v587:

Closes #1354

# Rationale for this change

On OpenCode 1.18.23, the previous setup path could write the PowerContext local plugin directory into the global configuration without activating it. Users then received:

```text
PowerContext OpenCode plugin is configured but did not activate
```

The previous diagnostic used `opencode debug config`, which only resolves configuration and does not prove that the plugin entered the real OpenCode lifecycle.

- Install the built PowerContext Server bundle into OpenCode's global local-plugin autoload directory:
  `~/.config/opencode/plugins/powercontext-opencode.js`.
- Write an ownership manifest and refuse to overwrite an unrelated user plugin.
- Verify activation through a model-free real `opencode run` lifecycle probe.
- Keep installation independent of a published npm package.
- Add regression coverage for setup, activation probing, and owned plugin installation.

Usage after installation:

```bash
uv tool install --force "powercontext[cli,server] @ git+https://github.com/oceanbase/powercontext.git@master"

powercontext setup opencode \
  --source oceanbase/powercontext \
  --ref master

powercontext doctor opencode
powercontext server run
opencode
```

Yes. `powercontext setup opencode` now installs the native Server bundle through OpenCode's supported global local-plugin directory and verifies actual lifecycle activation. The setup remains fail-open when the PowerContext Server is unavailable.

This PR is limited to the OpenCode Server plugin installation bug. The separate TUI integration is being handled on another branch.

# How was this change tested?

- `make check`
- `make test`: 707 passed, 9 skipped
- Fresh isolated setup with OpenCode 1.18.23
- `powercontext doctor opencode --json`: `ok: true`
- Real model-free OpenCode plugin lifecycle activation probe

# AI usage statement

Implemented and validated with OpenAI Codex.

### DIFF
--- a/src/powercontext/cli/opencode.py
+++ b/src/powercontext/cli/opencode.py
@@ -21,13 +21,17 @@ import json
 import os
 import re
 import shutil
+import socket
 import subprocess
 import tempfile
+import time
 from contextlib import suppress
 from dataclasses import dataclass
 from pathlib import Path
 from shutil import which
+from urllib.error import URLError
 from urllib.parse import unquote, urlparse
+from urllib.request import urlopen
 from uuid import uuid4
 
 from powercontext.cli.git_source import InvalidGitHubSourceError, clone_github_source, github_clone_url
@@ -46,6 +50,7 @@ _VERSION = re.compile(r"^(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)")
 _COMMIT = re.compile(r"^[0-9a-f]{40,64}$")
 _ACTIVATION_PROBE_PATH = "POWERCONTEXT_OPENCODE_ACTIVATION_PROBE_PATH"
 _ACTIVATION_PROBE_NONCE = "POWERCONTEXT_OPENCODE_ACTIVATION_PROBE_NONCE"
+_ACTIVATION_PROBE_TIMEOUT = 15
 
 
 @dataclass(frozen=True, slots=True)
@@ -90,9 +95,10 @@ def install_opencode_plugin(*, source: str, ref: str) -> OpenCodeSetupResult:
     require_complete_plugin(plugin_dir)
     config_dir = opencode_config_dir()
     plugin_target = config_dir / "plugins" / f"{OPENCODE_PLUGIN_NAME}.js"
-    _install_plugin(plugin_dir / OPENCODE_BUNDLE, plugin_target)
     skill_target = config_dir / "skills" / "project-context"
+    require_replaceable_plugin(plugin_target)
     require_replaceable_skill(skill_target)
+    _install_plugin(plugin_dir / OPENCODE_BUNDLE, plugin_target)
     _install_skill(plugin_dir / OPENCODE_SKILL.parent, skill_target)
     return OpenCodeSetupResult(
         plugin=OPENCODE_PLUGIN_NAME,
@@ -160,6 +166,11 @@ def require_replaceable_skill(target: Path) -> None:
         raise SetupError.opencode_skill_conflict(target)
 
 
+def require_replaceable_plugin(target: Path) -> None:
+    if target.exists() and not _owned_plugin(target):
+        raise SetupError.opencode_plugin_conflict(target)
+
+
 def _install_skill(source: Path, target: Path) -> None:
     require_replaceable_skill(target)
     target.parent.mkdir(parents=True, exist_ok=True)
@@ -214,8 +225,7 @@ def _owned_plugin(path: Path) -> bool:
 
 
 def _install_plugin(source: Path, target: Path) -> None:
-    if target.exists() and not _owned_plugin(target):
-        raise SetupError.opencode_plugin_conflict(target)
+    require_replaceable_plugin(target)
     target.parent.mkdir(parents=True, exist_ok=True)
     staging = target.parent / f".{target.name}.tmp"
     manifest = _plugin_manifest_path(target)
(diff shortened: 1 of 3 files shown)


src/powercontext/cli/opencode.py
```diff
@@ -341,18 +378,56 @@ def _probe_plugin_activation() -> tuple[str, bool]:
...
-            "debug",
-            "config",
-            env={_ACTIVATION_PROBE_PATH: str(path), _ACTIVATION_PROBE_NONCE: token},
-        )
+        project = Path(directory) / "project"
+        project.mkdir()
+        output = _run_opencode("debug", "config")
+        command = [
+            opencode_executable(),
+            "run",
```
PsiACE: `opencode run` creates a persistent session before model validation. Each setup or doctor call leaves an empty session for a deleted temporary directory. Please avoid creating a session or remove it afterward.

src/powercontext/cli/opencode.py
```diff
@@ -88,9 +89,10 @@ def install_opencode_plugin(*, source: str, ref: str) -> OpenCodeSetupResult:
     plugin_dir = resolve_opencode_plugin_dir(source=source, ref=ref)
     require_complete_plugin(plugin_dir)
     config_dir = opencode_config_dir()
+    plugin_target = config_dir / "plugins" / f"{OPENCODE_PLUGIN_NAME}.js"
+    _install_plugin(plugin_dir / OPENCODE_BUNDLE, plugin_target)
```
PsiACE: The global plugin is installed before the Skill conflict check. If a user-owned Skill exists, setup fails but the plugin remains active. Please complete all conflict checks before writing files.

src/powercontext/cli/opencode.py
```diff
@@ -341,18 +378,56 @@ def _probe_plugin_activation() -> tuple[str, bool]:
...
+        project = Path(directory) / "project"
+        project.mkdir()
+        output = _run_opencode("debug", "config")
+        command = [
+            opencode_executable(),
+            "run",
+            "--dir",
+            str(project),
+            "--model",
+            "invalid/model",
```
PsiACE: `invalid/model` can be a valid custom provider and model. This probe may send real, billable requests. Please use a lifecycle probe that cannot invoke a model.

tests/test_opencode_cli.py
```diff
@@ -163,6 +167,36 @@ def test_setup_opencode_refuses_unowned_skill(tmp_path: Path, monkeypatch) -> No
...
+def test_activation_probe_uses_headless_server_without_model(tmp_path: Path, monkeypatch) -> None:
+    import powercontext.cli.opencode as opencode_cli
+
+    plugin = _write_plugin(tmp_path / "checkout")
+    config = tmp_path / "config"
+    observed: list[list[str]] = []
+
+    monkeypatch.setattr(opencode_cli, "which", lambda _name: "/usr/bin/opencode")
+    monkeypatch.setattr(opencode_cli, "_run_opencode", _fake_opencode(plugin, config))
+
```
AlexStocks, line 255: [P1] 这条回归测试没有执行真正的生命周期探针。这里把 `_run_opencode_probe` 整体替换成直接写 marker，所以即使生产实现完全不再启动 OpenCode、不请求 `/session`、也不等待或清理子进程，测试仍然通过；我在隔离副本中把该函数变异成空实现后，这个用例仍是 1 passed。当前修复的核心正是 `debug config` 不会激活插件，建议用临时可执行程序/本地测试服务跑过真实 `_run_opencode_probe`，至少断言它发起激活请求、等待正确 nonce，并在成功、退出和超时路径回收进程。

happy-v587: @AlexStocks Addressed in c3116ff. The regression coverage now runs the real `_run_opencode_probe` against a temporary local HTTP server process. It verifies that `/session` is requested, an incorrect nonce is ignored until the expected nonce arrives, and the subprocess is cleaned up on success, early exit, and timeout. Validation: `make check`; full test suite with OpenCode hidden: `710 passed, 10 skipped`; OpenCode 1.18.23 doctor and activation smoke tests also passed. Please take another look.

Teingi (approved):
LGTM

PR #1343, opened 2026-08-25
feat(hermes): hermes coding agent capabailities completely alignment

Alanxtl:
# Which issue or RFC does this PR close?

#1338 hermes part

# Rationale for this change

This change integrates PowerContext with Hermes Agent as a standard Memory Provider.

It provides Hermes with scoped external memory, memory capture and retrieval, cross-session continuity, handoff workflows, evaluation tracing, and explicit memory management commands while keeping PowerContext independent from Hermes' core runtime.

alignment most of the capabailities stated in the #1338

# What changes are included in this PR?

- Add a PowerContext Memory Provider for Hermes Agent v0.20.4+.
- Add Hermes setup schema and configuration persistence.
- Add the `powercontext setup hermes` and `powercontext doctor hermes` workflows.
- Add the standalone `powercontext-command` companion plugin.
- Register `/pc` and `/powercontext` as slash-command aliases.
- Add slash-command autocomplete for first-level PowerContext commands.
- Support memory search, listing, changes, retrieval, remember, revise, retire, and flush operations.
- Support Work Contracts, Handoff, Task Outcome, Experience, Skill, External Skill, and Artifact Candidate operations.
- Add per-session Evaluation Trace persistence.
- Add Git-private Workstream scope persistence.
- Add secret filtering and overlap-aware pre-compression capture.
- Add stable memory-entry mapping for Hermes memory updates and removals.
- Refactor the integration into separate provider, command, trace, workstream, helper, and operation modules.
- Update Hermes integration documentation and tests.

# Are there any user-facing changes?

Yes.

Users can configure PowerContext through Hermes memory setup and use:

```text
/pc
/powercontext
```

Typing `/pc ` or `/powercontext ` displays the available first-level commands.

# How was this change tested?

- `make check`
- `uv run pytest tests/integrations/test_hermes_provider.py tests/test_hermes_cli.py -q`
- Verified Hermes v0.20.4 plugin discovery and registration.
- Verified `/pc` and `/powercontext` aliases.
- Verified first-level slash-command completion for both aliases.
- Tested provider setup and doctor workflows.
- Tested memory capture, retrieval, revision, retirement, pre-compression filtering, Workstream binding, and Evaluation Trace behavior.

# AI usage statement

OpenAI Codex was used to assist with implementation, refactoring, documentation, and test validation.

PR #1360, opened 2026-08-26
feat: add host-visible integration diagnostics

### LINKED ISSUE #1298: Tracking Issue: local Server availability and service installation

Alanxtl:

This PR implements the first step of #1298. It does not close #1298.

This PR is downstream of #1343 and should be reviewed after #1343 is merged.

# Rationale for this change

When PowerContext is unavailable or returns an incompatible response, integrations currently provide limited or inconsistent feedback. Users need a bounded, fail-open diagnostic that explains the failure and points to `powercontext doctor` without exposing request details or disrupting the host.

- Add common failure classification:
  - `authentication_failed`
  - `version_mismatch`
  - `server_unavailable`
  - `invalid_response`
- Add host-visible diagnostics for:

| Host | Diagnostic channel | Component |
| --- | --- | --- |
| Codex | Hook `stderr` | `powercontext.codex.recall` |
| Claude Code | Hook `stderr` | `powercontext.claude_code.recall` |
| DSH | Host logger warning | `powercontext.dsh` |
| OpenClaw | Plugin logger warning | `powercontext.openclaw` |
| Pi | Host terminal warning | `powercontext.pi` |
| Hermes | Python host logger warning | `powercontext.hermes` |

- Use each host’s supported logging/output channel.
- Add 60-second deduplication to avoid repeated warnings.
- Keep integrations fail-open and content-free.
- Preserve generic tool errors for model-facing responses.
- Add the `powercontext doctor` recovery hint for unavailable servers.
- Update troubleshooting documentation and integration tests.
- Bub is out of scope for this PR.

Yes.

Users now receive a bounded diagnostic when PowerContext is unavailable or returns a classified failure. Normal host execution continues, and no URL, token, response body, or user content is included in the diagnostic.

There are no breaking API or persistence-format changes.

# How was this change tested?

- `uv run pytest tests/integrations/test_hermes_provider.py -q` — 36 passed
- `uv run ruff check integrations/hermes/plugins/powercontext/commands.py tests/integrations/test_hermes_provider.py` — passed
- Codex and Claude Code diagnostic tests passed
- Pi and OpenClaw lifecycle diagnostic tests passed
- DSH bundled output passed `node --check`
- Manual validation with an unreachable PowerContext endpoint confirmed fail-open behavior and generic tool errors.
- Verified diagnostic deduplication and `powercontext doctor` recovery metadata.

# AI usage statement

Implemented and reviewed with OpenAI Codex


integrations/codex/plugins/powercontext/hooks/recall.py
```diff
@@ -421,8 +449,47 @@ def _emit_context_event(
         event["context_status"] = context_status
     if content_bytes is not None:
         event["content_bytes"] = content_bytes
+    if recovery is not None:
+        event["recovery"] = recovery
     sys.stderr.write(json.dumps(event, separators=(",", ":")) + "\n")
```
Teingi: [P1] This hook still exits with status 0 and writes the diagnostic only to stderr. Codex parses stdout and ignores stderr on a successful `UserPromptSubmit`; Claude Code uses the same success-path contract, and the mirrored hook has the same issue. As a result, the advertised host-visible warning remains silent. Please return the content-free event through the top-level `systemMessage` in the successful hook JSON (alongside `hookSpecificOutput` when context is present) and cover both adapters with host-level fixtures.

integrations/codex/plugins/powercontext/hooks/recall.py
```diff
@@ -105,15 +106,22 @@ def main(settings: CodexPluginSettings | None = None) -> int:
             payload = cast(dict[str, Any], json.load(stdin))
         if not _is_user_prompt_submit(payload.get("hook_event_name")):
             return 0
+        emitted_diagnostics: set[str] = set()
```
Teingi, line 152: [P2] This set is recreated for every hook invocation, and the host starts a new Python process for each prompt, so it cannot implement the claimed 60-second throttle. Two immediate invocations emit two identical warnings. Please keep the cooldown state outside the short-lived invocation, apply the same fix to the Claude hook, and test across separate invocations.

integrations/hermes/plugins/powercontext/provider.py
```diff
@@ -0,0 +1,1053 @@
...
+                outcome = "authentication_failed"
+            elif status == 404:
+                outcome = "version_mismatch"
+            elif status == 503:
+                outcome = "server_unavailable"
+            else:
+                outcome = "invalid_response"
+        elif isinstance(error, PowerContextTransportError):
+            status = None
+            outcome = "server_unavailable"
```
Teingi: [P2] `PowerContextTransportError` also represents oversized, invalid-JSON, and non-object responses in `client.py`. Mapping the whole class to `server_unavailable` therefore reports an HTTP 200 malformed response as an outage and points users to the wrong recovery. Please introduce or map a distinct invalid-response error to `invalid_response`.

integrations/codex/plugins/powercontext/hooks/recall.py
```diff
@@ -421,8 +449,47 @@ def _emit_context_event(
...
+        else:
+            outcome = "invalid_response"
+        _emit_context_event(
+            outcome,
+            event_name=event_name,
+            http_status=error.status,
+            recovery="powercontext doctor" if outcome == "server_unavailable" else None,
+            emitted_diagnostics=emitted_diagnostics,
+        )
+    elif isinstance(error, _ServerUnavailableError):
```
Teingi: [P2] A raw `TimeoutError` can escape `_post_json` when the shared deadline expires before its `try` block. It falls through here to `invalid_response`, although the contract maps timeouts to `server_unavailable`. Please handle `TimeoutError` with `_ServerUnavailableError`, as the Claude hook does.

tests/codex_plugin/test_recall.py
```diff
@@ -179,6 +179,7 @@ def test_recall_failure_is_non_blocking(
         "component": "powercontext.codex.recall",
         "event": "context_prepare",
         "outcome": "server_unavailable",
+        "recovery": "powercontext doctor",
```
Teingi, line 184: [P2] This assertion is environment-dependent: the test mocks recall but leaves prompt capture enabled, so `main()` can contact the default local Server and append a second, different diagnostic. I reproduced `JSONDecodeError: Extra data` with a live endpoint returning 404. Please mock `_capture_prompt` or disable capture here, and test capture diagnostics separately.

integrations/codex/plugins/powercontext/hooks/recall.py
Alanxtl: Fixed. Diagnostic cooldown state is now persisted outside the hook process with file locking, so the 60-second throttle survives separate invocations. Added coverage for repeated invocations.

integrations/codex/plugins/powercontext/hooks/recall.py
Alanxtl: Fixed. Both Codex and Claude Code hooks now return content-free diagnostics through the top-level systemMessage on stdout, while preserving hookSpecificOutput when context is available. Added host-level fixtures for both adapters.

integrations/hermes/plugins/powercontext/provider.py
Alanxtl: Fixed. Added PowerContextInvalidResponseError for oversized, malformed JSON, and non-object responses. These now report invalid_response, while transport failures remain server_unavailable.

integrations/codex/plugins/powercontext/hooks/recall.py
Alanxtl: Fixed. _post_json now normalizes deadline-expiry TimeoutError to _ServerUnavailableError, and the recall/failure paths handle timeouts as server_unavailable.

tests/codex_plugin/test_recall.py
Alanxtl: Fixed. The failure test now mocks _capture_prompt to avoid contacting a live Server. Capture diagnostics are covered separately.

integrations/codex/plugins/powercontext/hooks/recall.py
```diff
@@ -83,15 +85,36 @@ def redirect_request(
...
+
+
+def _http_failure_outcome(error: _HttpStatusError) -> str | None:
+    if error.status == 401:
+        return "authentication_failed"
+    if error.status == 404 and error.path in _COMPATIBILITY_OR_AVAILABILITY_PATHS:
+        return "version_mismatch"
+    if error.status == 503:
+        return "server_unavailable"
+    if error.status in {404, 409, 422}:
```
Teingi, line 120: [P1] Suppressing 404/409/422 by status alone makes automatic integration failures silent. I reproduced `422 /v1/context/prepare` returning no context and no diagnostic; capture and flush have the same issue for their real endpoints. The new capture test uses `/v1/memory/entries/get` even though `_capture_prompt` calls `/v1/sources/content`, so it misses this path. Please classify by operation, actual endpoint, and structured error code, and keep automatic prepare/capture/flush failures visible.

integrations/codex/plugins/powercontext/hooks/diagnostics.py
```diff
@@ -0,0 +1,103 @@
...
+            lock_file.seek(0, os.SEEK_END)
+            if lock_file.tell() == 0:
+                lock_file.write(b"\0")
+                lock_file.flush()
+            lock_file.seek(0)
+            msvcrt.locking(lock_file.fileno(), msvcrt.LK_LOCK, 1)
+        else:
+            import fcntl
+
+            fcntl.flock(lock_file.fileno(), fcntl.LOCK_EX)
```
Teingi: [P2] This lock is unbounded on the hook critical path (`LOCK_EX` / `LK_LOCK`). With another process holding the file lock, I reproduced `should_emit()` blocking for 6.5 seconds, so a diagnostic cooldown can stall an otherwise fail-open host invocation. Please use non-blocking locking or a very small bounded retry, fall back to emit or skip on contention, and add a subprocess contention regression test.

tests/codex_plugin/test_recall.py
```diff
@@ -173,13 +174,87 @@ def test_recall_failure_is_non_blocking(
...
+        "component": "powercontext.codex.recall",
+        "event": "capture_source",
+        "outcome": "server_unavailable",
+        "http_status": 503,
+        "recovery": "powercontext doctor",
+    }
+    assert errors.getvalue() == ""
+
+
+def test_host_diagnostic_is_throttled_across_hook_invocations(
```
Teingi, line 230: [P2] This is still a direct Python `main()` call, not a host-level fixture. It proves the JSON shape but not that Codex actually renders `systemMessage`; the Claude test has the same gap. RFC 1299 requires an acceptance test or recorded host fixture showing that the selected channel is user-visible. Please exercise the installed hook through each real host, or check in a recorded host fixture from that boundary.

Alanxtl: image
fixed and tested in codex

integrations/codex/plugins/powercontext/hooks/diagnostics.py
Alanxtl: add test

integrations/codex/plugins/powercontext/hooks/recall.py
Alanxtl: fixed

Teingi: The Codex and Claude paths are fixed, but this remains open for DSH, Pi, OpenClaw, and Hermes. Their classifiers still return `undefined` or `None` for 404/409/422 before considering the automatic operation and real endpoint. On the current head I reproduced silent `422 context_prepare`, `422 capture`, and `409 flush` failures in DSH and OpenClaw, and the same three classifications returning `None` in Hermes; Pi has the same implementation. Please carry the operation/path/error-code classification and real-endpoint regression tests across those adapters.

integrations/codex/plugins/powercontext/hooks/recall.py
```diff
@@ -261,15 +341,22 @@ def _post_json(
...
         with _URL_OPENER.open(request, timeout=request_timeout) as response:
             if expected_status is not None and response.status != expected_status:
-                raise _HttpStatusError(response.status)
+                code = _decode_error_code(_read_response(response, deadline=request_deadline))
+                raise _HttpStatusError(response.status, path, code)
             result = json.loads(_read_response(response, deadline=request_deadline))
     except HTTPError as error:
-        raise _HttpStatusError(error.code) from error
+        try:
+            error_body = error.read(_MAX_RESPONSE_BYTES + 1)
```
Teingi: [P2] Keep error-body parsing under the shared wall-clock deadline. This direct `error.read(...)` bypasses `_read_response(..., deadline=request_deadline)`, so a peer that trickles one byte within each socket timeout can exceed the hook budget indefinitely. On the current head, a 422 response sending one byte every 20 ms took 0.714 s despite a 0.1 s deadline; Claude has the same branch. Please use a deadline- and size-bounded error-body reader and add a slow-4xx regression test.

Alanxtl: fixed

integrations/codex/plugins/powercontext/hooks/recall.py
Alanxtl: done

Teingi (approved):
LGTM

PR #1367, opened 2026-08-26
feat(cli): add interactive PowerContext config init

### LINKED ISSUE #1333: docs: add a full-capability runtime setup guide
## Feature description

新增一份完整能力运行指南，说明如何在不改变默认最小启动行为的前提下，配置并启动一个功能完整的 PowerContext Server。

文档应覆盖 Generation、Embedding、Scheduler、SQLite Vec1、Dashboard、Scope ID、数据库路径和宿主集成之间的关系，并提供一套可复制的配置与验证流程。

当前的 `powercontext server run` 可以在没有推理服务和向量扩展的情况下启动，因此适合最小体验，但很多能力默认处于关闭或降级状态。用户需要自行从安装、配置、推理、向量索引和集成文档中拼接出一套完整运行方式，容易遇到以下问题：

- 未配置 Generation model，Source 无法自动提取为 Memory；
- 未配置 Embedding model、Profile ID 和 dimension，无法使用 Embedding；
- Vec1 路径不存在或版本不兼容时，Server 会在启动阶段失败；
- 数据库路径被示例配置覆盖后，可能误连到空数据库；
- Server scope、Dashboard scope 和 Codex/OpenCode 等宿主 scope 不一致时，看不到预期数据；
- Scheduler、Dashboard 和 MCP 的启用状态不容易确认。

建议新增一篇 how-to 文档，并提供一份不包含密钥、明确标注可选 Vec1 路径和模型占位符的完整配置示例。文档应将运行方式命名为“完整能力运行”或 `full-capability runtime`，而不是修改默认启动命令或引入 `--profile full`。

## Alternatives considered

- 继续让用户分别阅读安装、配置、推理和集成文档：信息已存在，但缺少端到端路径；
- 修改默认 `powercontext server run`，自动开启所有能力：会引入模型费用、动态库和外部服务依赖，不适合作为默认行为；
- 增加 `--profile full` 或 `make server-full`：可以后续讨论，但本 Issue 首先聚焦文档和规范配置示例。

## Additional context

建议产出：

- `docs/zh/docs/how-to/full-capability-runtime.md`；
- `docs/en/docs/how-to/full-capability-runtime.md`；
- 在 `zensical.toml` 的 Get started / 开始使用导航中加入入口；
- 必要时调整 `.env.example`，避免把不存在的 Vec1 动态库或项目内空数据库作为默认可复制配置。

文档至少应包含：

1. 默认最小运行与完整能力运行的能力矩阵；
2. Generation、Embedding、数据库、Vec1、Scheduler 和 Dashboard 环境变量分组；
3. Scope ID 的来源、显式覆盖方式和跨宿主一致性要求；
4. 从加载环境变量、启动 Server 到 `doctor`、`ready`、`capabilities`、Memory flush 的完整命令；
5. 预期输出、数据位置、停止方式和常见故障排查；
6. 明确说明没有 Vec1 时仍可使用 SQLite FTS，但不会启用 vector/hybrid 搜索；
7. 不把 API Key、连接凭据或其他敏感信息写入文档示例或 Memory。

验收标准：读者只依赖该指南和项目仓库中的示例，就能启动一套配置明确、可验证、不会误连空库的完整能力运行环境，并能判断每项能力是否真正 ready。

happy-v587:
Closes #1333

# Rationale for this change

The existing server can run in a minimal mode, but users had to manually combine inference, vector search, scheduler, Dashboard, Scope ID, database, and Coding Agent settings. This change provides one guided command for a complete local setup without changing the default server behavior.

- Add `powercontext config init` with a short interactive setup for Scope ID, API protocol, Base URL, API key, Generation model, Embedding model, and dimension.
- Derive internal provider adapters and environment variable names from stable API protocol choices.
- Generate a protected `0600` `.env` containing full-capability Server, Dashboard, Scope, inference, database, scheduler, and integration settings.
- Print setup and launch commands for Codex, Claude Code, DeepSeek Harness, OpenCode, and Pi using the same Scope ID.
- Keep `config show` and `config validate`, update the environment example, and add the bilingual full-capability Quick Start guide.
- Preserve the default minimal `powercontext server run` behavior.

Yes. Users can run `powercontext config init --output .env` to create a complete local configuration. The generated file is local and ignored by Git; API keys are hidden during interactive input and redacted by `config show`. This introduces the new `config init` workflow and does not alter the default server startup contract.

# How was this change tested?

- `uv run pytest -q tests/test_config_cli.py tests/test_cli.py tests/test_server.py` — 60 passed.
- `make check` — passed.
- `make docs-test` — passed.
- `make build` — passed.
- Manual CLI matrix: default, Advanced, OpenAI Responses, Anthropic with independent Embedding, Custom adapter, overwrite protection, and real TTY fuzzy selection.
- Manual runtime validation with a local OpenAI-compatible stub: Generation and Embedding requests were received, `ready` reported ready, Source ingestion returned HTTP 202, and `flush` returned HTTP 200 with the Source processed.
- Isolated setup validation: Codex, Claude Code, DeepSeek Harness, and Pi setup completed successfully. OpenCode setup and the existing OpenCode Host E2E test exceeded the timeout on the locally installed OpenCode 1.18.23 and returned `Unexpected server error`; this remains an environment-specific limitation and is not caused by `config init`.
- Full regression: 714 passed, 9 skipped, 1 failed at `tests/e2e/test_opencode_plugin_host.py`.

# AI usage statement

Built and validated with OpenAI Codex (GPT-5). The agent inspected the repository, implemented the CLI and documentation changes, ran automated and manual tests, and recorded the OpenCode environment limitation above.


src/powercontext/cli/config.py
```diff
@@ -0,0 +1,863 @@
...
+    env_file: Annotated[Path, typer.Option(help="Environment file to inspect.")] = Path(".env"),
+) -> None:
+    """Print effective assignments with credentials redacted."""
+
+    try:
+        values = read_environment_file(env_file)
+    except (EnvironmentFileError, OSError) as error:
+        _fail(str(error))
+    for name in sorted(values):
+        value = "<redacted>" if _is_secret_name(name) else values[name]
```
Teingi: `config show` promises redacted credentials, but this name-based check still prints standard credential containers such as `OTEL_EXPORTER_OTLP_HEADERS=authorization=Bearer-demo-secret` verbatim, and it does the same for custom credential names. That makes the output unsafe to share. Could we redact supported header/container variables and credential names recorded by the generator, with a regression test for OTLP headers?

src/powercontext/cli/env_file.py
```diff
@@ -0,0 +1,123 @@
...
+
+    environment: dict[str, str] = {}
+    for line_number, line in enumerate(content.splitlines(), start=1):
+        stripped = line.strip()
+        if not stripped or stripped.startswith("#"):
+            continue
+        if stripped.startswith("export "):
+            stripped = stripped.removeprefix("export ").lstrip()
+        try:
+            tokens = shlex.split(stripped, comments=True, posix=True)
```
Teingi: `shlex.split(..., comments=True)` treats every `#` as a comment start, so `TOKEN=abc#123` is loaded as `abc`. A shell assignment preserves the full value, which means `server run --env-file` can silently use a truncated API key or bearer token. Could we recognize `#` only at a real comment boundary and cover this case in the parser tests?

docs/en/docs/how-to/full-capability-runtime.md
```diff
@@ -0,0 +1,102 @@
...
+
+#### 4. Verify the Server
+
+Run this in a second terminal:
+
+```bash
+set -a
+. ./.env
+set +a
+powercontext ready
```
Teingi, line 94: This stops verification at `ready`/`capabilities`, but #1333's acceptance criteria also require `doctor`, a Source-to-flush-to-Memory round trip, data-location and shutdown guidance, a minimal/full capability matrix, and the FTS-only fallback when vector search is unavailable. Without the round trip, readers cannot verify that scheduling and extraction actually work. Please complete the flow here and mirror it in the Chinese guide.

Teingi (commented):
I found three issues in the new configuration flow.

src/powercontext/cli/config.py
happy-v587: Fixed. Added `_CREDENTIAL_CONTAINER_SUFFIXES` (`_HEADERS`, `_HEADER`, `_COOKIES`, `_COOKIE`) to `_is_secret_name`, and extended the generator to record credential names via `# credentials=...` metadata in the managed block. `config show` now reads that metadata and redacts recorded names in addition to pattern-matched ones. Added regression tests for both OTLP headers and custom credential name redaction.

src/powercontext/cli/env_file.py
happy-v587: Fixed. Replaced `shlex.split(..., comments=True)` with a custom `_strip_comment` state machine that only treats `#` as a comment at an unquoted word boundary (preceded by whitespace or at start-of-line). Values like `TOKEN=abc#123` and `URL=https://example.com/#frag` are now preserved. Added parser tests covering mid-value hashes, trailing comments, quoted values, URL fragments, and unterminated quotes.

docs/en/docs/how-to/full-capability-runtime.md
happy-v587: Completed both guides. Added: (1) minimal/full capability matrix table, (2) `powercontext doctor` in verification flow, (3) Source-to-flush-to-Memory round trip section with curl examples and `powercontext stats` output, (4) data-location section covering `POWERCONTEXT_HOME` and platform-specific paths, (5) stop/restart guidance, (6) FTS-only fallback explanation. Chinese guide mirrors all additions.

src/powercontext/cli/config.py
```diff
@@ -0,0 +1,883 @@
...
+    name: str,
+    shared: dict[str, str],
+) -> ProviderVariable | None:
+    if name in shared:
+        typer.echo(f"Reusing {name} for {role}.")
+        return ProviderVariable(name, shared[name])
+    value = typer.prompt(
+        name,
+        default="",
+        hide_input=_is_secret_name(name),
```
Teingi: The generator already knows that this first variable is the selected credential, but `_collect_provider_variable` decides `hide_input` only from its name. A valid custom name such as `SERVICE_CREDENTIAL` therefore prompts with `hide_input=False`, even though the same name is later recorded as a credential for `config show` redaction, so the secret is visibly echoed while it is entered. Could we pass the credential semantics into this prompt and add a regression test that asserts the input is hidden?

docs/en/docs/how-to/full-capability-runtime.md
```diff
@@ -0,0 +1,178 @@
...
+
+The Server replies `202` with `"status":"accepted"`. Then flush the Scope, which runs Memory extraction:
+
+```bash
+curl -X POST http://127.0.0.1:8000/v1/memory/flush \
+  -H 'content-type: application/json' \
+  -d '{"scope_id":"project:quickstart"}'
+```
+
+Expect `"status":"processed"` and `"processed_source_count":1`. Confirm the inventory:
```
Teingi: This expected result is not deterministic. The scheduler can process the source before the manual flush, making `idle`/0 valid, while extraction may legally produce zero candidates. Also, active Memory does not prove Embedding worked: `_attach_embeddings` deliberately returns unembedded projections when embedding is unavailable or times out. Could the guide use cursor-aware accepted outcomes and verify Embedding with an actual vector/hybrid query, with the same correction in the Chinese guide?

Teingi (commented):
I found three remaining issues on the current head.

src/powercontext/cli/env_file.py
```diff
@@ -0,0 +1,150 @@
...
+        if escaped:
+            escaped = False
+        elif quote:
+            if character == "\\" and quote == '"':
+                escaped = True
+            elif character == quote:
+                quote = ""
+        elif character in {"'", '"'}:
+            quote = character
+        elif character == "#" and (index == 0 or line[index - 1] in {" ", "\t"}):
```
Teingi: This boundary check treats any immediately preceding space as starting a comment, even when that space was escaped. As a result, `TOKEN=abc\ #123` parses as `abc `, whereas Bash assigns `abc #123`, so a valid value is silently truncated. Could `_strip_comment` track backslash escaping outside quotes as well, with a regression test for this case?

src/powercontext/cli/config.py
happy-v587: Fixed. Added `is_credential` parameter to `_collect_provider_variable`; when called from `_collect_initial_provider_variable` for the credential variable, it passes `is_credential=True`. This hides input for custom credential names (e.g. `SERVICE_CREDENTIAL`) that do not match `_is_secret_name` patterns. Added a regression test using `unittest.mock.patch` to assert `hide_input=True` is passed.

src/powercontext/cli/env_file.py
happy-v587: Fixed. Rewrote the backslash handling to use a sentinel approach: `_replace_escaped_spaces` replaces `\\ ` (backslash-space) with a NUL sentinel before `shlex.split` runs, then the sentinel is restored to a space in the final value. This correctly handles `TOKEN=abc\\ #123` → value `abc #123`, matching Bash behavior. Added a regression test for this case.

docs/en/docs/how-to/full-capability-runtime.md
happy-v587: Fixed. Replaced the non-deterministic `processed_source_count:1` assertion with cursor-aware verification (`current_cursor > previous_cursor`). Added a `/v1/memory/search` query with `mode=vector` to confirm embedding vectors exist, and `powercontext stats` to show Embedding model usage. Both EN and ZH guides updated.

docs/en/docs/how-to/full-capability-runtime.md
```diff
@@ -0,0 +1,192 @@
...
+
+Confirm the inventory and verify Embedding by searching with vector mode:
+
+```bash
+curl -X POST http://127.0.0.1:8000/v1/memory/search \
+  -H 'content-type: application/json' \
+  -d '{"scope_id":"project:quickstart","query":"verifiable steps","mode":"vector"}'
+```
+
+A non-empty `hits` list confirms the Embedding model produced vectors. An empty list with `"status":"ok"` on
```
Teingi: `powercontext ready` cannot produce `status: ok`; readiness uses `ready`, `degraded`, or `not_ready`. More importantly, an empty explicit vector search does not prove FTS fallback: with no Memory, the API returns `mode: null` and `hits: []` before capability validation, while with Memory but no vector capability, explicit `mode=vector` returns 422. This can label a zero-candidate or failed setup as a healthy fallback. Could the guide detect fallback via `powercontext capabilities` and require `mode: vector` plus a vector-matched hit when verifying Embedding, with the same correction in the Chinese guide?

src/powercontext/cli/env_file.py
```diff
@@ -0,0 +1,196 @@
...
+                escaped = True
+                result.append(character)
+            elif character == quote:
+                quote = ""
+                result.append(character)
+            else:
+                result.append(character)
+        elif character == "\\":
+            next_index = index + 1
+            if next_index < len(chars) and chars[next_index] == " ":
```
Teingi: The sentinel workaround only handles a single backslash followed by a literal space; it still does not model escapes outside quotes. For example, `TOKEN=abc\\ #comment` is parsed as `abc #comment` while Bash assigns `abc\`, and `TOKEN=abc\" #comment` is rejected while Bash assigns `abc"`. Escaped tabs are truncated as well. Could we replace this with one quote/escape-aware lexer and add parity, tab, and escaped-quote regression tests?

Teingi (commented):
I found two remaining issues on the current head.

src/powercontext/cli/config.py
```diff
@@ -0,0 +1,885 @@
...
+def show_command(
+    env_file: Annotated[Path, typer.Option(help="Environment file to inspect.")] = Path(".env"),
+) -> None:
+    """Print effective assignments with credentials redacted."""
+
+    try:
+        content = env_file.read_text(encoding="utf-8")
+        values = parse_environment(content, source=str(env_file))
+        recorded = _managed_metadata(content).get("credentials", "")
+    except (EnvironmentFileError, OSError) as error:
```
AsperforMias: `_managed_metadata` raises `ConfigError` when the managed markers are mismatched or repeated, but this handler only catches `EnvironmentFileError`/`OSError`, and the two classes are independent `ValueError` subclasses. As a result, `config show` on a hand-edited file containing two managed blocks (or one dangling marker) dies with an uncaught traceback instead of the clean `Error:`/exit 2 that `config validate` produces for the same file, since `validate` does catch `ConfigError`. Could we catch `ConfigError` here as well and add a regression test for `show` on a malformed-marker file?

src/powercontext/cli/config.py
```diff
@@ -0,0 +1,885 @@
...
+
+
+def _collect_custom_connection(role: str) -> tuple[str, list[ProviderVariable], tuple[str, ...]]:
+    reference_model = "openai-chat:model-name"
+    typer.echo("Advanced mode uses a complete Pydantic AI provider:model identifier.")
+    typer.echo(f"Reference: {MODEL_CONFIGURATION_URL}")
+    model = typer.prompt(f"{role} model identifier", default=reference_model).strip()
+    shared: dict[str, str] = {}
+    variables, credentials = _collect_initial_provider_variables(role, model, shared)
+    _collect_additional_provider_variables(role, shared, variables)
```
AsperforMias: Variables collected by `_collect_additional_provider_variables` never extend `credentials`, and the calls there pass no `is_credential`, so `hide_input` depends solely on the name pattern. An additional credential under a non-pattern name such as `MYTOKEN` or `SERVICE_CREDENTIAL` is therefore echoed visibly while typed during init, never reaches the `# credentials=` metadata, and prints verbatim in `config show` — the same class of leak the earlier fix closed for the initial credential variable. Could we record additional credential names (or ask whether each additional variable is a credential and pass `is_credential` accordingly), with a regression test for both the hidden prompt and the `config show` redaction?

AsperforMias (commented):
Two remaining issues on the current head.

src/powercontext/cli/config.py
happy-v587: Good catch on the exception split. The handler already catches `ConfigError` as of `fb79a0a` (`show_command` now uses `except (ConfigError, EnvironmentFileError, OSError, UnicodeError)`), so `config show` exits with the same clean `Error:` / exit 2 that `config validate` produces for the same file — the review was written against the previous head. The requested regression test is added in `5618d3b`: `tests/test_config_cli.py::test_show_reports_malformed_managed_markers_without_a_traceback` writes a file containing two begin markers and asserts exit code 2, the `mismatched or repeated PowerContext managed markers` message, and no Traceback in the output.

docs/en/docs/how-to/full-capability-runtime.md
happy-v587: Fixed on head (`fb79a0a`). The guide no longer attributes `status: ok` to `powercontext ready`: the verification block shows `Status: ready` from `powercontext ready`, while the liveness/readiness lines come from `doctor` output. Fallback detection is done through `powercontext capabilities` — the runtime is declared full-capability only when all four search modes are listed, and a capability line of `auto, fts` explicitly signals FTS-only fallback. Embedding is verified only by a `mode: vector` search hit whose `citation.entry_id` equals the source-linked entry previously recorded from `/v1/memory/entries/list`; a hit for another entry, an empty `hits` list, or `"mode":null` fails the round-trip check. The Chinese guide mirrors all of it, and `tests/test_full_capability_docs.py::test_full_capability_guide_binds_memory_evidence_to_the_captured_source` pins these rules for both locales.

src/powercontext/cli/env_file.py
happy-v587: Fixed in `8a24737`: the sentinel workaround was dropped and the parse path is now a single quote/escape-aware lexer, `_split_shell_words` in `src/powercontext/cli/env_file.py`. Outside quotes, `\` escapes any following character literally; inside double quotes only `$`, `` ` ``, `"` and `\` are escapable; single-quoted text stays literal; escaped tabs survive instead of being truncated. `#` still starts a comment only at an unquoted word boundary, so values like `TOKEN=abc#123` remain intact. Parity regressions cover the reported cases against shell behavior: `TOKEN=abc\ #comment` (even backslashes) yields a trailing `abc\`, `TOKEN=abc\" #comment` yields `abc"`, `TOKEN=abc\ #123` keeps the space and drops only the real comment, and escaped tabs are preserved — see `tests/test_env_file.py::test_even_backslash_before_space_starts_a_comment_like_shell`, `test_backslash_escaped_quote_outside_quotes_is_literal`, `test_backslash_escaped_space_preserves_hash_value`, and `test_backslash_escaped_tab_preserves_hash_value`.

src/powercontext/cli/config.py
happy-v587: Implemented the second suggested option on head (`5618d3b`). While collecting each additional variable, init now asks "Treat {name} as a credential (hidden input and redaction)?" — skipped when the name already matches the secret-name patterns — passes that answer into `_collect_provider_variable`, so input is hidden for non-pattern names like `MYTOKEN` or `SERVICE_CREDENTIAL`, and records every marked name into the returned credential tuple. Those names flow into the managed block's `# credentials=` metadata, which `config show` redacts.

Regression tests cover both observables you called out:
- hidden prompt: `test_collect_additional_provider_variables_hides_and_records_marked_credentials` (asserts `hide_input=True` reaches the value prompt for `MYTOKEN`) plus `..._keeps_unmarked_values_visible` for opt-out;
- redaction: `test_custom_connection_marks_prompted_credential_for_show_redaction` checks `# credentials=MYTOKEN`, `configuration_from_document(...).credentials`, and `<redacted>` output with no value leak; `test_init_hides_and_redacts_marked_additional_credentials` runs the whole interactive `init` through the CLI runner.

Teingi (approved):
LGTM

PR #1371, opened 2026-08-26
fix(dsh): declare required plugin services

### LINKED ISSUE #1370: bug: declare every required DSH plugin service
## Describe the bug

The DSH plugin registers against five runtime services (`tools`, `agents`, `commands`, `skills`, and `systemPrompt`), but its top-level `inject` contract currently declares only `tools` and `agents`. The command, skill, and system-prompt registration paths retrieve their services with `ctx.get(...)` and silently return when a service is unavailable.

This makes plugin readiness asymmetric with the services it actually consumes. Under a custom DSH profile, startup ordering, service replacement, or HMR, the plugin can finish applying with only part of its public surface mounted. A later service appearance does not cause the already-applied plugin to register the missing surface.

## Steps to reproduce

1. Load the PowerContext DSH plugin while `tools` and `agents` are available but one of `commands`, `skills`, or `systemPrompt` is not yet available.
2. Observe that the plugin is considered ready and `apply()` runs.
3. Make the missing service available after the plugin has applied.
4. Check `/pc`, the `project-context` skill, and the PowerContext system-prompt guidance.

The same lifecycle problem can be reproduced by unloading and restoring one of those services after an initial successful mount.

## Expected behavior

The plugin should declare every service required by `apply()`. Cordis should keep it pending until all dependencies are ready, dispose the complete plugin when a required service disappears, and remount every surface when the service returns. Invalid direct registration calls should fail visibly instead of silently producing a partial plugin.

## Actual behavior

The plugin can remain partially mounted: tools and recall may work while `/pc`, the skill, or guidance is absent. Restoring the missing service does not reliably restore the skipped registration because the service was not part of the plugin's injection lifecycle.

## Impact

Without this change, users can silently lose diagnostics, project-memory instructions, or the `/pc` workflow depending on profile composition and service timing. The resulting behavior is difficult to diagnose because startup succeeds and no error identifies the missing required service. It also makes service reload/HMR behavior inconsistent with Cordis lifecycle guarantees.

## Environment

- PowerContext: current `master` (`3a6cb01`)
- Integration: `integrations/dsh/plugins/powercontext`
- DSH: `0.1.0-rc.6`
- Cordis: `4.0.1`
- Node.js: `24.14.1`
- OS used for reproduction: Windows 11

## Additional context

A regression test should exercise the real Cordis plugin lifecycle, not only call registration helpers directly: pending before all five services exist, complete mount after readiness, complete disposal on dependency removal, and complete remount when the dependency returns.

- [x] I am willing to submit a pull request to fix this bug.

knqiufan:

Closes #1370.

# Rationale for this change

The DSH plugin used five runtime services but declared only `tools` and `agents` in its top-level Cordis injection contract. If `commands`, `skills`, or `systemPrompt` was absent during `apply()`, registration silently returned and left a partially mounted plugin. Because those services were outside the injection lifecycle, later startup, replacement, or HMR did not reliably restore the skipped surfaces.

The plugin is incomplete without any of these services. Declaring all five lets Cordis enforce readiness and lifecycle ownership: the plugin waits for every dependency, disposes as a unit when one disappears, and remounts as a unit when it returns.

- Declare `tools`, `agents`, `commands`, `skills`, and `systemPrompt` as required plugin services.
- Add a shared `requireService()` guard so invalid direct registration paths fail visibly instead of silently returning.
- Add focused registration contract tests for `/pc`, the `project-context` skill, and system-prompt guidance.
- Upgrade the DSH call-through suite to install the plugin through a real Cordis `Context` and verify pending, mount, dependency removal, remount, and live `/pc` command behavior.
- Pin Cordis 4.0.1 as a development dependency so the lifecycle regression test runs against the DSH-compatible runtime contract.
- Make the repository OpenAPI contract take precedence during monorepo builds while retaining the plugin copy as a standalone fallback.
- Synchronize the standalone OpenAPI copy, regenerate the operation table and bundle, and normalize generated EOF newlines so release builds are reproducible.
- Add a dedicated DSH CI job that tests before build, builds, verifies zero generated drift, tests again, and runs the Cordis e2e suite.

Yes. `/pc`, the project-context skill, tool registrations, recall guidance, and the pre-step hook now mount and unmount atomically. Profiles that omit a required service remain pending instead of appearing healthy with silently missing functionality. Release bundles also retain the current repository API operation metadata. There is no configuration format or persisted-data change.

# How was this change tested?

- `uv lock --locked`
- `uv run prek run -a`
- `uv run python scripts/generate_api.py --check`
- `uv run python scripts/generate_js_operations.py --check`
- `uv run python -m pytest tests/test_api_contract.py tests/test_js_operations.py` (29 passed)
- `uv run python -m pytest tests/e2e/test_dsh_http_chain.py tests/test_dsh_cli.py -q` (8 passed)
- `pnpm test` before and after `pnpm build` (72 passed each)
- `pnpm test:e2e` (8 passed)
- generated OpenAPI, operation table, and bundle remain unchanged after `pnpm build`
- `pnpm pack:release`, with the packed bundle verified to retain `list_handoff_report_known_scopes` and `get_handoff_report.scope = true`
- `pnpm test && pnpm run typecheck` in `integrations/pi/plugins/powercontext` (29 passed; type check passed)
- `uv run zensical build -s` from a clean worktree

# AI usage statement

OpenAI Codex was used to review the Cordis/DSH lifecycle and release-generation contracts, implement the fixes and regression tests, and run validation. The resulting diff, generated artifacts, package contents, and test behavior were manually reviewed before submission.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -52,6 +52,18 @@ jobs:
       - name: Run end-to-end tests
         run: make e2e-test
 
+  dsh-package:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out
+        uses: actions/checkout@v7
+
+      - name: Set up project tools
+        uses: jdx/mise-action@v2
+
+      - name: Test DSH package
+        run: make js-test
+
   pi-package:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -58,6 +58,9 @@ jobs:
       - name: Check out
         uses: actions/checkout@v7
 
+      - name: Set up the Python environment
+        uses: ./.github/actions/setup-python-env
+
       - name: Set up project tools
         uses: jdx/mise-action@v2
 
--- a/integrations/dsh/plugins/powercontext/lib/index.js
+++ b/integrations/dsh/plugins/powercontext/lib/index.js
@@ -321,6 +321,12 @@ const OPERATIONS = {
 		location: "body",
 		scope: false
 	},
+	list_handoff_report_known_scopes: {
+		method: "POST",
+		path: "/v1/handoff-reports/scopes/list-known",
+		location: "body",
+		scope: false
+	},
 	get_handoff_report_project: {
 		method: "POST",
 		path: "/v1/handoff-reports/projects/get",
@@ -355,7 +361,7 @@ const OPERATIONS = {
 		method: "POST",
 		path: "/v1/handoff-reports/get",
 		location: "body",
-		scope: false
+		scope: true
 	},
 	record_handoff_report_activity: {
 		method: "POST",
@@ -566,6 +572,14 @@ var PowerContextClient = class {
 		});
 	}
 };
+
+//#endregion
+//#region src/dsh-service.ts
+function requireService(ctx, name$1) {
+	const service = ctx.get(name$1);
+	if (service == null) throw new Error(`${PLUGIN_NAME} requires the "${name$1}" service`);
+	return service;
+}
 
 //#endregion
 //#region src/secrets.ts
@@ -920,9 +934,7 @@ async function handlePcCommand(rawInput, runtime, scopeId, signal) {
 	};
 }
 function registerCommands(ctx, runtime) {
-	const commands = ctx.get("commands");
-	if (!commands) return;
-	commands.register({
+	requireService(ctx, "commands").register({
 		name: "pc",
 		description: "PowerContext status, search, review, and diagnostics",
 		handler: async (invocation) => {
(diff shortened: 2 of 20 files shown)


integrations/dsh/plugins/powercontext/lib/index.js
```diff
@@ -321,6 +321,12 @@ const OPERATIONS = {
 		location: "body",
 		scope: false
 	},
+	list_handoff_report_known_scopes: {
```
Teingi, line 324: [P1] The committed bundle is not reproducible through the release path. On this head, pnpm test fails five generated-operation checks, and pnpm run build succeeds but removes this operation and changes get_handoff_report.scope back to false. sync-openapi.mjs resolves the stale plugin-local openapi/powercontext.yaml before the repository contract, while pack:release invokes that build. The release artifact therefore silently discards the API-table changes committed here. Please make the repository OpenAPI take precedence, keep the plugin copy only as a standalone fallback, sync the copy and update the stale assertion, regenerate the bundle, and cover pnpm build && pnpm test.

knqiufan: Thank you for the thorough review. I've fixed the issue and please take a moment to review again to see if there are any other issues.

.github/workflows/master.yml
```diff
@@ -52,6 +52,18 @@ jobs:
...
         run: make e2e-test
 
+  dsh-package:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out
+        uses: actions/checkout@v7
+
+      - name: Set up project tools
+        uses: jdx/mise-action@v2
```
Teingi, line 65: [P1] This job does not install uv. mise only provisions Node and pnpm from .mise.toml, but make js-test ends with the live E2E suite, whose server helper spawns uv run powercontext server run. With the declared toolset and uv absent, both E2E suites fail with spawn uv ENOENT. The current head also has no Main workflow run, so the green checks do not exercise this job. Please install uv before make js-test, preferably by reusing ./.github/actions/setup-python-env (or setup-uv), so this new gate can actually run on ubuntu-latest.

Teingi (approved):
LGTM

PR #1376, opened 2026-08-27
fix(opencode): keep interrupted plugin installs recoverable

happy-v587:
# Rationale for this change

`_install_plugin` placed the plugin bundle before its ownership manifest. An interruption between the two atomic renames left the plugin installed without `.powercontext-opencode.json`: every later `setup opencode` run then reported an unowned-plugin conflict until the file was deleted by hand.

- Install the ownership manifest before the plugin bundle. The manifest content is constant, so the interrupted state becomes a manifest without a plugin, which the next setup run replaces instead of refusing to overwrite.
- Add a regression test that interrupts the second rename and asserts a retry completes without manual cleanup.

Yes. A setup interrupted mid-install is now self-healing on retry instead of requiring manual deletion of the installed plugin.

# How was this change tested?

- `uv run pytest -q tests/test_opencode_cli.py` — 14 passed.
- Full suite: 908 passed, 9 skipped (the OpenCode host e2e fails identically on stock master with locally installed OpenCode 1.18.23 due to the documented environment limitation).
- Mutation check: reverting the ordering change makes the new test fail.

# AI usage statement

Prepared with an AI coding assistant; the change was verified with automated tests and mutation checks as listed above.

### DIFF
--- a/src/powercontext/cli/opencode.py
+++ b/src/powercontext/cli/opencode.py
@@ -236,8 +236,10 @@ def _install_plugin(source: Path, target: Path) -> None:
             json.dumps({"schema": 1, "owner": "powercontext", "integration": "opencode-plugin"}, indent=2) + "\n",
             encoding="utf-8",
         )
-        os.replace(staging, target)
+        # Install the ownership manifest first: an interruption then leaves a manifest without a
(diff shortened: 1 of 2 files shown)


Teingi (approved):
LGTM

PR #1377, opened 2026-08-27
fix(server): keep successful Skill publication responses intact

happy-v587:
# Rationale for this change

The Dashboard publish route ran the external Skill registry scan outside error handling. If the scan raised (filesystem or registry errors), the client received a 500 even though the Skill had already been published to the Agent target, and there was no signal about what had failed.

- Wrap the post-publication scan in logging-and-continue: the projection is already written to disk, so a scan failure is logged as a warning instead of failing the request.
- Add a regression test that makes `scan()` raise, asserts publish still returns 200 with `state: current`, verifies the file landed on disk, and checks exactly one warning is logged.

Yes. A registry scan failure after a successful publication now returns the normal Dashboard response (with registry discovery possibly stale) plus a logged warning, rather than an unrelated 500.

# How was this change tested?

- `uv run pytest -q tests/test_dashboard.py` — 8 passed.
- Full suite: 908 passed, 9 skipped (the OpenCode host e2e fails identically on stock master with locally installed OpenCode 1.18.23 due to the documented environment limitation).
- Mutation check: reverting the handler change makes the new test fail with the propagated OSError.

# AI usage statement

Prepared with an AI coding assistant; the change was verified with automated tests and mutation checks as listed above.

### DIFF
--- a/src/powercontext/server/web.py
+++ b/src/powercontext/server/web.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 from collections.abc import Mapping
 from functools import cache
 from typing import Literal
@@ -28,6 +29,7 @@ from fastapi.templating import Jinja2Templates
 from jinja2 import Environment, PackageLoader, select_autoescape
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 
+from powercontext._logging import log_safely
 from powercontext.artifacts import ArtifactRef
 from powercontext.builtin.artifacts.skill import AgentKind, AgentSkillTarget, ExternalSkillResolutionStatus
 from powercontext.builtin.artifacts.skill.projection import (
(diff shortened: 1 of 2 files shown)


src/powercontext/server/web.py
```diff
@@ -160,7 +164,14 @@ async def publish(
...
-        await application.external_skills.for_scope(request.scope_id).scan()
+        # The publication itself succeeded above; registry bookkeeping failure must not turn the
+        # response into a 500 because _skill_projection_response reports on-disk state anyway.
+        try:
+            await application.external_skills.for_scope(request.scope_id).scan()
+        except Exception as error:
+            log_safely(
+                logger, logging.WARNING, "PowerContext external Skill scan failed after publication", exc_info=error
+            )
         return await _skill_projection_response(application, request.scope_id, skill, self._targets)
```
Teingi, line 175: The response path still re-enters the registry here. If scan() failed because the registry database is unavailable, rather than only because of a filesystem scan error, _skill_projection_response() immediately calls external_skills.list() and the request still returns 500 after the Skill was written. I reproduced this by making both scan() and list() raise the same OSError: the projection file existed, but the endpoint returned 500. Could we skip the registry lookup after a failed scan, or isolate list() too, and report discovery as unavailable?

happy-v587: Fixed in b254b15: `_skill_projection_response()` now treats registry reads as best-effort — when `list()` fails (for example an unavailable registry database), registrations degrade to empty with a logged warning while the response still reports on-disk state (`state: current`, stale `discovery: unavailable`) instead of a 500 after successful publication.

Covered by `test_publish_reports_stale_discovery_when_registry_database_is_unavailable`, which makes both scan and list fail and asserts exactly one warning for each failure path.

Teingi (approved):
LGTM

PR #1380, opened 2026-08-27
feat(inference): support workload-specific model endpoints

### LINKED ISSUE #1379: feat: support workload-specific inference endpoints
## Feature description

Allow generation, embeddings, and LLM reranking to use separate model identifiers, base URLs, static headers, and model settings.

Generation and embeddings may run on different OpenAI-compatible services and require different routing or authentication headers.

Each workload should be configurable independently. Header values must be treated as secrets and excluded from ordinary logs. Base URLs may include a gateway path prefix, while operation paths remain owned by the selected Pydantic AI provider.

## Alternatives considered

A dedicated rerank API is out of scope. Reranking continues to use an LLM through Pydantic AI.

PsiACE:

Closes #1379.

# Rationale for this change

Generation, embeddings, and LLM reranking may use different model services and require different routing or authentication headers.

- Add workload-specific models, base URLs, static headers, and model settings.
- Configure sensitive headers on provider clients instead of request settings.
- Keep LLM reranking on the regular model interface.
- Document provider-owned operation paths and configuration limits.
- Add acceptance coverage for separate generation, embedding, and rerank services.

New inference environment variables are available. Existing configurations keep their current behavior.

Custom endpoints and headers require an explicit OpenAI- or Anthropic-compatible model identifier. Arbitrary operation-path rewriting is not supported.

# How was this change tested?

- `make check`
- `make test`
- `make docs-test`
- Local acceptance services for generation, embeddings, and LLM reranking

# AI usage statement

OpenAI Codex with GPT-5 was used for implementation and review.


src/powercontext/builtin/runtime/composition.py
```diff
@@ -445,85 +452,270 @@ async def _generation_pipelines(
...
+                default_headers = _resolve_headers(headers)
+                api_key = _pop_header(default_headers, "x-api-key")
+                client = AsyncAnthropic(
+                    base_url=None if base_url is None else str(base_url),
+                    api_key=api_key or os.getenv("ANTHROPIC_API_KEY") or "api-key-not-set",
+                    default_headers=default_headers,
+                )
+                resources.push_async_callback(client.close)
+                return AnthropicProvider(anthropic_client=client)
+            return AnthropicProvider(base_url=str(base_url), api_key="api-key-not-set")
```
Teingi: Custom Anthropic base URLs discard the standard environment credential when no workload headers are set. This branch always passes `api-key-not-set`, so even with `ANTHROPIC_API_KEY` configured the SDK sends the placeholder and authenticated endpoints return 401. Please preserve the environment/provider credential resolution here, falling back to a placeholder only when no credential exists, and add a request-level regression test for a custom base URL with empty headers.

Teingi (approved):
LGTM

PR #1386, opened 2026-08-28
fix: pass embedding dimension to provider and surface sqlite-vec probe failures

### LINKED ISSUE #1385: bug: Configured embedding dimension is never sent to the provider; dimension mismatch surfaces as swallowed degraded/unavailable or a misleading sqlite-vec probe failed
## Describe the bug

The config has an `EMBEDDING_DIMENSION=1536` parameter, but this requirement is only used internally (table creation, result validation) — it is never sent to the Provider in the actual request. The Provider doesn't know how many dimensions you want, so it returns the model default: 1024.

So: we ask for 1536, we get 1024, validation fails with "embedding must contain exactly 1536 dimensions". The key problem: this error is never shown to the user. The server log only says `PowerContext Server is degraded`, and `doctor` only reports `inference.embedding: unavailable` — no reason anywhere. I only found the real cause by writing a script that calls the underlying interface directly.

Second issue: after changing the dimension, an old database crashes on startup, and the error is misleading. The vector table fixes its dimension at creation time. If you started once with 1024, then change the config to 1536 and restart: `CREATE VIRTUAL TABLE IF NOT EXISTS ... float[1536]` is a no-op because the table already exists (still 1024), so the probe insert of a 1536-dimension vector fails. This error is uniformly wrapped into `CapabilityNotSupportedError: ... sqlite-vec probe failed`, which looks like the extension is not installed. In fact the extension works fine — it's just a table/config dimension mismatch. Deleting the database "fixed" it and hid the real cause.

Third issue: leftover probe row. The self-check inserts a temporary row (`vector_id=-1`) and deletes it only on success. If the process is killed between insert and delete, the row stays and every subsequent startup fails with the same misleading error — the only fix is deleting the database.

## Steps to reproduce

1. Install: `uv tool install "powercontext[cli,server] @ git+https://github.com/oceanbase/powercontext.git@master"`
2. Configure `.env`:
   - `POWERCONTEXT_SERVER_DATABASE_KIND=sqlite`
   - `POWERCONTEXT_SERVER_INFERENCE_EMBEDDING_MODEL=openai:<your embedding model>`
   - `POWERCONTEXT_SERVER_INFERENCE_EMBEDDING_DIMENSION=1536`
   - `OPENAI_BASE_URL=<your OpenAI-compatible endpoint>`
   - `OPENAI_API_KEY=<your key>`
   - Requirement: the model returns 1024 dimensions by default but supports returning 1536 via the `dimensions` parameter.
3. `powercontext server run --env-file .env` → starts "successfully", log only shows `degraded`
4. `powercontext doctor` → `inference.embedding: unavailable`, no reason given
5. Prove the Provider actually supports 1536: curl the embeddings endpoint with `"dimensions":1536` → returns 1536 numbers; without it → returns 1024. This shows the client simply doesn't send the parameter.
6. Reproduce the crash: start once with 1024 to create the DB, change to 1536, restart → `CapabilityNotSupportedError: vector (sqlite-vec probe failed)`; meanwhile loading the sqlite-vec extension manually works fine.
7. Reproduce the leftover row: kill -9 while the probe row is inserted but not yet deleted, then restart → fails every time; deleting the DB recovers.

## Expected behavior

- Send the configured dimension to the Provider via the `dimensions` parameter; or, if not passing it, validate at startup that "configured dimension == Provider default" and fail with an explicit message.
- Surface the underlying error (in logs and/or doctor output) instead of just `degraded`/`unavailable`.
- Detect and clearly report a mismatch between the existing table dimension and the configured profile (ideally auto-migrate).
- Make the probe idempotent: tolerate and clean up a pre-existing `vector_id=-1` leftover row.

## Actual behavior

- Fresh DB: server starts degraded, doctor reports `unavailable`, no real cause anywhere.
- Existing DB: startup crashes with `CapabilityNotSupportedError: vector (sqlite-vec probe failed)`.
- Leftover row: every startup crashes; only deleting the database recovers.

## Environment

- PowerContext version: 0.0.3.dev55+g3fcfff19f (uv tool install from master)
- Python version: 3.12.13 (uv-managed CPython, aarch64)
- OS: macOS (Apple Silicon)
- Relevant dependencies: sqlite-vec 0.1.9
- Embedding Provider: OpenAI-compatible endpoint (model with 1024-dim default that supports the `dimensions` parameter)

happy-v587:

Closes #1385.

# Rationale for this change

The configured `EMBEDDING_DIMENSION` was never sent to the embedding provider, so any configured dimension different from the provider default failed validation. The failure was invisible to users: the readiness probe swallowed the underlying error and only reported `degraded` / `inference.embedding: unavailable`, and if the failure happened during vector index initialization it was uniformly wrapped as `CapabilityNotSupportedError: vector (sqlite-vec probe failed)`, which misleadingly suggests a missing sqlite-vec extension. Additionally, an interrupted startup could leave a `vector_id=-1` probe row behind, which poisoned every subsequent startup with the same masked error until the database file was deleted manually.

- `composition.py`: pass the configured embedding profile dimension to the `Embedder` via `EmbeddingSettings(dimensions=...)`, so providers actually receive the requested dimension instead of returning their default.
- `sqlite/memory_index.py`: clear a leftover `vector_id=-1` probe row before the vector index probe, making startup idempotent after an interrupted run.
- `sqlite/memory_index.py`: include the underlying error in the `CapabilityNotSupportedError` detail, with a dedicated hint when the existing vec table dimension does not match the configured profile (message contains both dimension values).
- `composition.py`: log embedding readiness probe failures via `logger.exception` before marking `inference.embedding` unavailable.

- Behavior: an explicitly configured embedding dimension is now honored by providers that support the `dimensions` parameter; providers that reject the parameter will now surface a visible error instead of silently mismatching dimensions.
- Error messages: vector index probe failures now carry the underlying cause text instead of a bare `sqlite-vec probe failed`; a stale vec table with a different dimension produces an actionable message naming both dimensions.
- No API, schema, or migration changes.

# How was this change tested?

- 5 new unit tests (`tests/builtin/persistence/test_sqlite_memory_vector_index.py`, `tests/builtin/runtime/test_composition_embedding.py`): leftover-row self-heal, dimension-mismatch message contains both dimensions plus chained cause, generic cause surfacing, `dimensions` settings reaching the `Embedder`, and no-embedding-config passthrough.
- `make test`: no new failures vs the base commit (16 vs 17 pre-existing environment-dependent failures); `make check` clean.
- Manual end-to-end validation against a real OpenAI-compatible endpoint: fresh DB with dimension 1536 now reaches `ready` with vector/hybrid search enabled; a stale 1024-dim table now fails with a message containing both dimensions; an injected `vector_id=-1` row is cleaned up on the next start.

# AI usage statement

Built with opencode (GLM, Z.ai) under human direction and review; all code, tests, and validation steps were reviewed and confirmed by the author.

### DIFF
--- a/src/powercontext/builtin/persistence/sqlite/memory_index.py
+++ b/src/powercontext/builtin/persistence/sqlite/memory_index.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import json
 import struct
 from collections.abc import Mapping
+from re import search
 from typing import Any
 
 from sqlalchemy import (
@@ -342,7 +343,8 @@ class SQLiteMemoryVectorIndex:
             ).one_or_none()
             await connection.execute(_DELETE_VECTOR_SQL, {"vector_id": -1})
         except SQLAlchemyError as error:
-            raise CapabilityNotSupportedError("vector", _probe_failure_detail(error, self.profile.dimension)) from error
+            detail = await _probe_failure_detail(connection, error, self.profile.dimension)
+            raise CapabilityNotSupportedError("vector", detail) from error
         if row is None or int(row[0]) != -1:
             raise CapabilityNotSupportedError("vector", "sqlite-vec probe returned an invalid row")
 
@@ -528,17 +530,34 @@ class SQLiteMemoryVectorIndex:
         return tuple(hydrated)
 
 
-def _probe_failure_detail(error: SQLAlchemyError, dimension: int) -> str:
+async def _probe_failure_detail(connection: AsyncConnection, error: SQLAlchemyError, dimension: int) -> str:
     orig = getattr(error, "orig", None)
     detail = str(orig) if orig is not None else str(error)
-    normalized = detail.casefold()
-    if "dimension" in normalized:
+    existing = await _existing_vec_dimension(connection)
+    if existing is not None and existing != dimension:
         return (
-            "sqlite-vec probe failed: the existing pc_memory_entry_vec table dimension does not match "
-            f"the configured embedding profile dimension {dimension}; migrate the table or align the "
-            f"embedding dimension configuration ({detail})"
+            "sqlite-vec probe failed: the existing pc_memory_entry_vec table dimension "
+            f"{existing} does not match the configured embedding profile dimension {dimension}; "
+            f"migrate the table or align the embedding dimension configuration ({detail})"
         )
     return f"sqlite-vec probe failed: {detail}"
+
+
+async def _existing_vec_dimension(connection: AsyncConnection) -> int | None:
+    """Return the dimension of a pre-existing vec0 table, or None when it cannot be confirmed."""
+
+    try:
+        row = (
+            await connection.exec_driver_sql(
+                "SELECT sql FROM sqlite_master WHERE type = 'table' AND name = 'pc_memory_entry_vec'"
+            )
+        ).one_or_none()
+    except SQLAlchemyError:
+        return None
+    if row is None:
+        return None
+    match = search(r"float\[(\d+)\]", str(row[0]))
+    return int(match.group(1)) if match is not None else None
 
 
 def _pack_vector(vector: tuple[float, ...]) -> bytes:
--- a/src/powercontext/builtin/persistence/sqlite/memory_index.py
+++ b/src/powercontext/builtin/persistence/sqlite/memory_index.py
@@ -327,6 +327,9 @@ class SQLiteMemoryVectorIndex:
                 f"USING vec0(embedding float[{self.profile.dimension}])"
             )
             probe = _pack_vector((0.0,) * self.profile.dimension)
+            # A previous run may have been interrupted between inserting and deleting
+            # the probe row; clear any leftover before probing again.
+            await connection.execute(_DELETE_VECTOR_SQL, {"vector_id": -1})
             await connection.execute(
                 _INSERT_VECTOR_SQL,
                 {"vector_id": -1, "embedding": probe},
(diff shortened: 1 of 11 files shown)


src/powercontext/builtin/runtime/composition.py
```diff
@@ -562,7 +562,11 @@ def provider_factory(provider_name: str) -> Provider[object]:
 
     def adapter(instrument: InstrumentationSettings | bool | None) -> EmbeddingModel:
         return PydanticAIEmbeddingModel(
-            embedder=Embedder(model, instrument=instrument),
+            embedder=Embedder(
+                model,
+                settings={"dimensions": profile.dimension},
```
Teingi, line 567: The configured dimension reaches the provider now, but if a provider or model rejects `dimensions`, the reason is still lost: `dependency_readiness_probe()` collapses the resulting configuration error to `misconfigured`, so the Server and `doctor` do not explain what was rejected. I reproduced this with an OpenAI-compatible mock endpoint: the request contained `dimensions: 1536`, the endpoint returned 400, and readiness exposed only `misconfigured`. Please surface a stable redacted reason without logging the raw provider response, and cover this rejection path.

src/powercontext/builtin/persistence/sqlite/memory_index.py
```diff
@@ -525,6 +528,19 @@ async def hydrate(
         return tuple(hydrated)
 
 
+def _probe_failure_detail(error: SQLAlchemyError, dimension: int) -> str:
+    orig = getattr(error, "orig", None)
+    detail = str(orig) if orig is not None else str(error)
+    normalized = detail.casefold()
+    if "dimension" in normalized:
```
Teingi: This classifies every sqlite error containing `dimension` as an existing-table mismatch. On a fresh database with dimension 65536, sqlite-vec correctly reports its 8192 maximum, but this code tells the user to migrate an existing table that does not exist. Please emit the mismatch hint only after confirming that a pre-existing table has a different dimension; otherwise preserve the actual probe error, and add a fresh oversized-dimension regression test.

Teingi (approved):
LGTM

PR #1387, opened 2026-08-28
docs(rfc):memory capacity and sharding

### LINKED ISSUE #1321: bug: Memory append storage and latency grow superlinearly with entry history
### Describe the bug

Repeated explicit Memory appends have superlinear persistent-storage growth and steadily increasing write latency, even when each entry contains only a few bytes of text.

Every Memory Revision stores the complete current manifest. Appending one entry therefore writes manifests of size 1, 2, 3, ... N across the immutable revision history. Entry bodies are not duplicated, but the cumulative manifest directory approaches quadratic growth.

There is currently no maximum entry count, manifest-byte budget, automatic Memory split, or tombstone compaction boundary. The unpaginated list_memory_entries path also materializes the complete current manifest and all referenced entry versions.

The Memory RFC identifies split and compaction thresholds as future work, but the current service can grow past practical limits without an explicit error or degradation signal.

### Steps to reproduce

At commit a6e0dfe8b0b20b9a40e905aaa07c62427c1418f4:

1. Create a fresh SQLite-backed Server with Dashboard, MCP, metrics, and access logging disabled.
2. Repeatedly call POST /v1/memory/remember in one scope with one short unique fact per request.
3. Close the Server.
4. Run PRAGMA wal_checkpoint(TRUNCATE) and VACUUM.
5. Record the database size and average duration of the final 20 writes.
6. Repeat with 50, 100, and 200 entries.

Observed:

| Entries | SQLite database bytes | Average final 20 writes |
| ---: | ---: | ---: |
| 50 | 671,744 | 40.57 ms |
| 100 | 1,634,304 | 65.21 ms |
| 200 | 5,222,400 | 111.81 ms |

The test body was only bounded fact NNNN.

Relevant paths:

- src/powercontext/builtin/artifacts/memory/models.py:62-65
- src/powercontext/builtin/artifacts/memory/service.py:699-731
- src/powercontext/builtin/persistence/artifacts.py:243-262
- src/powercontext/builtin/persistence/memory.py:157-175
- docs/en/rfcs/0014_memory_layer_design.md, especially the deferred split/compaction thresholds

### Expected behavior

A long-lived Memory service needs an explicit, observable capacity contract.

At minimum:

- Enforce a current-manifest entry or byte budget before unbounded growth.
- Return a stable error or route new entries to another Memory when the budget is reached.
- Paginate or otherwise bound list_memory_entries and complete-history operations.
- Define and test split and inactive-tombstone compaction behavior.
- Add a scale regression test that measures storage and write amplification at increasing entry counts.

The exact split/compaction design may require an RFC, but the current API should not accept growth indefinitely without a budget.

### Actual behavior

Each append reconstructs and persists the complete manifest in a new immutable Artifact Revision. Storage and write work increase with the number of prior entries, with no configured ceiling or automatic split.

### Environment

- Python: 3.14.3
- Backend: SQLite with WAL, checkpointed and vacuumed before measurement
- OS: Windows 11
- External inference and embedding providers: disabled

MaoMengww:

Closes #1321

# Rationale for this change

Currently, because the manifest persists a full snapshot on every write, storage and write costs grow super-linearly, and in extreme cases the system can run beyond its practical limits. RFC 0014 explicitly listed Memory splitting, inactive tombstone compaction, and a shared routing manifest as unresolved. To contain these extreme cases, this RFC caps unbounded growth while preserving the manifest semantics, offering a concise solution under the conditions it defines.

- Introduces entry-count-bounded shards (`memory_manifest_max_entries`, default 200) and a routing domain derived from the configured root ID through strict `root.sNNNN` suffixes; no schema migration, no entry migration, and no change to the flat-v1 persistence shape.
- Defines write routing: unconditional pure adds can automatically split to the next-ordinal shard; writes carrying `expected_revision`, citation-driven revise/retire/reactivate, and organize retain their owner shard and return the stable `memory_capacity_exceeded` error when capacity is insufficient.
- Extends flush to single-plan semantics covering the entire routing domain: one candidate extraction over the active union, global deduplication by canonical bytes, and per-shard commits submitted together with the Source cursor in the same outer transaction, protected by a routing-domain barrier.
- Defines candidate identity as `(memory_artifact_id, entry_id)` to prevent cross-shard target misidentification.
- Defines bounded reads through exact immutable per-shard Revision snapshots and cursors that cannot be replayed across routing domains, as well as tombstone compaction audited by `drop` changes with no cooldown period.
- Fully documents capacity/compaction, flush transactions and concurrency, API semantics, compatibility, drawbacks, alternatives, prior art, unresolved questions, and acceptance criteria.

No released behavior changes.

# How was this change tested?

make check

# AI usage statement

Claude Code was used in this change to analyze the Memory service implementation and the related design discussions, and to assist in drafting and cross-checking the synchronized English and Chinese RFC documents. The proposal was submitted by the author for review and revision by maintainers within the RFC process.

### DIFF
--- a/docs/en/rfcs/1387_memory_capacity_and_sharding.md
+++ b/docs/en/rfcs/1387_memory_capacity_and_sharding.md
@@ -1,6 +1,6 @@
 - Proposal Name: `memory_capacity_and_sharding`
-- Start Date: 2026-08-27
-- RFC PR: [to be created]
+- Start Date: 2026-08-28
+- RFC PR: [oceanbase/powercontext#1387](https://github.com/oceanbase/powercontext/pull/1387)
 
 # Summary
 
--- a/docs/zh/rfcs/1387_memory_capacity_and_sharding.md
+++ b/docs/zh/rfcs/1387_memory_capacity_and_sharding.md
@@ -1,6 +1,6 @@
 - 提案名称：`memory_capacity_and_sharding`
-- 开始日期：2026-08-27
-- RFC PR：[待创建]
+- 开始日期：2026-08-28
+- RFC PR：[oceanbase/powercontext#1387](https://github.com/oceanbase/powercontext/pull/1387)
 
 # Summary
 
@@ -262,4 +262,3 @@ head 在分页期间会前插新 entry；以 `(ordinal, entry_id)` 继续读取�
 - `drop` 可审计、不会破坏旧 citation，mutation response 不把 drop 当作业务目标。
 - search、prepare context、statistics 和 citation lifecycle 对合法 root shards 的 identity 处理一致。
 - 合法大批量 remember/candidate 不被人为的 100-item 上限截断；只受最终 manifest budget 影响。
-
--- a/docs/en/rfcs/0000_memory_capacity_and_sharding.md
+++ b/docs/en/rfcs/0000_memory_capacity_and_sharding.md
@@ -1,0 +1,263 @@
+- Proposal Name: `memory_capacity_and_sharding`
+- Start Date: 2026-08-27
+- RFC PR: [to be created]
+
+# Summary
+
+This RFC introduces entry-count-bounded shards for flat-v1 Memory, plus a routing domain centered on the configured root Artifact ID. One logical Memory may consist of multiple non-migrating Memory Artifact shards; each shard continues to use the existing Artifact Revision, manifest, immutable entry versions, active-head projection, and head CAS.
+
+Pure unconditional adds may automatically advance to the next shard once capacity is reached. Writes carrying `expected_revision`, citation-driven revise/retire/reactivate, and citation-driven organize all retain their original shard identity and return a stable capacity error when capacity is insufficient. Source flush observes all current heads of the routing domain in a single plan, performs candidate generation and exact deduplication over the full active-entry union; revise candidates commit in place on their owner shard, add candidates route to the write-target shard, and all affected shards plus the Source cursor commit in one outer transaction.
+
+This RFC does not change the flat-v1 persistence shape, does not migrate historical entries, does not physically delete old Revisions, and does not turn scope or routing domain into a new Core identity. It addresses the capacity and read-amplification problems that the current single-Artifact manifest develops as entries grow.
+
+# Motivation
+
+RFC 0014 defines Memory as a lifecycle of immutable Artifact Revisions, with the manifest as the authoritative directory of current entry state. The current implementation writes the complete directory into every new Revision on append; consecutive adds therefore persist 1, 2, …, N-entry directories, and storage and write amplification grow quadratically. `entries()` and `changes()` also read the complete manifest or complete history as a single unit, becoming unbounded single operations as data grows. While preserving RFC 0014's semantics of a manifest as a full snapshot, this RFC caps the extreme cases driven by unbounded growth, preventing unbounded storage and write amplification.
+
+This RFC aims to:
+
+- preserve the Memory flat-v1 design, capping storage and write amplification at a suitable place;
+- confine add pressure to a single shard, with a deterministic next shard when needed;
+- preserve the immutable semantics of entry identity, entry versions, and citations;
+- make flush's candidate context and exact deduplication cover the routing domain, not a single shard;
+- keep cross-shard revise on its original owner shard, and give multi-shard flush all-or-nothing commit semantics;
+- bind list and changes to an immutable Revision as the snapshot boundary;
+
+# Guide-level explanation
+
+## Shards and the routing domain
+
+A scope's configuration designates a `root_artifact_id` for Memory. The root itself is ordinal 0; automatically created follow-up shards use IDs like `root.s0001`, `root.s0002`. Only the root and strict suffixes matching that root belong to the same routing domain; other Memory Artifacts under the same scope are not included merely because their family is `memory`.
+
+```mermaid
+flowchart LR
+    Root["root: tenant-memory\nordinal 0"] --> S1["tenant-memory.s0001\nordinal 1"]
+    S1 --> S2["tenant-memory.s0002\nordinal 2"]
+    Domain["routing domain\nstrict root + suffix"] -.discovery and ordering.-> Root
+    Domain -.discovery and ordering.-> S1
+    Domain -.discovery and ordering.-> S2
+    Other["other-memory\nother Memory in same scope"] -.excluded.-> Domain
+```
+
+A shard is a storage and concurrency boundary, not a new logical entry identity. Once an entry lands in a shard, its `memory_artifact_id` is fixed; subsequent revise, retire, reactivate, and citation all use that shard's exact ArtifactRef. Split only decides where new entries are written; it never moves old entries.
+
+## Write behavior
+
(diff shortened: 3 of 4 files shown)


Teingi (commented):
Thanks for writing this up. Bounded shards and stable entry ownership are useful directions, but the RFC still has several correctness and liveness gaps that should be resolved before acceptance. Because this PR defines design rather than implementing the open bug, please also change `Closes #1321` to `Refs #1321`, add both RFC pages to the English and Chinese navigation, and bring the Chinese text fully back into sync with the English document.

docs/en/rfcs/1387_memory_capacity_and_sharding.md
```diff
@@ -0,0 +1,263 @@
...
+`drop` is an audit operation for manifest compaction, not a physical deletion, and it must not be treated as a return target of a user business change.
+
+# Reference-level explanation
+
+## Invariants
+
+### Identity
+
+- Memory's persistent identity remains `(family, artifact_id)`; the routing domain is only a discovery set derived from the configured root.
+- Shard ID resolution first matches the exact root, then applies a strict `root.s([1-9][0-9]*)` full match; ordering uses the parsed integer, forbidding accidental prefix matches.
```
Teingi, line 94: The generated names and this parser disagree: `root.s0001` does not match `root.s([1-9][0-9]*)` under a full match (`root.s1` does). The first automatic split would therefore create a shard that routing cannot rediscover. Please define one canonical suffix grammar and use it for both generation and parsing.

docs/en/rfcs/1387_memory_capacity_and_sharding.md
```diff
@@ -0,0 +1,263 @@
...
+
+`drop` is an audit operation for manifest compaction, not a physical deletion, and it must not be treated as a return target of a user business change.
+
+# Reference-level explanation
+
+## Invariants
+
+### Identity
+
+- Memory's persistent identity remains `(family, artifact_id)`; the routing domain is only a discovery set derived from the configured root.
```
Teingi, line 93: Deriving membership solely from Artifact ID spelling can absorb an existing valid Artifact. For example, `memory.s0001` is already a valid Memory ID; after configuring `memory` as the root, it would suddenly join this routing domain even if it was created independently. Please persist explicit routing membership or ownership, including upgrade conflict handling, rather than treating the suffix as proof of ownership.

docs/en/rfcs/1387_memory_capacity_and_sharding.md
```diff
@@ -0,0 +1,263 @@
...
+```
+
+## Flush transaction and concurrency
+
+Comparing the set of heads before and after planning is not enough to protect shards used only as candidate/dedupe context. Flush must lock the entire routing domain:
+
+- when the root exists, lock the root head row right after the outer transaction begins, then read all legal shard heads by ordinal;
+- when the root does not exist, protect bootstrap with a database-supported `SERIALIZABLE` or equivalent predicate-conflict mechanism; a non-existent row is not a lockable object;
+- every writer that changes the routing domain — explicit shard writes, split, flush, organize, and new-shard creation — first acquires the same barrier;
+- head reads, candidate generation, global deduplication seeding, per-shard plans, ordinal-ordered apply, and cursor save all use the same bound connection;
```
Teingi, line 155: This keeps the database barrier and connection open across model candidate generation, whose latency and retry duration are unbounded. That can block every writer in the routing domain. Please read an immutable head vector, run extraction outside the transaction, then acquire a short barrier, re-enumerate and recheck every head, and commit atomically.

docs/en/rfcs/1387_memory_capacity_and_sharding.md
```diff
@@ -0,0 +1,263 @@
...
+
+- explicit remember carrying `expected_revision`;
+- revise, retire, reactivate targeting an exact citation;
+- organize against a specified Memory.
+
+When capacity or identity requirements are not met, these operations return a stable `memory_capacity_exceeded`, so the caller sees the failure reason instead of having conditions pinned to an old Revision quietly applied to a different Artifact.
+
+## Flush: single-shard semantics extended
+
+Source flush first reads all current heads of the routing domain. The candidate pipeline receives the active entries of all shards at once; its input is not truncated by this RFC, and candidate rules do not change with shard count. When the model proposes a revise, it must specify both `memory_artifact_id` and `entry_id`; an add carries no target identity.
```
Teingi, line 58: Sharding bounds each Artifact, but this call still receives an unbounded domain-wide union. Once it exceeds the model context, the same Source window will fail on every retry and the cursor cannot advance. Please make selection token-budgeted and bounded, with a progress-safe fallback; global exact deduplication can remain server-side through canonical hashes and bytes.

docs/en/rfcs/1387_memory_capacity_and_sharding.md
```diff
@@ -0,0 +1,263 @@
...
+version membership = (memory_artifact_id, entry_version_id)
+```
+
+Targets must belong to the validated current entries of this snapshot's exact heads; unknown owners, bare targets, old-Revision targets, or cross-shard versions are rejected as untrusted candidates. Ordinary `MemoryEntryInput` is produced only after the mapping succeeds — a bare ID output by the model must never enter a commit directly.
+
+## Capacity and compaction
+
+All write paths that construct a new Memory Revision apply the same final rule to the resulting manifest: first compact droppable inactive tombstones according to the threshold and capacity conditions, then check `len(final_manifest) <= memory_manifest_max_entries`. The capacity check happens before embedding/projection preparation; a failure writes no Artifact, entry, projection, or index.
+
+`memory_tombstone_compaction_threshold` controls whether compaction is triggered by tombstone count, and `memory_compaction_max_drops` only bounds the internal drop batch of one Revision; it does not limit the number of remember or candidate inputs. Legitimate large batches are not implicitly truncated; only a final manifest over budget fails as a whole.
```
Teingi, line 138: `max_drops` can prevent compaction from making progress. With capacity 200, 200 existing entries, 10 droppable tombstones, `max_drops=5`, and six conditional adds, compaction leaves 195 entries, the adds produce 201, and the all-or-nothing rollback restores all ten tombstones. Retrying repeats forever even though enough entries are droppable. Under pressure, please allow enough drops to fit the request or define a separately committable compaction step.

docs/zh/rfcs/1387_memory_capacity_and_sharding.md
```diff
@@ -0,0 +1,264 @@
...
+### Identity
+
+- `Memory` 的持久化 identity 仍是 `(family, artifact_id)`；routing domain 只是由配置 root 派生的发现集合。
+- shard ID 解析先匹配 exact root，再使用严格的 `root.s([1-9][0-9]*)` full match；排序使用解析后的整数，禁止前缀误匹配。
+- root 和 suffix 以及 entry/version ID 继续遵守既有 identifier 长度和 ASCII 约束。生成 suffix 超出限制时返回 `split_id_exhausted`。
+- `MemoryEntryVersion.memory_artifact_id`、manifest 中的 entry pointer 和 citation 的 `memory_ref` 必须指向同一 shard。
+
+### Manifest 与版本
+
+每个 shard 继续使用 flat-v1：manifest 只保存 `entry_id`、`entry_version_id`、`entry_content_hash` 和 `state`，正文在不可变 `MemoryEntryVersion` 中。最终 manifest 的 entry 数不得超过 `memory_manifest_max_entries`（默认 200，`ge=1`）；条目数是容量判断维度。
```
Teingi, line 100: The translations are not currently synchronized: the Chinese RFC jumps from the capacity limit directly to Candidate identity and omits the English rationale and measurements for the default of 200. Please carry the same normative rationale into both documents, and clean up the incomplete wording near the Summary (`以及。`) before acceptance.

PR #1277, opened 2026-08-19
fix: oceanbase identity collation collapses case-variant scope_id and source_id

thunguo:
# Which issue or RFC does this PR close?

Closes #1276.

# Rationale for this change

`scope_id` and `source_id` are opaque, byte-exact identities. Shared schema columns were declared as dialect-neutral `String(...)` with no collation, so MySQL/OceanBase inherited the server default `utf8mb4_general_ci`. SQLite compares the same columns with BINARY semantics, which is why the SQLite test matrix never caught this.

On OceanBase that collation collapses case-variant and accent-variant keys:

- `scope_id="Alpha"` and `"alpha"` are one Scope, so listing the second returns the first Scope's Memory.
- `source_id="Turn-1"` and `"turn-1"` are one Source, so a legitimate second capture is rejected as `409 source_conflict` and that turn never enters the Source journal.

# What changes are included in this PR?

- Add `identity_string(length)` in `tables.py`. SQLite keeps `String`; MySQL/OceanBase emit `VARCHAR(..., charset="utf8mb4", collation="utf8mb4_bin")`.
- Use that type for identity columns in the shared persistence schema, Memory search indexes, and Handoff Report catalog / activity / workspace tables.
- Assert compiled MySQL DDL includes `CHARACTER SET utf8mb4 COLLATE utf8mb4_bin` for `scope_id`, `source_id`, and `source_type`.
- Add a SQLite and OceanBase parametrized e2e test that writes Memory under a case-variant and an accent-variant `scope_id`, asserts the other variant lists no entries, and captures both `Turn-1` and `turn-1`.

`create_all(checkfirst=True)` does not rewrite existing column collations, and OceanBase rejects `ALTER COLUMN ... COLLATE` when foreign keys exist. Existing MySQL/OceanBase schemas must be recreated on a new database (or by dropping and recreating the tables). Startup does not run an in-place ALTER.

# Are there any user-facing changes?

Public HTTP and SDK APIs are unchanged. On a fresh MySQL/OceanBase schema, identity comparison now matches SQLite and the Runtime: case-variant and accent-variant `scope_id` / `source_id` values are distinct.

This is a persisted-schema break for existing OceanBase/MySQL deployments. Databases created before this change keep `utf8mb4_general_ci` on identity columns and will still leak or reject those variants until the schema is recreated. SQLite files are unaffected.

# How was this change tested?

[...]

PR #1389, opened 2026-08-28
fix(oceanbase): reject legacy schema collations

### LINKED ISSUE #1320: bug: OceanBase collation fix does not upgrade existing schemas
### Describe the bug

The fix for #1276 changes the DDL used for new MySQL/OceanBase tables, but an existing deployment is not upgraded and is not rejected at startup.

identity_string() now uses utf8mb4_bin for opaque identity columns. Its own docstring states that create_all(checkfirst=True) does not rewrite existing collations and that existing MySQL/OceanBase schemas must be recreated.

The startup path still calls SQLAlchemy create_all(checkfirst=True) and has no schema-version or collation compatibility check. Consequently, upgrading the PowerContext package can leave the old utf8mb4_general_ci columns active while the Server starts normally.

The required recreation procedure is currently present only in a source-code comment, not as an operator-visible migration or startup diagnostic.

Related issue: #1276 fixed fresh-schema behavior. This issue tracks the existing-schema upgrade path.

### Steps to reproduce

1. Install a PowerContext revision from before 3ce87532780bc365547ee05c5ddf1c3d9e4a8acd.
2. Initialize an OceanBase database whose identity columns inherit utf8mb4_general_ci.
3. Upgrade the package to commit a6e0dfe8b0b20b9a40e905aaa07c62427c1418f4.
4. Start the Server against the existing database.
5. Inspect the identity-column collations, or rerun the Alpha/alpha and Turn-1/turn-1 reproduction from #1276.

Current initialization path:

- src/powercontext/builtin/persistence/tables.py:51-62 explicitly says existing schemas must be recreated.
- src/powercontext/builtin/persistence/schema.py:25-39 only calls create_all(checkfirst=True).
- src/powercontext/builtin/persistence/oceanbase/profile.py initializes the profile without validating existing identity-column collations.
- No current CLI command or documented operator migration performs this upgrade.

This conclusion follows directly from the current initialization contract. A real legacy OceanBase instance was not mutated during this report.

### Expected behavior

Upgrading an existing deployment must not silently retain a schema that violates byte-exact scope and source identity semantics.

One of the following should happen:

1. A tested migration safely converts every related identity column and constraint; or
2. Startup inspects the actual schema and fails closed with an actionable, credential-free error that identifies the required backup/rebuild/restore procedure.

The repository should include an old-schema-to-current integration test and operator documentation.

### Actual behavior

The Server creates missing objects but does not alter or validate existing collations. An old database can start successfully while preserving the cross-scope collision behavior fixed for fresh schemas in #1276.

### Environment

- Affected backend: existing OceanBase/MySQL-mode schema created before 3ce87532780bc365547ee05c5ddf1c3d9e4a8acd
- Fresh SQLite schemas are not affected by this collation issue.

Vanillaxi:

Closes #1320 .

# Rationale for this change
The fresh-schema fix in #1277 makes opaque identity columns use `utf8mb4_bin`, but SQLAlchemy's `create_all(checkfirst=True)` neither alters nor rejects tables created by earlier PowerContext versions. Therefore,upgraded deployments can start with legacy `utf8mb4_general_ci` identity columns and still collapse case- or accent-variant scope and source IDs.

- Inspect existing PowerContext `VARCHAR` identity columns through `information_schema.COLUMNS` before `create_tables()`,
- Reject every incompatible identity column with `IncompatibleOceanBaseSchemaError`.
- Add a deterministic regression test that verifies rejection happens before table creation.
- Add a real OceanBase integration test that creates the legacy `pc_sources` schema with `utf8mb4_general_ci` and verifies that the current profile rejects it.
- Document the data-only migration procedure in the English and Chinese troubleshooting guides.

Existing OceanBase/MySQL-mode deployments with incompatible identity-column collations now fail during Server startup .The error identifies the affected columns and directs operators to a backup, rebuild, and data-only restore procedure. 

# How was this change tested?
- `uv run pytest -q tests/builtin/persistence/test_oceanbase_profile.py` 
 ` tests/builtin/persistence/test_mysql_schema.py` —  17 passed, 2 skipped
- The regression failed before the fix with `DID NOT RAISE IncompatibleOceanBaseSchemaError`; the opt-in live OceanBase case passed with `POWERCONTEXT_TEST_OCEANBASE_URL`.
- `make check` — passed
- `make test` — `943 passed, 10 skipped`
- `make docs-test` — passed

# AI usage statement
OpenAI Codex was used to implement and review.

### DIFF
--- a/docs/en/docs/how-to/troubleshoot.md
+++ b/docs/en/docs/how-to/troubleshoot.md
@@ -174,10 +174,50 @@ so the previous database remains available for recovery:
 4. Create a new empty OceanBase MySQL-mode database and point `POWERCONTEXT_SERVER_DATABASE_URL` at it. Start the
    current PowerContext version once to create tables with `utf8mb4_bin`, then stop it before restoring data.
 5. Import only the exported row data into the existing new tables with OceanBase `obloader`, again **without
-   `--ddl`**. Importing the old DDL would recreate the incompatible collations.
-6. Compare row counts, inspect the identity-column collations, and test identities that differ only by case or accent.
-   Start normal traffic only after these checks pass. Retain the old database and backup until the new deployment has
-   completed your rollback window.
+   `--ddl`**. Keep foreign-key enforcement enabled and run these three layers separately. The examples use CSV; if
+   you exported SQL data, replace `--csv` with `--sql` in all three commands. Fill in `<connection-options>` through
+   your approved secret-handling process and make `<new-database>` select the database created in step 4.
+
+   Before running the commands, compare the exported table files with `SHOW TABLES` in the target database. Every
+   exported table named below must exist in the target; if one is missing, stop and create it with the current
+   PowerContext configuration before importing. Remove a name only when the source export does not contain that table.
+   If the source predates the seven `pc_handoff_report_*` tables, remove them from Layer 1. If the export contains them
+   but Handoff Report is disabled, temporarily enable it in step 4 to create the current tables, restore their data,
+   and return the setting to its intended value only after verification.
+
+   Layer 1 contains parents and tables without foreign keys:
+
+   ```bash
+   obloader <connection-options> -D <new-database> --csv \
+     --table 'pc_source_journal_heads,pc_sources,pc_artifacts,pc_source_cursors,pc_external_skill_registrations,pc_model_usage_daily,pc_recall_token_daily,pc_handoff_report_projects,pc_handoff_report_project_revisions,pc_handoff_report_workstreams,pc_handoff_report_workstream_revisions,pc_handoff_report_workspace_bindings,pc_handoff_report_activity_heads,pc_handoff_report_activities' \
+     -f <export-directory>
+   ```
+
+   After Layer 1 completes successfully, import its children in Layer 2:
+
+   ```bash
+   obloader <connection-options> -D <new-database> --csv \
+     --table 'pc_artifact_heads,pc_artifact_lineage_sources,pc_artifact_lineage_artifacts,pc_artifact_candidate_versions,pc_memory_entry_versions' \
+     -f <export-directory>
+   ```
+
+   After Layer 2 completes successfully, import the remaining children in Layer 3:
+
+   ```bash
+   obloader <connection-options> -D <new-database> --csv \
+     --table 'pc_artifact_candidate_heads,pc_memory_entry_heads' \
+     -f <export-directory>
+   ```
+
+   Wait for each invocation to complete successfully before starting the next. Treat any OBLoader error, bad record,
+   or conflict record as a failed restore. Order within a layer is irrelevant because no table in a layer references
+   another table in the same layer.
+6. If the installation has additional PowerContext-managed tables not listed above, these tested layers do not
+   classify them. Inspect their foreign-key constraints and place each table after all of its parents; do not add
+   them to an all-table invocation.
+7. Compare source and target row counts for every restored table, inspect the identity-column collations, and test
+   identities that differ only by case or accent. Start normal traffic only after every check passes. Retain the
+   source database, verified backup, and export through the rollback window.
 
 If records were previously merged because the old collation considered their identities equal, changing the schema
 cannot reconstruct them. Resolve those records from an authoritative source before accepting writes.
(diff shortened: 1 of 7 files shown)


tests/builtin/persistence/test_oceanbase_profile.py
```diff
@@ -186,3 +275,40 @@ async def scenario() -> None:
...
+    reason="set POWERCONTEXT_TEST_OCEANBASE_URL to a dedicated OceanBase MySQL-mode test database",
+)
+def test_live_oceanbase_profile_rejects_old_schema_collation() -> None:
+    async def scenario() -> None:
+        assert LIVE_URL is not None
+        config = OceanBaseConfig(url=SecretStr(LIVE_URL))
+        async with OceanBaseProfile.open(config, tables=()) as setup_profile:
+            try:
+                async with setup_profile.database.transaction() as connection:
+                    await connection.exec_driver_sql("DROP TABLE IF EXISTS `pc_sources`")
```
Teingi: This test unconditionally drops `pc_sources`. It passed on an owned empty database, but failed when I ran it against an existing valid PowerContext schema, where `pc_sources` is referenced by foreign keys; in a sources-only database it would remove existing data. Please give the test its own disposable database/schema and only drop objects it created.

src/powercontext/builtin/persistence/oceanbase/profile.py
```diff
@@ -30,11 +31,36 @@
...
 
 _DIALECT_DRIVER = "mysql+aoceanbase"
 _DIALECT_REGISTRY_NAME = "mysql.aoceanbase"
 _DIALECT_MODULE = "pyobvector"
 _DIALECT_CLASS = "AsyncOceanBaseDialect"
+_SCHEMA_COLUMNS_QUERY = """
+SELECT TABLE_NAME, COLUMN_NAME, COLLATION_NAME
+FROM information_schema.COLUMNS
+WHERE TABLE_SCHEMA = DATABASE()
+  AND LEFT(TABLE_NAME, 3) = 'pc_'
```
Teingi: This query treats every `VARCHAR` column in any `pc_*` table as an identity column. On the configured OceanBase instance, adding a valid `pc_notes(display_name ... utf8mb4_general_ci)` table made startup fail for `pc_notes.display_name` even though the managed identity columns were compatible. Please validate only the owned identity `(table, column)` pairs derived from `profile.tables` / metadata.

tests/builtin/persistence/test_oceanbase_profile.py
Vanillaxi: Thanks for the review. This test now creates a UUID-named disposable database, builds the legacy schema there, and drops only that database in `finally` after confirming it was created successfully. It no longer modifies or drops `pc_sources` in the configured database.

src/powercontext/builtin/persistence/oceanbase/profile.py
Vanillaxi: Thanks ,done.

docs/en/docs/how-to/troubleshoot.md
```diff
@@ -154,6 +154,34 @@ powercontext server run
...
+so the previous database remains available for recovery:
+
+1. Stop the Server and every process that writes to the database.
+2. Take and verify a full recoverable backup using your normal OceanBase backup procedure.
+3. Export the PowerContext table data with OceanBase `obdumper` in CSV or SQL data mode **without `--ddl`**. Keep the
+   export and the original database unchanged until the migration is verified. Supply credentials through your
+   approved secret-handling process rather than placing them in logs or documentation.
+4. Create a new empty OceanBase MySQL-mode database and point `POWERCONTEXT_SERVER_DATABASE_URL` at it. Start the
+   current PowerContext version once to create tables with `utf8mb4_bin`, then stop it before restoring data.
+5. Import only the exported row data into the existing new tables with OceanBase `obloader`, again **without
```
Teingi, line 176: Importing all row data here is still unsafe for the current schema. `BUILTIN_TABLES` has 11 foreign-key constraints, and on the configured OceanBase instance a `pc_artifact_heads` row imported before its `pc_artifacts` parent is rejected. The [OBLOADER documentation](https://en.oceanbase.com/docs/community-obloader-obdumper-en-10000000001028883) also warns that data may not be imported in dependency order. Please provide a tested parent-first table order, or another explicit FK-handling procedure with post-restore integrity checks, instead of a generic all-table import.

Vanillaxi: done

Teingi (approved):
LGTM

PR #1396, opened 2026-08-30
docs(rfc): define access control

Teingi:
# Which issue or RFC does this PR close?

Closes #1395.

# Rationale for this change

PowerContext has a complete Handoff continuity model, but the current optional static Bearer authentication does not express per-user, per-Workstream, or exact-Handoff visibility. A receiver's Handoff authorization observation is also not an ACL.

This RFC defines a least-privilege Access Control boundary for transferring one exact Handoff from user A to user B while preserving the existing Handoff, Receipt, evidence, and Runtime domain semantics. It also defines a replaceable authorization-provider boundary so deployments can use the built-in store, Casbin, OpenFGA, or an AuthZEN-compatible PDP.

# What changes are included in this PR?

- Add matching English and Chinese RFC documents for Handoff Access Control.
- Define trusted Principal, resource, action, role, Access Binding, and audit models.
- Separate one-off exact-Handoff receiver grants from long-term scope roles.
- Define bounded citation-manifest evidence access without exposing the complete scope.
- Define public Access API shapes and `x-powercontext-access` OpenAPI metadata.
- Define one Server PEP shared by HTTP, MCP, and the Dashboard.
- Define `AuthorizationProvider` and optional `RelationshipWriter` integration boundaries.
- Document built-in, Casbin, OpenFGA, AuthZEN, OPA, and Cerbos provider profiles.
- Specify failure behavior, safe listing and pagination, migration modes, implementation slices, and acceptance tests.

# Are there any user-facing changes?

No released behavior changes are included. This PR adds an RFC design only. A later implementation PR would add public Access APIs, authorization configuration, 403 behavior, and Handoff receiver UX after separate review.

# How was this change tested?

- `make docs-test`
- `uv run prek run --files docs/en/rfcs/1396_handoff_access_control.md docs/zh/rfcs/1396_handoff_access_control.md`
- `git diff --cached --check`

# AI usage statement

OpenAI Codex (GPT-5) was used to inspect the existing Handoff, HTTP, MCP, and RFC contracts; draft the bilingual RFC; and run repository validation. The author directed the design scope and reviewed the resulting artifacts.

Issue #1395, opened 2026-08-30
RFC: define Handoff access control and RBAC interoperability


Teingi:
## Feature description

Define a resource-level Access Control boundary for PowerContext, using Handoff transfer as the first concrete profile.

The design should answer how user A can transfer work to user B while controlling which Handoff, evidence, and follow-up operations B can access. It should also provide an integration boundary for open-source authorization systems without putting users, roles, or ACLs into Handoff content or `scope_id`.

## Problem and proposed solution

The current Server supports an optional global static Bearer token. A valid token authenticates a caller but does not distinguish users, Workstreams, exact Handoff Revisions, or actions. The existing Handoff Receipt authorization observation records whether a receiver believes it can continue; it is not authentication or an ACL.

The proposed RFC should define:

- a trusted `PrincipalRef` established by authentication;
- one Server Policy Enforcement Point shared by HTTP, MCP, and the Dashboard;
- a replaceable `AuthorizationProvider` using subject, action, resource, and context decisions;
- an optional `RelationshipWriter` for systems that support grant mutation;
- scope roles for long-term collaboration;
- an exact `handoff.receiver` grant for one committed immutable Handoff Revision;
- citation-manifest evidence access that does not expose the complete scope;
- public Access APIs for self-checks, safe resource listing, role and Binding management, and audit;
- fail-closed 401, 403, and 503 behavior;
- adapters or conformance profiles for Casbin, OpenFGA, and AuthZEN-compatible PDPs such as OPA or Cerbos.

An exact Handoff receiver should be able to read, inspect evidence, and acknowledge only the granted Revision. This grant must not imply access to `latest`, adjacent Revisions, aggregate Handoff Reports, general Memory or Source queries, Handoff commits, Task Outcomes, tools, networks, or credentials. Continuing the Workstream requires a separate scope role.

`scope_id` remains a business partition and must not become an authorization credential or tenant proof. Runtime domain APIs should remain free of Principal, role, and permission parameters; enforcement belongs at the Server boundary.

## Alternatives considered

- **ACL fields on Handoff or scope:** couples immutable domain data to identity lifecycle and external policy management.
- **Scope roles only:** exposes more Workstream history and evidence than a one-off receiver needs.
- **Anonymous bearer share links:** make possession of a URL the identity and are difficult to audit or revoke safely.
- **Dashboard-only visibility controls:** do not protect direct HTTP or MCP calls.
- **One mandatory policy engine:** limits embedded and enterprise deployment choices.

[...]

PR #1398, opened 2026-08-30
feat(server): add access control

Teingi:

Implements #1395.

Design reference: #1396.

# Rationale for this change

The current optional static Bearer token authenticates one deployment-local caller but cannot express per-user, per-scope, or exact-Handoff visibility. Handoff Receipts also record receiver observations; they are not authorization grants.

This change adds an independent Server Access Control boundary so user A can grant user B least-privilege access to one committed Handoff Revision without exposing the surrounding scope. Runtime domain APIs remain free of Principal, role, and permission parameters, and deployments can replace the built-in decision provider and relationship store with an external authorization system.

- Add typed Principal, action, structured resource, fixed role, binding, decision, and data-minimized audit models.
- Add a built-in hierarchical RBAC provider and Server-owned relational binding/audit schema for SQLite, OceanBase, and seekDB.
- Add replaceable `AuthorizationProvider`, `RelationshipWriter`, and `AccessAuditStore` boundaries for integrations such as OpenFGA, Casbin, Oso, or an AuthZEN-compatible PDP.
- Add `/v1/access/me`, check, batch-check, safe resource listing, role listing, binding management, and audit APIs plus typed Python SDK methods.
- Annotate protected OpenAPI operations with generated `x-powercontext-access` requirements and enforce them at one Server PEP before Runtime dependencies or business metrics.
- Preserve the authenticated Principal across the MCP internal ASGI bridge while still executing authorization for logical MCP operations.
- Add `disabled`, `legacy-static-admin`, and `enforced` rollout modes; the default legacy mode preserves the existing static-token administrator behavior.
- Add English and Chinese configuration and HTTP API documentation.
- Fix OceanBase Handoff trigger cursor creation to avoid an incompatible nested SAVEPOINT while retaining SQLite's concurrent-insert behavior.
- Add focused Access Control, HTTP, MCP, SDK, OpenAPI, persistence, and Server tests.

Yes. Authenticated deployments gain public Access APIs, stable 403 behavior, exact-Handoff receiver grants, Access audit records, and two Access configuration settings. Existing static-token deployments remain compatible through the default `legacy-static-admin` mode.

The change adds Server-owned Access tables but does not add identity or ACL fields to Handoff, Source, Memory, Work, or other Runtime domain records. `scope_id` remains a business partition rather than an authorization credential.

# How was this change tested?

- `PRE_COMMIT_HOME=/tmp/powercontext-prek-cache make check`
- `.venv/bin/python -m pytest -q -p no:cacheprovider` (`974 passed, 9 skipped`)
- `make contract-test` (`30 passed`)
- `make docs-test`
- Real configured acceptance journey using the local `.env` OceanBase, generation LLM, and Embedding provider:
  - persisted and vector-searched Memory;
  - generated, finalized, and committed a real Handoff;
  - granted one exact Revision from A to B;
  - verified HTTP allow/deny boundaries, MCP Principal propagation, audit redaction, restart persistence, revocation, and post-revocation denial;
  - revoked the test binding and removed the failed-run Runtime fixture while preserving append-only low-sensitivity audit evidence.

# AI usage statement

OpenAI Codex (GPT-5) was used to inspect the repository contracts, implement the Access Control layer and generated API changes, diagnose the real OceanBase SAVEPOINT failure, and run the validation described above. The author directed the design and publication scope.


src/powercontext/server/authz/models.py
```diff
@@ -0,0 +1,465 @@
...
+        scope_id: str,
+        *,
+        family: str,
+        artifact_id: str,
+        selector: MemoryEntrySelector | None = None,
+    ) -> ResourceRef:
+        return cls(
+            type=AccessResourceType.ARTIFACT,
+            scope_id=scope_id,
+            identity=ArtifactIdentity(family=family, artifact_id=artifact_id),
```
PsiACE, line 211: `ResourceRef` drops the revision, so a grant also covers adjacent and future revisions. Is this intentional despite RFC 1396 requiring exact-revision access?

src/powercontext/server/app.py
```diff
@@ -1444,36 +2056,94 @@ async def finalize_handoff(
...
+            memory = citation.memory_citation
+            resource = ResourceRef.artifact(
+                scope_id,
+                family=memory.memory_ref.family,
+                artifact_id=memory.memory_ref.artifact_id,
+                selector=MemoryEntrySelector(entry_id=memory.entry_id),
+            )
+        else:
+            return False
+        decision = await access.check(principal, action, resource, context=context)
```
PsiACE: Citation checks require separate scope or artifact access, so a Handoff receiver cannot inspect its evidence. Is this intended despite the manifest-only access defined by the RFC?

src/powercontext/server/settings.py
```diff
@@ -122,19 +122,33 @@ def validate_path(cls, value: str) -> str:
...
             raise MissingBearerTokenError("Bearer token is required when authentication is enabled")  # noqa: TRY003
+        if self.provider != "static-bearer" and self.token is not None:
+            raise ValueError("AUTH_TOKEN is only valid for AUTH_PROVIDER=static-bearer")  # noqa: TRY003
         return self
 
 
+class AccessControlConfig(BaseModel):
+    """Server security profile and deployment-local authorization identity."""
+
+    mode: Literal["disabled", "enforced"] = "disabled"
```
PsiACE, line 148: `legacy-static-admin` was removed, so existing static-token configurations fail validation. Is this compatibility break intentional?

src/powercontext/http/_generated/models.py
```diff
@@ -21,6 +21,397 @@
...
+        extra="forbid",
+    )
+    type: Type4
+    entry_id: Annotated[StrictStr, Field(max_length=128, min_length=1, pattern="^[\\x21-\\x7E]+$")]
+
+
+class AccessArtifactIdentity(BaseModel):
+    model_config = ConfigDict(
+        extra="forbid",
+    )
```
thunguo, line 156: This collapses an exact Artifact or Memory grant into a version-independent logical resource: ArtifactIdentity has no revision, and MemoryEntrySelector has no entry version, so the same binding authorizes adjacent and future content. The new E2E test at tests/e2e/test_access_control_http.py:178-181 demonstrates the leak by allowing an exact receiver to read a later latest revision. Please include the positive Artifact revision (and Memory entry_version_id) in the authorization identity and all persistence/wire representations. If ownership is intentionally logical, use a separate owner identity rather than making exact access bindings versionless, and add negative tests for adjacent and future versions.

src/powercontext/server/app.py
```diff
@@ -2072,19 +3196,325 @@ def _add_route(
...
+        return ((AccessAction.SCOPE_READ, ResourceRef.scope(scope_id)),)
+    resource = (
+        _artifact_resource(payload, "revision", family="handoff")
+        if selection == "exact"
+        else ResourceRef.artifact(scope_id, family="handoff", artifact_id="handoff")
+    )
+    return (
+        (AccessAction.ARTIFACT_READ, resource),
+        (AccessAction.HANDOFF_EVIDENCE_INSPECT, resource),
+    )
```
thunguo, line 3692: selection=latest falls through to the exact-resource branch and is authorized with artifact.read plus handoff.evidence.*. A principal holding only a narrow Handoff grant can therefore resolve a newer revision that was never granted; the added E2E test currently asserts this behavior. Please require scope.read for latest and reserve the exact-resource resolver for selection=exact, then add a regression test proving that an exact-only receiver gets 403 for latest.

openapi/powercontext.yaml
```diff
@@ -982,6 +1082,80 @@ paths:
...
           $ref: "#/components/responses/Unauthorized"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        "422":
+          $ref: "#/components/responses/InvalidRequest"
+        "503":
+          $ref: "#/components/responses/Unavailable"
+        "500":
+          $ref: "#/components/responses/InternalError"
+  /v1/skills/publication-targets/list:
```
thunguo: Both publication endpoints reuse exact_skill_access, which resolves only to artifact.read. This lets an artifact.viewer perform a host-side write through /v1/skills/publish; Dashboard publish/unpublish has the same missing mutation permission, because its additional server.observe check is also read-only. Please add the RFC-defined skill.publish action and skill.publisher role, and require artifact.read plus skill.publish on the same exact revision for target listing, publish, and unpublish. Add negative tests proving that artifact.viewer can retrieve the Skill but cannot mutate any publication target.

src/powercontext/server/authz/service.py
```diff
@@ -0,0 +1,1447 @@
...
+                decisions.append(AccessDecision(True, "authenticated", revision))
+                continue
+            owner = (
+                await self._repository.get_artifact_owner(request.resource)
+                if request.resource.type is AccessResourceType.ARTIFACT
+                else None
+            )
+            if request.resource.type is AccessResourceType.ARTIFACT and owner is None:
+                decisions.append(AccessDecision(False, "artifact-owner-pending", revision))
+                continue
```
thunguo, line 377: Every Artifact decision returns artifact-owner-pending before evaluating server, scope, or exact bindings when no owner row exists. Artifacts created while access control was disabled have no such row and there is no startup backfill; I reproduced an enforced restart where even the bootstrap server.admin receives 503 for an existing Handoff. Creation also commits domain state before writing the owner, so a relationship-store failure leaves committed data with no in-product recovery path. Please add an idempotent, audited migration/preflight and make owner establishment transactional or durably recoverable (for an external store, for example via reservation/outbox), with disabled-to-enforced and failure-injection tests.

src/powercontext/server/authz/authzen.py
```diff
@@ -0,0 +1,202 @@
...
+            response.raise_for_status()
+            value = response.json()
+        except (httpx.HTTPError, ValueError, TypeError) as error:
+            raise AccessUnavailableError from error
+        if not isinstance(value, Mapping):
+            raise AccessUnavailableError
+        return value
+
+
+def _access_request(request: AccessRequest) -> dict[str, object]:
```
thunguo, line 117: AccessRequest.context contains trusted actor and subject_groups facts, but this adapter silently drops both. An external PDP therefore cannot enforce group membership or on-behalf-of actor constraints and can disagree with Builtin/Casbin for the same authenticated request. Please serialize these facts in a versioned, namespaced AuthZEN context (including the issuer-qualified identity), or reject configurations that claim unsupported capabilities, and add payload-level and cross-provider conformance tests.

src/powercontext/server/authz/models.py
```diff
@@ -0,0 +1,465 @@
...
+
+        return self.id
+
+
+AccessSubjectRef: TypeAlias = PrincipalRef | GroupRef
+
+
+@dataclass(frozen=True, slots=True)
+class ArtifactIdentity:
+    """Version-independent identity of one logical Artifact."""
```
Kairo-J, line 139: This identifier does not include version information. Consequently, granting access to version 1 also grants access to version 2 and subsequent versions of the same artifact.

Is this the intended behavior? If not, please include version information in the resource identifier and add a test to ensure that access to other versions is denied.

openapi/powercontext.yaml
Teingi: This is also intentional in the first version. Host-local Skill publish/unpublish is modeled as a projection operation that does not mutate the source Skill identity or Revision. Its selected boundary is `server.observe` plus logical Skill `artifact.read`; remote publication instead requires `scope.admin` plus logical Skill read. RFC 1396 documents those requirements in the trust invariants and Artifact Family operation table. We are therefore not adding a separate `skill.publish` action or `skill.publisher` role in this RFC.

src/powercontext/server/app.py
Teingi: This is intentional for a logical Handoff grant. Both `selection=exact` and `selection=latest` may read existing and future Revisions of the same logical Handoff; `selection=prepared` still requires `scope.read`, and the grant does not expose another Handoff or any parent-scope collection. RFC 1396 invariant 7 and the Handoff operation table specify the same logical-resource requirements for exact and latest selection. I will keep `latest` on the narrow logical-Handoff path.

src/powercontext/server/authz/models.py
Teingi: Yes, this is the intended behavior. Access is granted to the logical Artifact identity across its existing and future Revisions, while write, lifecycle, and next-Revision operations remain restricted to the owner. RFC 1396 explicitly excludes Revision fields from the Access Resource and Binding, and the acceptance tests cover future-Revision visibility. I will keep this identifier version-independent.

src/powercontext/http/_generated/models.py
Teingi: This is intentional. The authorization object is the logical share unit, not an exact-Revision grant: an Artifact uses `{family, artifact_id}`, and a Memory entry adds only its logical `entry_id` selector. Revision and `entry_version_id` remain business-selection fields and deliberately do not enter the Binding. The Binding therefore covers existing and future versions of the same logical identity, while all mutations remain owner-only. RFC 1396 now states this explicitly in “Sharing other Artifact Families,” invariants 7 and 10, and the Artifact Family operation requirements. The E2E assertion protects that intended behavior, so I will keep the authorization identity version-independent.

src/powercontext/server/authz/authzen.py
Teingi: Fixed in 82572ef3. The AuthZEN adapter now forwards the trusted actor and subject_groups under context.powercontext for both point and batch evaluations. The payload-level test pins the wire shape, and the conformance test verifies that AuthZEN agrees with Builtin and Casbin for group-based decisions. I intentionally did not add a schema version because this contract has not been released yet. Identities use the deployment-wide opaque IDs already normalized by the Authentication Provider, so there is no separate caller-supplied issuer field. The bilingual RFC now documents this boundary.

PsiACE (commented):
I noticed two behaviors in `resources/list`:

- An Experience or Skill created through the base Artifact API can be shared and read by the grantee, but it does not appear in `/v1/access/resources/list` because discovery only scans approved candidates.
- A `server.admin` calling `/v1/access/resources/list` for `scope.admin` Scopes receives a 503 because the server-level parent constraint is rejected, even though this action/resource combination is accepted by validation.

Could you confirm whether these are intended limitations for this version? Other than these, LGTM.

Zxf-xufeng (approved):
LGTM

Issue #1363, opened 2026-08-26
Tracking Issue: deliver supported Source ingestion


PsiACE:
## Goal

Track delivery of supported Source ingestion after #1240 defines the Source identity and materialization contract.

Parent roadmap: #1352.

This issue coordinates the accepted Source contract, one supported document integration, historical prompt import, and any derived processing those deliveries require. It does not redefine the public Source model or API.

## Current baseline

PowerContext currently accepts idempotent captured text through `ContentSource` and `POST /v1/sources/content`. #1240 decides whether this remains the common capture boundary for external integrations or one Source implementation alongside others.

#1300 separately tracks explicit, default-off import of historical user prompts from one Agent host. It does not import full sessions or restore an executable Agent session.

## Tracking

- [ ] Complete #1240 and record the accepted Source identity and materialization decisions in an RFC.
- [ ] After #1240 is accepted, select one supported document Source integration. The bounded validation from #1240 may graduate only if it meets the support criteria below; otherwise, track the supported integration in a separate child issue.
- [ ] Deliver initial import and subsequent provider-supported change handling through the accepted Source semantics.
- [ ] Complete #1300 for explicit historical user-prompt import from one supported Agent host.
- [ ] Decide whether either delivered integration requires asynchronous derived processing. If it does, create a bounded child issue for status, retries, idempotency, failure isolation, and resource limits. If it does not, record that no background processing is required.
- [ ] Add maintained setup and troubleshooting documentation, plus tests for each supported integration.

## Completion criteria

- The RFC from #1240 is accepted, and the public API, code, tests, and maintained documentation describe the same Source contract.
- At least one document integration is supported as a product integration rather than retained as sample or validation code.
- Repeated observation is idempotent. Updates and other provider-supported changes preserve exact provenance according to the accepted contract.
- Users can see whether ingestion succeeded and which items failed.
- The external system remains authoritative. Read permissions, retention, deletion, and disconnect behavior are documented.
- Historical prompt import remains an explicit Source import. It does not claim session restore, Handoff, Task Outcome, or full transcript support.
- Product status and documentation distinguish Source acceptance from completion of any later derived processing.

## Boundaries

- This issue does not select a document provider before #1240 settles the contract.
- #1240 owns the Source design and bounded validation. Do not start a parallel Source pilot without coordinating with its owner.
- Provider discovery, synchronization, credentials, checkpoints, retries, and provider-specific change handling remain connector responsibilities.
[...]

PR #1400, opened 2026-08-30
feat(sources): add Source Definition and Connector contracts

### LINKED ISSUE #1240: research: define the Source integration shape
## Objective

Define a practical Source integration shape for external systems.

An external object may change while evidence used by an Artifact must remain exact. Research how a stable logical Source identity relates to an immutable observation or revision. Capture, Ref, and any hybrid form describe how that observation is materialized.

Provider connectors own discovery, synchronization, checkpoints, retries, credentials, and provider-specific change handling. This issue covers the Source values they produce and how PowerContext accepts, reads, stores, and cites those values.

Record the chosen shape in an RFC before changing the public Source model or API. The RFC records the design decision. Maintained docs, OpenAPI, and code describe current behavior.

## Current baseline

Core provides typed Source adapters and supports captured and referenced materialization.

The builtin Runtime provides `ContentSource` and `POST /v1/sources/content` as an idempotent captured-text path. Reusing an identity with different content is a conflict. Runtime assembly, Server capabilities, and evidence projection currently handle `ContentSource` explicitly.

The research should decide whether `ContentSource` and the content API are the common capture boundary for external integrations or one concrete Source implementation alongside other typed Sources.

## Research references

- [ ] Git repositories and GitHub-like services
- [ ] Notion
- [ ] Slack
- [ ] Linear
- [ ] Relevant memory and context projects

These are reference cases, not committed implementation targets.

## Tracking

- [ ] Audit the current Source model, `ContentSource`, persistence, Runtime assembly, HTTP API, and evidence projection.
- [ ] Define the relationship between logical Source identity and immutable observation or revision identity.
- [ ] Compare Capture, Ref, and possible hybrid materialization.
- [ ] Define the minimum provenance required for an exact `SourceRef`.
- [ ] Decide the role of `ContentSource` and `POST /v1/sources/content`.
- [ ] Define how additional Source types register with persistence, Runtime, transport, and Artifact pipelines.
- [ ] Draft and review the RFC.
- [ ] Implement the accepted improvements.
- [ ] Validate the shape with one bounded Source integration.

Target completion is around 2026-09-01. The validation integration does not need to be a complete production connector.

This work does not design a full connector framework or move provider synchronization into the Source model.

PsiACE:

Closes #1240.
Closes #1388.
Its author is credited as a co-author.
Relates to #1363.

# Rationale for this change

External objects can change, but Artifact lineage must retain exact evidence. PowerContext needs a typed Source contract while provider access and synchronization remain in independent workers.

- Define Source identity, observations, materialization, named projections, and Connector lifecycle in an English and Chinese RFC.
- Add Source Definition registration, worker-materialized observation ingestion, durable receipts, and checkpoint comparison.
- Preserve existing `ContentSource` behavior.
- Include `powercontext-connector-opendal` as an evaluation-only reference integration. It tests the contract with UTF-8 text files but does not make files a normative Source type or complete the supported integration tracked by #1363.

The remote Source ingestion API and client are new. Existing `ContentSource` callers remain compatible. The OpenDAL reference integration is opt-in and is not a product support commitment.

# How was this change tested?

- `make check`
- `make test` (`979 passed`, `10 skipped`)
- `make contract-test` (`29 passed`)
- `git diff --check`
- `make docs-test` is currently blocked by the existing HTTP API link in `docs/en/modules.md`.

# AI usage statement

OpenAI Codex with GPT-5 was used for repository analysis, implementation, tests, and documentation. The maintainer reviewed the design and changes.


Teingi (commented):
I found four issues in the current implementation.

src/powercontext/builtin/runtime/relational.py
```diff
@@ -850,6 +946,67 @@ def _validate_experience_plans(
...
+                raise InvalidSourceObservationError(
+                    "text-evidence",
+                    "source identity does not match the observation envelope",
+                )
+
+
+def _json_schema_validator(name: str, schema: Mapping[str, Any]) -> Validator:
+    try:
+        Draft202012Validator.check_schema(schema)
+        return Draft202012Validator(schema)
```
Teingi: Please disable remote schema retrieval here. `Draft202012Validator(schema).validate(...)` dereferences attacker-controlled HTTP `$ref`s by default; I reproduced a request to a loopback schema server through this path. That makes source ingestion an SSRF primitive. Use a no-retrieval registry and reject non-local `$ref`/`$dynamicRef` values during manifest registration, with a regression test asserting that validation makes no network request.

src/powercontext/builtin/runtime/relational.py
```diff
@@ -850,6 +946,67 @@ def _validate_experience_plans(
         )
 
 
+def _validate_source_definition_manifest(manifest: SourceDefinitionManifest) -> None:
+    if len(manifest.model_dump_json(by_alias=True).encode()) > 64 * 1024:
+        raise InvalidSourceDefinitionError(type(manifest), "manifest", "must not exceed 64 KiB")
+    try:
+        BUILTIN_SOURCE_REGISTRY.definition_for_name(manifest.name)
```
Teingi: This only reserves built-in names, while persistence later chooses between a registered source class and `SourceObservation` solely by `source_type`. With an injected custom definition named `review-collision`, a remote v2 manifest under the same name registered successfully; submission then persisted the row and failed with `InvalidStoredPayloadError`, and subsequent `sources.list()` calls kept failing. Reserve names from the active registry and persist an explicit representation or manifest discriminator so configuration changes cannot reclassify stored rows.

src/powercontext/builtin/persistence/sources.py
```diff
@@ -73,9 +73,12 @@ async def add(
         """Add one stable Source or return an identical existing capture."""
 
         _require_identity("scope_id", scope_id, MAX_SCOPE_ID_LENGTH)
-        adapter = self._adapter_for_value(source)
-        ref = SourceRef(source_type=adapter.name, source_id=source.name)
-        payload = dump_model(source, kind="source", name=adapter.name)
+        if isinstance(source, SourceObservation):
```
Teingi, line 87: Remote observations can contain arbitrary JSON objects, but this path still decides idempotency by comparing unsorted serialized bytes below. I submitted equal observations whose only difference was `{"a":1,"b":2}` versus `{"b":2,"a":1}`; the models compared equal, but the retry returned `SourceConflictError`, so the connector could not safely resume and commit its checkpoint. Compare canonical JSON (for example RFC 8785) or decoded model equality, and add a reordered nested-object regression test.

integrations/opendal/src/powercontext_connector_opendal/connector.py
```diff
@@ -0,0 +1,301 @@
...
+
+    def __init__(
+        self,
+        filesystem: _FsspecFileSystem,
+        *,
+        source_namespace: str,
+        root: str = "",
+        patterns: Sequence[str] = _DEFAULT_PATTERNS,
+        max_files: int = 10_000,
+        max_file_size: int = 2 * 1024 * 1024,
```
Teingi: The default 2 MiB raw-file limit is not compatible with the server's 4 MiB observation limit because file content is serialized once in the source payload and again in the text-evidence projection. A 2 MiB ASCII file produced a 4,195,493-byte observation and was rejected by the server. Budget the final serialized observation (including JSON escaping and projections), lower the safe default, and surface oversize files as a rejected item instead of aborting the run.

Teingi (approved):
LGTM

PR #1345, opened 2026-08-25
docs(rfc): define scope organization and agent integration

PsiACE:
# Which issue or RFC does this PR close?

Refs #1219 #1238 .

Close #1238

# Rationale for this change

The current Scope model does not clearly separate ownership, hierarchy, sharing, delivery, and reporting. This RFC defines how these concerns work together without adding another Project identity.

# What changes are included in this PR?

- Add the RFC in English and Chinese.
- Define Scope ownership, Parent, Context References, Artifact publication, and observation selection.
- Describe Agent binding, Agent handoff, Dashboard aggregation, and Handoff Report behavior.
- Define how integrations should allocate Scope IDs and remove the dependency on `project_id`.

# Are there any user-facing changes?

No. This PR only adds documentation. It does not change runtime behavior, public APIs, or persisted data.

The RFC describes proposed API and data model changes for later implementation.

# How was this change tested?

```text
make docs-test
git diff --check
```

The commit hooks also passed.

# AI usage statement

Used Codex with GPT-5 to inspect the repository and help draft and edit the RFC.

PR #1401, opened 2026-08-30
feat: implement scope organization runtime for RFC #1345

PsiACE:

Implements the runtime design in #1345.

# Rationale for this change

Scope is the durable boundary for memory ownership, isolation, sharing, and reporting. The runtime and Agent integrations need to use the same server-owned Scope model instead of Project, Workstream, or client-generated Scope identifiers.

- Adds durable Scope hierarchy, metadata, default selection, and session or workspace bindings.
- Adds direct Context References and exact cross-Scope Artifact References with provenance.
- Uses Scope selection for context reads, statistics, Handoff Reports, and artifact publication.
- Removes the Project and Workstream report model and its compatibility paths.
- Updates bundled Agent integrations to resolve, reuse, and derive server-owned Scopes.
- Updates HTTP, MCP, generated clients, configuration, documentation, and acceptance coverage.
- Makes the E2E harness provision real Scopes before running Agent workflows.

Yes. This is a breaking change:

- Scope replaces Project and Workstream as the organization and reporting boundary.
- Scope IDs are generated by the service; integrations must resolve or create a Scope before writing.
- Context and report queries select all visible Scopes, one exact Scope, or a Scope subtree.
- No compatibility or migration period is included.

# How was this change tested?

- `make check`
- `make test`
- `make contract-test`
- `make docs-test`
- `make harness-check`
- GitHub Actions on Python 3.11, 3.12, 3.13, and 3.14
- SQLite and OceanBase acceptance scenarios

# AI usage statement

OpenAI Codex was used for code analysis, implementation, tests, and documentation editing.


src/powercontext/builtin/runtime/application.py
```diff
@@ -336,26 +365,103 @@ async def prepare(self, request: PrepareContextRequest, /) -> PreparedContext:
 
     async def _prepare(self, request: PrepareContextRequest, /) -> PreparedContext:
         builder = PreparedContextBuilder()
+        scope_ids = [self.scope_id]
+        if self._runtime.scopes is not None:
+            try:
+                scope = await self._runtime.scopes.get(self.scope_id)
+            except ScopeNotFoundError:
+                pass
```
Teingi: Swallowing `ScopeNotFoundError` makes an unregistered Scope behave like a valid legacy partition. On this head, capturing a Source under `project:orphan` succeeded, but `list_scopes` and an `all` selection contained only the default Scope, leaving the durable data invisible to Statistics, Dashboard, and Handoff Report. Please validate every current-scope data-plane operation through `ScopeApplication` (with an explicit migration path for legacy partitions) instead of continuing after this exception.

integrations/codex/plugins/powercontext/hooks/bind_tools.py
```diff
@@ -0,0 +1,148 @@
...
+            _allow({
+                "explicit_scope_id": settings.scope_id,
+                "binding_keys": binding_keys(cwd, session_id=session_id),
+            })
+            return 0
+        if operation in _HOST_OPERATIONS:
+            _allow(dict(tool_input))
+            return 0
+        if operation not in _SCOPE_BOUND_OPERATIONS:
+            return 0
```
Teingi, line 124: This pass-through leaves 14 OpenAPI operations marked `x-powercontext-scope-mode: current` outside the binding guard, including `prepare_context`, `flush_memory`, and `prepare_handoff`. A `prepare_context` PreToolUse payload with `scope_id: "agent-selected"` produces no hook output, so the model-supplied Scope is neither replaced nor denied. Please derive or validate this set against the generated OpenAPI metadata and add a contract test that every `current` operation is guarded.

src/powercontext/builtin/persistence/artifacts.py
```diff
@@ -240,6 +243,42 @@ async def revisions(
...
+
+        _require_scope(target_scope_id)
+        artifact_type = self._artifact_type(source.family)
+        ref = ArtifactRef(family=source.family, artifact_id=target_artifact_id, revision=1)
+        copied = await self._insert_revision(
+            connection,
+            target_scope_id,
+            artifact_type,
+            ref,
+            source.content,
```
Teingi, line 266: Copying only `source.content` is insufficient for the Memory family. Publishing an exact Memory revision on this head returned a new `memory/pub_*@1`, but reading an entry through that target citation raised `ArtifactNotFoundError`, and the target Scope still had no current Memory. The entry versions/heads and target head integration were not copied. Please make Memory publication family-aware and atomic, or reject it until the complete target state can be created.

src/powercontext/builtin/scope/application.py
```diff
@@ -0,0 +1,291 @@
...
+        self,
+        database: AsyncDatabase,
+        *,
+        repository: ScopeRepository | None = None,
+        id_factory: ScopeIdFactory | None = None,
+    ) -> None:
+        self._database = database
+        self._repository = ScopeRepository() if repository is None else repository
+        self._id_factory = generate_scope_id if id_factory is None else id_factory
+        self._write_lock = asyncio.Lock()
```
Teingi, line 62: This lock only serializes writes within one `ScopeApplication` instance. With two server instances sharing OceanBase, concurrent `A.parent=B` and `B.parent=A` updates can both validate the old graph and commit different rows; the FK and per-row version checks do not detect the resulting cycle. Please serialize relationship mutations at the database level (or use a retryable serializable/locked ancestor check) and cover this race with two application instances.

integrations/codex/plugins/powercontext/hooks/bind_tools.py
PsiACE: Fixed in `056dd9ef`: the hook now guards every OpenAPI `current` and `selection` operation, replaces model-provided scope inputs with the Session binding, and validates complete coverage against generated metadata.

src/powercontext/builtin/scope/application.py
PsiACE: Fixed in `f45cc692`: hierarchy updates now acquire a database-level serialization boundary before validation and CAS, with a two-application concurrency test proving inverse Parent updates cannot both commit.

src/powercontext/builtin/runtime/application.py
PsiACE: Fixed in `325f20d6`: every scope-bound Runtime operation now resolves the Scope through `ScopeApplication` before touching storage, with an E2E regression test proving unknown Scopes fail closed without persisting data.

src/powercontext/builtin/persistence/artifacts.py
PsiACE: Fixed in `5e35afde`: Memory publication is now rejected before any target write with `artifact_publication_unsupported`, and the regression test verifies that no target Artifact, head, or Memory entry is created.

src/powercontext/builtin/runtime/application.py
```diff
@@ -365,31 +395,101 @@ def __init__(self, runtime: BuiltinRuntime, scope_id: str) -> None:
...
+
+        memory_candidates: list[PreparedMemoryCandidates] = []
+        experience_candidates: list[PreparedExperienceCandidates] = []
+        remaining_memory = builder.memory_candidate_limit
+        remaining_experience = builder.experience_candidate_limit
+        for scope_id in scope_ids:
+            memory, experiences = await self._recall_scope(
+                scope_id,
+                request,
+                memory_limit=remaining_memory,
```
Teingi: [P1] Keep referenced scopes eligible for recall

`remaining_memory` and `remaining_experience` are consumed in scope order, with the current scope first. Once it returns the full candidate limit, every context reference is queried with a limit of zero, so a relevant referenced item disappears solely because the current scope has enough lower-quality fillers. Fetch candidates from all referenced scopes before applying the shared reranker/budget, or reserve a per-scope quota.

src/powercontext/builtin/publication.py
```diff
@@ -0,0 +1,232 @@
...
+            if source.family == "memory":
+                raise ArtifactPublicationUnsupportedError(source.family)
+            existing = await self._find_request(connection, request.target_scope_id, request.idempotency_key)
+            if existing is not None:
+                if existing.source != request.source:
+                    raise ArtifactPublicationConflictError(request.idempotency_key)
+                return existing
+
+            content_digest = hashlib.sha256(dump_model(source.content, kind="artifact", name=source.family)).hexdigest()
+            target = await self._artifacts.copy_exact(
```
Teingi: [P1] Update the Experience search index when publishing

`copy_exact` persists the target Experience revision, but this path never updates the target scope's `ExperienceIndex`. On a fresh SQLite database, publication succeeds and an immediate search returns no hit; the same search returns the item after restart rebuilds the index. Update or invalidate the index atomically with publication so the new artifact is visible immediately.

src/powercontext/builtin/publication.py
```diff
@@ -0,0 +1,232 @@
...
+            return await self._publish(request)
+
+    async def _publish(self, request: ArtifactPublicationRequest) -> ArtifactPublication:
+        await self._scopes.get(request.source.scope_id)
+        await self._scopes.get(request.target_scope_id)
+        async with self._database.transaction() as connection:
+            source = await self._artifacts.get(connection, request.source.scope_id, request.source.artifact)
+            if source.family == "memory":
+                raise ArtifactPublicationUnsupportedError(source.family)
+            existing = await self._find_request(connection, request.target_scope_id, request.idempotency_key)
```
Teingi: [P1] Make idempotency atomic across application instances

`_write_lock` only protects one `ArtifactPublicationApplication`. Two instances can both miss `_find_request` and race through the copy/insert; in a fresh SQLite reproduction, one succeeds and the other leaks `IntegrityError` instead of returning the same publication. Reserve or upsert the idempotency key at the database layer and translate the race into the documented replay or conflict result.

docs/en/docs/how-to/configure-pydantic-ai.md
```diff
@@ -60,7 +60,7 @@ export POWERCONTEXT_PYDANTIC_AI_SCOPE_ID=project:example
 | --- | --- | --- |
 | `POWERCONTEXT_PYDANTIC_AI_BASE_URL` | `http://127.0.0.1:8000` | HTTP(S), without credentials, query, or fragment |
 | `POWERCONTEXT_PYDANTIC_AI_TOKEN` | unset | Bare printable token stored as `SecretStr` |
-| `POWERCONTEXT_PYDANTIC_AI_SCOPE_ID` | derived | Non-empty scope, deterministically bounded to 256 characters |
+| `POWERCONTEXT_PYDANTIC_AI_SCOPE_ID` | unset | Existing explicit Server Scope, up to 256 characters; unset selects the Server default |
```
Teingi, line 62: [P2] Keep this example valid under registered Scope semantics

The guide now says explicit IDs must identify an existing Server Scope, but the command above still exports `project:example`. A fresh server rejects that value with `404 scope_not_found`. Either leave the variable unset to use the default Scope or show how to create or resolve a Scope and use the returned `scp_*` ID; please update the equivalent remaining `project:*` examples as well.

src/powercontext/builtin/runtime/application.py
PsiACE: Fixed by recalling every direct Context Reference before applying a round-robin global candidate/output budget; the regression saturates the current Scope and verifies referenced evidence remains eligible.

src/powercontext/builtin/publication.py
PsiACE: Fixed by updating the target Experience projection in the same transaction as the exact copy and publication record, with immediate-search and projection-failure rollback regressions.

src/powercontext/builtin/publication.py
PsiACE: Fixed by rolling back the whole losing transaction and resolving the unique-key winner in a fresh transaction as replay or conflict; a two-application concurrency regression now covers this.

docs/en/docs/how-to/configure-pydantic-ai.md
PsiACE: Fixed by leaving the Pydantic AI Scope override unset by default and replacing all active documentation and integration `project:*` examples with an existing ID returned by `create_scope`.

Teingi (approved):
LGTM

PR #1403, opened 2026-08-31
feat(metrics): expose process runtime metrics

PsiACE:

Related to #1214 and RFC 0046.

## Rationale for this change

PowerContext uses a private Prometheus registry, so the Python client default process and runtime metrics were missing.

Registers the official process, platform, and GC collectors. Adds a behavior test confirming that /metrics exposes process CPU and RSS.

Linux deployments gain standard process and Python runtime metrics. There are no configuration or API changes.

- make check
- make test: 965 passed, 10 skipped
- make docs-test
- Real Server scrape against /metrics

## AI usage statement

Codex with GPT-5 was used for implementation, testing, and PR wording.

### DIFF
--- a/src/powercontext/server/metrics.py
+++ b/src/powercontext/server/metrics.py
@@ -22,7 +22,17 @@ from time import perf_counter
 from typing import Any
 
 from fastmcp.server.middleware import CallNext, Middleware, MiddlewareContext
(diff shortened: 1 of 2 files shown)


Teingi (approved):
LGTM

PR #1418, opened 2026-09-02
feat: add personal Server service lifecycle

Pika-Yalei:

Tracks #1298.

Implements the accepted service-lifecycle portions of RFC #1299. The tracking issue remains open for host-visible integration diagnostics, Windows support, and remaining platform acceptance coverage.

# Rationale for this change

`powercontext server run` is intentionally a foreground process, so closing its terminal or logging in again can leave personal agent integrations without a reachable Server. RFC #1299 defines an opt-in, distribution-owned native per-user service layer that preserves the foreground Server boundary and does not require privilege elevation.

- Adds `powercontext service install`, `status`, and `uninstall` with stable human-readable and JSON state.
- Adds a shared service definition/state model, endpoint preflight, ownership validation, idempotent reconciliation, rollback, executable-drift detection, and an internal one-shot launcher.
- Adds Linux `systemd --user` and macOS current-user LaunchAgent adapters with bounded restart behavior and explicit log locations.
- Keeps Windows explicitly unsupported until a qualified adapter is delivered.
- Reuses the existing Server runner and configuration semantics without changing `powercontext server run` foreground behavior.
- Extends `powercontext doctor` to correlate an optional matching local registration with manager and Server state.
- Updates English, Chinese, and Japanese user documentation and adds focused behavioral tests.

This PR does not implement host-visible `server_unavailable` warnings in agent integrations or a Windows Task Scheduler adapter; those remain separate delivery items under #1298.

Yes. Personal macOS and Linux installations gain the opt-in `powercontext service` command group for persistent background operation. The existing foreground command, HTTP API, configuration defaults, and persisted formats remain compatible. Managed deployments should continue to use their existing container or administrator-owned service manager.

# How was this change tested?

- `make check`
- `make test` — 990 passed, 11 skipped
- Equivalent strict documentation build with `uv run zensical build --clean -s`, plus generated OpenAPI comparison. The existing `make docs-test` wrapper uses GNU `install -D` and is not directly portable to the macOS BSD `install` command.
- `uv run powercontext service status --json` on macOS as a read-only supported/not-installed status check; no real service registration was created on the development machine.

Focused tests cover service definition rendering and tamper detection, install/idempotency/rollback/uninstall behavior, env-file permissions, endpoint conflicts, liveness identification, launcher preflight, CLI JSON output, and doctor output changes. Matching-runner native register/query/start/stop/remove coverage can be reviewed as a follow-up platform acceptance item under #1298.

# AI usage statement

OpenAI Codex (GPT-5.6-sol) was used to implement, review, and test this change. The author reviewed the resulting code and submits it for maintainer review.


src/powercontext/service/cli.py
```diff
@@ -0,0 +1,119 @@
...
+"""User-facing lifecycle commands for the personal PowerContext Server service."""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import Annotated
+
+import typer
+import uvicorn as _server_role_dependency
```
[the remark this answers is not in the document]
Pika-Yalei: 已在 98b8f79 修复：service 命令门控现在导入完整的 powercontext.server.cli，而不是只检查 uvicorn。新增子进程测试，验证缺少 FastAPI 等完整 Server role 依赖时不会暴露该命令。

src/powercontext/service/model.py
```diff
@@ -0,0 +1,240 @@
...
+    def launcher_arguments(self) -> list[str]:
+        arguments = [
+            self.python_executable,
+            "-m",
+            "powercontext.service.launcher",
+            "--endpoint",
+            self.endpoint,
+        ]
+        if self.env_file is not None:
+            arguments.extend(("--env-file", self.env_file.path))
```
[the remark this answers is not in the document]
Pika-Yalei: 已在 98b8f79 修复：原生定义的 launcher 参数始终携带已记录的绝对 data dir，launcher 构造 ServerSettings 时强制应用它，并覆盖 env file 或 ambient POWERCONTEXT_HOME。新增遗漏持久化与覆盖优先级测试。

src/powercontext/service/adapters/launchd.py
```diff
@@ -0,0 +1,185 @@
...
+        return NativeRegistration(RegistrationState.INSTALLED, definition=definition, content=content)
+
+    def render(self, definition: ServiceDefinition) -> bytes:
+        log_dir = Path(definition.data_dir) / "logs"
+        payload: dict[str, Any] = {
+            "Label": self.identifier,
+            "ProgramArguments": definition.launcher_arguments(),
+            "RunAtLoad": True,
+            "KeepAlive": {"SuccessfulExit": False},
+            "ThrottleInterval": 5,
```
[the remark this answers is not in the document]
Pika-Yalei: 已在 98b8f79 修复：launchd 启动器现在用持久化失败状态记录 60 秒窗口内的启动尝试，最多 3 次；达到上限后返回成功状态以终止 KeepAlive 重启循环。手动再次执行 powercontext service install 会清空预算。新增了有限重试回归测试。

src/powercontext/service/adapters/launchd.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def start(self, *, reload_definition: bool) -> None:
+        loaded = self._is_loaded()
+        if loaded and reload_definition:
+            self._run("bootout", self._target)
+            loaded = False
+        if not loaded:
+            registration = self.inspect()
+            if registration.definition is not None:
+                (Path(registration.definition.data_dir) / "logs").mkdir(mode=0o700, parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
Pika-Yalei: 已在 98b8f79 修复：launchd 适配器在写入 plist 和 enable 注册前都会创建 data_dir/logs，不再依赖 start 阶段。测试验证日志目录在写入定义时已经存在。

src/powercontext/service/controller.py
```diff
@@ -0,0 +1,310 @@
...
+            changed = registration.definition != definition
+            manager_before = self._adapter.manager_state()
+            if changed:
+                self._commit_definition(definition, registration.content)
+            else:
+                self._adapter.enable()
+
+            should_restart = changed and manager_before is ManagerState.ACTIVE
+            if should_restart or initial_probe.state is not ProbeState.LIVE:
+                self._adapter.start(reload_definition=changed)
```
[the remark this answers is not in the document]
Pika-Yalei: 已在 98b8f79 修复：当 manager 已 active 但 endpoint unreachable 时，install 会执行受控 restart 并重新探活。新增 active-but-unreachable 恢复测试。

src/powercontext/service/probe.py
```diff
@@ -0,0 +1,79 @@
...
+            pass
+    except OSError:
+        return ProbeResult(ProbeState.UNREACHABLE, f"cannot reach {endpoint}")
+
+    request = Request(  # noqa: S310 - the endpoint is validated as loopback by the service controller.
+        f"{endpoint.rstrip('/')}/health/live",
+        headers={"Accept": "application/json", "User-Agent": "powercontext-service"},
+    )
+    try:
+        with urlopen(request, timeout=timeout) as response:  # noqa: S310
```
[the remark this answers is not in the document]
Pika-Yalei: 已在 98b8f79 修复：环回健康探针改为通过 ProxyHandler({}) 构造无代理 opener，不再读取 HTTP_PROXY、ALL_PROXY 等环境代理。新增了污染代理环境下的真实环回探针回归测试。

src/powercontext/service/adapters/systemd.py
```diff
@@ -0,0 +1,176 @@
...
+
+_METADATA = re.compile(rb"^# X-PowerContext-Metadata: ([A-Za-z0-9_=-]+)$", re.MULTILINE)
+
+
+class SystemdUserAdapter:
+    identifier = "powercontext.service"
+
+    def __init__(self, *, config_home: Path | None = None) -> None:
+        root = config_home or Path(os.environ.get("XDG_CONFIG_HOME", Path.home() / ".config"))
+        self.artifact_path = root / "systemd" / "user" / self.identifier
```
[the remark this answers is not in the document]
Pika-Yalei: 已在 98b8f79 修复：XDG_CONFIG_HOME 仅在非空且为绝对路径时使用；空值和相对路径均回退到 ~/.config。两个场景都有参数化回归测试。

src/powercontext/service/controller.py
```diff
@@ -0,0 +1,310 @@
...
+                final_probe = self._wait_until_live(definition.endpoint)
+                if final_probe.state is not ProbeState.LIVE:
+                    raise ServiceError(  # noqa: TRY003
+                        f"the personal service was registered but did not become live: {final_probe.detail}; "
+                        f"inspect {self._adapter.log_location(definition) or 'the native service logs'}"
+                    )
```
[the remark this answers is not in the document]
Pika-Yalei: 已在 98b8f79 修复：定义提交后的 native start 失败现在携带重新采集的 ServiceStatus，CLI 会在错误后输出 registration、manager、liveness、logs 和 recovery facts。控制器与 CLI 均新增回归测试。

src/powercontext/service/controller.py
```diff
@@ -0,0 +1,328 @@
...
+        except (EnvironmentFileError, OSError, ValidationError) as error:
+            raise ServiceError(  # noqa: TRY003
+                f"invalid personal service configuration: {error}", exit_code=2
+            ) from error
+
+        return ServiceDefinition(
+            ownership=OWNERSHIP_MARKER,
+            definition_version=DEFINITION_VERSION,
+            package_version=version("powercontext"),
+            python_executable=str(Path(sys.executable).resolve()),
```
Kairo-J: Consulting the Python documentation (https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve) reveals that `Path(sys.executable).resolve()` resolves the Python symbolic link within the virtual environment (venv), potentially pointing to the base Python environment where `powercontext` is not installed.

I have reproduced the issue: `.venv/bin/python` resolves to the base Python 3.14 environment, and executing `<resolved-python> -m powercontext.service.launcher --help` raises a `ModuleNotFoundError: No module named 'powercontext'`.
Although the full test suite still passes (1,000 passed, 11 skipped), this indicates that the current tests do not cover this specific scenario. I believe it is necessary to add a test case for this scenario.

Teingi (commented):
I found five remaining issues on the current head.

src/powercontext/service/launcher.py
```diff
@@ -0,0 +1,163 @@
...
+        path=options.failure_state,
+        limit=options.failure_limit,
+        window_seconds=options.failure_window_seconds,
+    )
+    if not budget.begin_attempt():
+        logger.error("Personal service retry budget is exhausted; run `powercontext service install` to retry")
+        return 0
+
+    try:
+        with server_settings_context(env_file=options.env_file, data_dir=options.data_dir) as settings:
```
Teingi: P1: The native launcher reopens `--env-file` by path without verifying the identity or protection recorded at installation. I reproduced replacing the file and changing it to mode 0644: the launcher still loaded the replacement configuration and exited successfully, while a permission-only change still reports the definition as `current`. In a parent directory writable by another local user, this allows configuration injection into the victim-owned service. Pass the recorded identity to the launcher and validate `O_NOFOLLOW`/`fstat` owner, mode, device, inode, size, and mtime before parsing the same file descriptor.

src/powercontext/service/adapters/launchd.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def support(self) -> tuple[SupportState, str]:
+        if sys.platform != "darwin":
+            return SupportState.UNSUPPORTED, "LaunchAgents are available only on macOS"
+        if shutil.which("launchctl") is None:
+            return SupportState.UNSUPPORTED, "launchctl is not installed or is not on PATH"
+        result = self._run("print", self._domain, check=False)
+        if result.returncode != 0:
+            return SupportState.UNSUPPORTED, "the current launchd user domain is unavailable"
+        return SupportState.SUPPORTED, "the launchd user domain is available"
```
Teingi, line 88: P1: This advertises the macOS adapter as supported even though this PR's validation notes that no real LaunchAgent registration was created and the repository has no macOS lifecycle job. RFC 1299 requires a matching-platform register/query/start/stop/remove test before an adapter reports `supported` or the platform is advertised. Please add that native acceptance coverage here, or keep this adapter unsupported and remove the GA documentation until it exists.

src/powercontext/server/cli.py
```diff
@@ -78,49 +76,38 @@ def run(
...
-                application,
-                host=settings.http.host,
-                port=settings.http.port,
-            )
-        finally:
-            tracing.shutdown()
+    try:
+        with server_settings_context(host=host, port=port, env_file=env_file) as settings:
+            _run_configured_server(settings)
+    except (EnvironmentFileError, OSError) as error:
```
Teingi: P2: This `try` now covers `_run_configured_server()`, so runtime/startup `OSError`s are converted into an `--env-file` usage error even when no env file was supplied. Reproducing with the runner raising `OSError("simulated runtime startup failure")` returns exit 2 and prints `Invalid value for --env-file`. Limit this translation to environment-file loading/settings construction; operational runner failures should retain their normal failure classification.

src/powercontext/service/adapters/launchd.py
```diff
@@ -0,0 +1,207 @@
...
+    def remove(self) -> None:
+        self.artifact_path.unlink(missing_ok=True)
+
+    def manager_state(self) -> ManagerState:
+        result = self._run("print", self._target, check=False)
+        if result.returncode != 0:
+            return ManagerState.INACTIVE
+        if "state = running" in result.stdout:
+            return ManagerState.ACTIVE
+        if "state = exited" in result.stdout:
```
Teingi: P2: The launcher intentionally exits 0 when the endpoint already hosts a live PowerContext Server, but this maps every launchd `state = exited` result to `FAILED`. With `state = exited` and `last exit code = 0`, `manager_state()` still returns `failed`, so status/doctor reports an expected duplicate-prevention exit as a manager failure. Parse the last exit status (clean exit -> inactive, nonzero -> failed) and cover the real launchctl output shape.

src/powercontext/service/controller.py
```diff
@@ -0,0 +1,328 @@
...
+        support, detail = self._adapter.support()
+        if support is SupportState.UNSUPPORTED:
+            raise ServiceError(detail)
+        with _service_lock(self._adapter.lock_path):
+            registration = self._adapter.inspect()
+            if registration.state is RegistrationState.NOT_INSTALLED:
+                return self.status()
+            self._require_mutable_registration(registration)
+            self._adapter.stop()
+            self._adapter.disable()
```
Teingi: P2: If `disable`, `remove`, or the final reload fails after `stop()` succeeds, this propagates a generic error without reporting whether the registration still exists or how to recover. RFC 1299 requires incomplete removal to report the remaining artifact and a native recovery command. Re-inspect and attach the remaining service facts/recovery action on each partial-failure path, with regression coverage for disable/remove/reload failures.

src/powercontext/service/controller.py
```diff
@@ -0,0 +1,328 @@
...
+        if initial_probe.state is ProbeState.CONFLICT:
+            raise ServiceError(  # noqa: TRY003
+                f"refusing to install over another listener: {initial_probe.detail}"
+            )
+
+        with _service_lock(self._adapter.lock_path):
+            registration = self._adapter.inspect()
+            self._require_mutable_registration(registration)
+            changed = registration.definition != definition
+            manager_before = self._adapter.manager_state()
```
Kairo-J: This operation only validates the expected plist or unit file; it does not verify whether the service currently registered under the same launchd/systemd identifier actually belongs to PowerContext. Consequently, a service sharing the same name—but not created by this process—could be erroneously stopped and replaced.

Before stopping or restarting the service, it may be necessary to verify its loaded configuration path and `ExecStart` parameters via the native manager; if ownership cannot be confirmed, the operation should be safely aborted and an error reported.

src/powercontext/cli/system.py
```diff
@@ -1054,11 +1055,83 @@ def run_diagnostics(*, server_url: str) -> dict[str, Diagnostic]:
...
+    if not is_loopback_host(parsed.hostname):
+        return {}
+
+    from powercontext.service.controller import ServiceController
+    from powercontext.service.model import (
+        DefinitionState,
+        ManagerState,
+        RegistrationState,
+        SupportState,
+    )
```
Alanxtl, line 1087: `_local_service_diagnostics()` calls `ServiceController().status()` first. Since `status()` probes the registered endpoint, the code only compares it with the requested `server_url` afterward.

For example:

- Registered endpoint: `127.0.0.1:8000`
- Command: `powercontext doctor --server-url http://127.0.0.1:9000`

The implementation still probes port `8000` first. If that probe fails, it may incorrectly return `service_support: degraded`. This violates the RFC requirement that a different loopback endpoint should only perform liveness/readiness checks for the requested target.

Please inspect the registration metadata first, compare normalized endpoints, and only query the manager and probe the endpoint after confirming they match. Equivalent loopback addresses such as `localhost` and `127.0.0.1` should also be normalized consistently.

src/powercontext/service/launcher.py
```diff
@@ -0,0 +1,163 @@
...
+    parser.add_argument("--failure-window-seconds", type=float, default=0)
+    options = parser.parse_args(arguments)
+    budget = _FailureBudget(
+        path=options.failure_state,
+        limit=options.failure_limit,
+        window_seconds=options.failure_window_seconds,
+    )
+    if not budget.begin_attempt():
+        logger.error("Personal service retry budget is exhausted; run `powercontext service install` to retry")
+        return 0
```
Alanxtl: `begin_attempt()` returns `False` for all of the following cases:

- The retry budget is exhausted
- The state file is corrupted
- The state file is unreadable
- The state file is unwritable
- The configuration is invalid

However, `main()` returns exit code `0` for every `False` result. As a result, launchd may consider the launch successful and stop retrying, even though the Server never started.

Please distinguish between an exhausted retry budget and a retry-state persistence failure. Exhaustion may exit cleanly, but corrupted or otherwise unreadable/unwritable state should return a non-zero exit code. Please also add regression tests for these cases.

Alanxtl (commented):
pls refer to the rfc and add Linux matching-runner and life cycle test

src/powercontext/service/adapters/launchd.py
Kairo-J, line 88: Before reverting to the SUPPORTED state, the adapter must pass native register, query, start, stop, and remove tests within the corresponding operating system runner environment.

Please add the necessary Linux/macOS acceptance tests, or keep the status as UNSUPPORTED until the adapter has passed the relevant verification.

This requirement also applies to SystemdUserAdapter.support().

src/powercontext/service/adapters/launchd.py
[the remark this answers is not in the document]
Kairo-J: The retry budget is initialized only after module imports and parameter parsing are complete.

Since the KeepAlive mechanism restarts the process if it exits with a non-zero status code, failures occurring before `budget.begin_attempt()` is called do not update the failure status file; consequently, the number of retries is not constrained by the budget.

I reproduced a scenario involving four consecutive exits caused by dependency import failures: each run terminated with status code 1, and no failure status file was generated. The previously observed `.resolve()` issue may also be contributing to this behavior.

Please consider implementing retry limits within a stable external launcher or avoiding the use of an unrestricted KeepAlive mechanism.

src/powercontext/service/adapters/launchd.py
```diff
@@ -0,0 +1,207 @@
...
+        state, content, detail = inspect_artifact(self.artifact_path)
+        if state is not RegistrationState.INSTALLED or content is None:
+            return NativeRegistration(state, content=content, detail=detail)
+        try:
+            payload = plistlib.loads(content)
+            definition = _definition_from_payload(payload)
+            expected = plistlib.loads(self.render(definition))
+        except (TypeError, ValueError, plistlib.InvalidFileException) as error:
+            return NativeRegistration(RegistrationState.INVALID, content=content, detail=str(error))
+        if definition.ownership != OWNERSHIP_MARKER or payload != expected:
```
Kairo-J: inspect() requires installed plists/units to match the current renderer exactly. Following legitimate template changes, old PowerContext definitions are marked as INVALID, causing automatic reconciliation to be erroneously rejected.

Verify ownership separately; classify old definitions owned by the current entity as STALE, while reserving the INVALID status for external or modified definitions.

The same issue exists in SystemdUserAdapter.inspect().

src/powercontext/service/adapters/systemd.py
```diff
@@ -0,0 +1,185 @@
...
+    def manager_state(self) -> ManagerState:
+        result = self._run("show", "--property=ActiveState", "--value", self.identifier, check=False)
+        if result.returncode != 0:
+            return ManagerState.INACTIVE
+        state = result.stdout.strip()
+        if state == "active":
+            return ManagerState.ACTIVE
+        if state == "failed":
+            return ManagerState.FAILED
+        if state in {"inactive", "activating", "deactivating"}:
```
Kairo-J: The "activating" state encompasses scenarios involving automatic restarts. Mapping it to INACTIVE causes `stop()` to skip the `systemctl stop` operation during uninstallation. Since `systemctl disable` neither stops the unit nor cancels queued restart tasks, the service might continue running even after the unit file is deleted.

However, I do not currently have the means to reproduce this specific scenario; you might want to reproduce it first to confirm whether the issue exists.

src/powercontext/service/controller.py
Pika-Yalei: Agreed, and thank you for the concrete reproducer. Resolving `sys.executable` is incorrect here because the service definition needs to preserve the virtualenv entry point, not the base interpreter to which that entry happens to be linked.

I’ll change definition generation to store an absolute but non-dereferenced executable path, and apply the same normalization in `definition_state()` so definitions created by the current implementation are detected as stale and can be reconciled by a subsequent install.

I’ll cover this with a POSIX symlink fixture, a subprocess test that executes the exact recorded argv and imports the launcher from the venv, a legacy resolved-definition stale test, and an install reconciliation test. I’ll leave this discussion open until the implementation and regression tests are pushed.

src/powercontext/service/launcher.py
Pika-Yalei: I’ll extend the recorded identity to include owner and mode, open the file with no-follow semantics, validate the regular-file type, owner, owner-only permissions, and recorded device/inode/size/mtime using `fstat()`, then parse the content read from that same descriptor. A second `fstat()` after reading will reject mutation during the read. Installation and launch will use the same protected loader, and the definition version will be bumped so existing definitions are reconciled.

Regression coverage will include a valid owned 0600 file, symlink and atomic replacement, chmod-only drift, owner/identity drift, mutation during read, and verification that the Server runner is never entered for rejected input. The tests will also verify that a replacement path is not reopened after the original descriptor has been validated.

src/powercontext/service/controller.py
Pika-Yalei: Implemented in 81a3ed9. Service definitions now record os.path.abspath(sys.executable) without dereferencing the virtualenv symlink, and definition_state applies the same normalization so definitions produced with the resolved base interpreter are stale and reconcilable. Regression coverage includes the POSIX symlink case, an exact recorded-argv subprocess import, stale detection, reconciliation, a missing interpreter, and the non-symlink control case.

src/powercontext/service/launcher.py
Pika-Yalei: Implemented in 81a3ed9. EnvironmentFileIdentity now records device, inode, size, mtime, owner UID, and mode. The protected loader opens with no-follow and close-on-exec semantics, validates a regular owner-only file with fstat before and after reading, and parses the bytes from that same descriptor. Installation and launch both use this loader, and the definition version was bumped. Tests cover 0600 success, mode and owner drift, symlinks, atomic replacement, replacement after open, mutation during read, stale detection, and refusing to enter the Server runner.

src/powercontext/service/adapters/launchd.py
Pika-Yalei: Implemented and verified on a matching runner. The native-service workflow now has a macOS LaunchAgent job that builds the wheel in a fresh virtualenv and exercises real install, query, command matching, stop, start, uninstall, foreign-registration rejection, bounded retry, and clean-exit classification. The macos-launchagent job passed at head 5396e4d: https://github.com/Pika-Yalei/powercontext/actions/runs/33640885614

src/powercontext/server/cli.py
Pika-Yalei: Implemented in 81a3ed9. ServerConfigurationError now limits CLI usage-error translation to env-file loading and settings construction; the context manager leaves that try boundary before yielding to the runner. Runtime OSError therefore remains an operational failure with exit code 1. Regression tests also verify that the temporary environment is restored after the runtime failure.

src/powercontext/service/adapters/launchd.py
Pika-Yalei: Implemented in 81a3ed9. The launchd parser now considers state, last exit code, and terminating signal: a clean exit is inactive, a nonzero exit or signal is failed, a missing job is inactive, and indeterminate output is unknown. A 10-case parameter table covers the launchctl output shapes, and the real macOS retry/exit-classification acceptance test passed at head 5396e4d.

src/powercontext/service/controller.py
Pika-Yalei: Implemented in 81a3ed9. Uninstall now handles stop, disable, remove, and reload as separate stages. On failure it re-inspects the remaining registration and manager facts, attaches ServiceStatus and an exact platform-scoped recovery action, and the CLI renders those facts while exiting nonzero. Regression coverage includes all four failure stages, exact systemctl and launchctl recovery commands, and CLI output.

src/powercontext/service/controller.py
Pika-Yalei: Implemented in 81a3ed9. loaded_registration now reports not_loaded, owned, foreign, or unknown independently from the disk artifact, and install/uninstall fail closed before mutation for foreign or unknown manager objects. systemd validates LoadState, FragmentPath, ExecStart, marker, and metadata; launchd validates path, program, arguments, marker, and metadata. Parameterized corruption cases, manager-query failures, controller no-mutation assertions, legacy ownership, and a real foreign-registration acceptance test cover this boundary.

src/powercontext/cli/system.py
Pika-Yalei: Implemented in 81a3ed9. Doctor first calls registration_status, which inspects only platform support, the artifact, definition, endpoint, and logs without touching the manager or endpoint. It canonicalizes the requested and registered loopback endpoints and calls full status only after they match; remote and mismatched loopback targets skip local service correlation. Tests cover mismatched ports, localhost, the IPv4 127/8 range, IPv6 loopback, default ports, remote targets, invalid schemes, and invalid ports.

src/powercontext/service/launcher.py
Pika-Yalei: Implemented in 81a3ed9 by moving retry accounting into the standard-library-only powercontext_service_bootstrap package, before application imports. RetryStart distinguishes started, exhausted, and state_error. Exhaustion exits cleanly only to terminate the bounded policy; corrupt, unsafe, unreadable or unwritable state and invalid configuration return nonzero and remove the PathState token. The regression matrix covers malformed JSON and schema values, write failure, permissions, symlink state, invalid limits, argparse failure, success cleanup, and dependency-import failure.

src/powercontext/service/adapters/launchd.py
Pika-Yalei: Implemented and verified at head 5396e4d. The native-service workflow runs the same wheel-installed acceptance suite against a real systemd user manager on ubuntu-latest and a real LaunchAgent on macos-latest. Both matching-platform jobs passed: Linux completed install, query, stop, restart, remove, definition matching, and foreign-registration rejection; macOS also covered bounded retry and exit classification. Run: https://github.com/Pika-Yalei/powercontext/actions/runs/33640885614

src/powercontext/service/adapters/launchd.py
Pika-Yalei: Implemented in 81a3ed9. Both adapters recognize only an intact, exactly rendered legacy PowerContext definition as owned; definition_state then classifies its older definition version as stale so install can reconcile it. A modified legacy plist or unit remains invalid. Separate regression tests cover intact and tampered legacy definitions for launchd and systemd, including loaded-manager ownership.

src/powercontext/service/adapters/systemd.py
Pika-Yalei: Confirmed and fixed in 81a3ed9. systemd activating and deactivating now map to unknown rather than inactive. For an owned loaded unit, stop therefore issues systemctl stop instead of skipping it, and a definition reload issues restart. Parameterized transitional-state tests plus explicit stop/restart command tests cover this, and the real Linux systemd user lifecycle job passed at head 5396e4d.

Alanxtl (approved):
lgtm

PR #1272, opened 2026-08-19
test: add LoCoMo-derived e2e acceptance cases

knqiufan:
## Summary

- Refs https://github.com/oceanbase/powercontext/issues/1263
- Add three pinned LoCoMo-derived acceptance cases. Reuse Harbor and `powercontext-e2e acceptance`; do not add a new runner.
  - `locomo-temporal-banker` — category 2, `conv-30:q001`, `D1`
  - `locomo-multihop-football` — category 3, `conv-47:q026`, `D13`
  - `locomo-open-pastries` — category 4, `conv-44:q064`, `D3`
- These are pinned samples, not a LoCoMo score.
- These three cases are positive recall probes only. Negative cases (abstention, conflict, or forbidden context) are out of scope here: the current scorer cannot assert `forbidden_context`. Those samples belong in a follow-up PR once the evaluation contract can express them. Later algorithm changes should then include both positive and negative cases on a named subset.
- The official EN/ZH sample guide is deferred until the aggregated-task path is stable, so this page does not go stale. A follow-up docs PR will use documentation-writer and humanizer, then a human pass.
- The LoCoMo README dead-link cleanup is a separate small PR: #1275.

## Test plan

- [x] `make harness-check` equivalent: ruff, ty, `e2e/bub/tests` (5 passed)
- [x] Fill `dataset.checksum` from CI `task_provenance_matches` / `replay.json` `harbor.task_checksum`
- [x] CI Acceptance (sqlite / oceanbase) passed after checksum backfill

## AI usage statement

Used an AI coding assistant to implement the checked plan and draft this PR. Harbor checksums were taken from CI observations, not invented.

PR #1432, opened 2026-09-02
feat(e2e): support forbidden recall context

russell001209-ai:

Refs #1263.

Follow-up to #1272.

# Rationale for this change

PR #1272 added declarative recall probes but deferred negative benchmark cases because the evaluator could not express context that must not be returned. Without a negative contract, a probe can satisfy its required fragments while still returning stale or otherwise forbidden context.

- Add an optional `forbidden_context` contract to recall probes.
- Match required and forbidden fragments case-insensitively with Unicode normalization.
- Reject blank fragments and contracts where a forbidden fragment is contained in required context.
- Treat any forbidden match as a hard acceptance failure, independent of `probe_coverage`.
- Record the pre-redaction Boolean verdict for replay scoring and discard matched prepared context before persistence.
- Preserve compatibility with existing manifests and older replays that do not contain the new field.
- Document the contract in the Bub README and the English and Chinese RFCs.

The related production substring paths were checked: recall-probe matching is updated here, while `expected_memory` remains a separate Memory-recording contract and is unchanged.

E2E workload authors can now declare context that a recall probe must not return. Existing manifests keep their previous behavior because `forbidden_context` defaults to empty. Existing evaluation reports are unchanged unless a workload declares a forbidden fragment. There are no PowerContext Server, Client, or persisted Memory API changes, and no new model calls or scheduled work.

# How was this change tested?

- `uv run --locked ruff check e2e/bub`
- `uv run --locked ruff format --check e2e/bub`
- `uv run --locked ty check --project e2e/bub --python e2e/bub/.venv --python-version 3.12 e2e/bub/src integrations/bub/src`
- `uv run --project e2e/bub python -m pytest e2e/bub/tests -q` (`23 passed`)
- `uv run --project e2e/bub powercontext-e2e --help`
- `git diff --check`

Regression coverage includes a forbidden match with `probe_coverage: 0`, canonically equivalent Unicode, manifest validation, removal of matched context before replay persistence, and live/offline replay parity for both recorded `true` and `false` verdicts.

# AI usage statement

OpenAI Codex with GPT-5.6-sol was used for implementation assistance and independent adversarial review. The final diff passed the repository checks listed above, and the final independent review reported no remaining findings.

### DIFF
--- a/e2e/bub/src/powercontext_e2e/evaluation.py
+++ b/e2e/bub/src/powercontext_e2e/evaluation.py
@@ -141,16 +141,17 @@ class MemoryEvaluator:
             and _probe_matches_forbidden_context(probe, probe_spec.forbidden_context)
         ]
         forbidden_context_match_ids = set(forbidden_context_matches)
+        positive_probes = [probe_spec for probe_spec in evaluation.probes if probe_spec.expected_context]
         supported_probes = [
             probe_spec
-            for probe_spec in evaluation.probes
+            for probe_spec in positive_probes
             if (probe := probes_by_id.get(probe_spec.id)) is not None
             and probe.prepared_context.status == "ready"
             and bool(probe.prepared_context.content.strip())
             and _contains_fragments(probe.prepared_context.content, probe_spec.expected_context)
             and probe_spec.id not in forbidden_context_match_ids
         ]
-        probe_coverage = len(supported_probes) / len(evaluation.probes)
+        probe_coverage = len(supported_probes) / len(positive_probes) if positive_probes else 1.0
         in_run_contexts = sum(
             record.event == "context"
             and record.status == "ready"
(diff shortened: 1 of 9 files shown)


e2e/bub/src/powercontext_e2e/evaluation.py
```diff
@@ -132,13 +133,22 @@ def evaluate(observation: TaskObservation, *, experiment: str) -> EvaluationRepo
...
+        ]
+        forbidden_context_match_ids = set(forbidden_context_matches)
         supported_probes = [
             probe_spec
             for probe_spec in evaluation.probes
             if (probe := probes_by_id.get(probe_spec.id)) is not None
             and probe.prepared_context.status == "ready"
             and bool(probe.prepared_context.content.strip())
             and _contains_fragments(probe.prepared_context.content, probe_spec.expected_context)
+            and probe_spec.id not in forbidden_context_match_ids
```
PsiACE, line 152: A negative-only probe still fails when the correct result is no context. With empty `expected_context`, non-empty `forbidden_context`, and `status: empty`, the forbidden check passes but `probe_coverage` remains zero. Please treat these probes separately from positive coverage, or add an explicit `require_context` contract.

PsiACE (approved):
lgtm

PR #1438, opened 2026-09-02
feat(api): implement source and artifact base REST APIs

Zxf-xufeng:
﻿
Implements RFC #1437.
﻿
Scope creation, retrieval, and binding remain covered by #1401.
﻿
# Rationale for this change
﻿
RFC #1437 defines a consistent base HTTP API for directly accessing durable Sources and committed Artifacts below an existing Scope.
﻿
This change implements that contract across the OpenAPI specification, server, persistence layer, Python client, and integration clients. It keeps Source creation independent from downstream generation and preserves the existing domain-specific Memory, Candidate review, Skill, and Handoff workflows.
﻿
﻿
- Add nine Scope-qualified HTTP operations for Source and Artifact access.
- Implement Source:
 - Create with a server-generated `source_id`
 - Get by `(scope_id, source_type, source_id)`
 - List or Search a Source type collection through one collection `GET`
- Implement Artifact:
 - Create Revision 1 with a server-generated `artifact_id`
 - Get the current head
 - Get an immutable historical revision
 - List or Search an Artifact family collection through one collection `GET`
 - Replace the current head by creating the next immutable revision
 - Logically delete the current head
- Add pagination and opaque cursors bound to the collection and query parameters.
- Return summary items without full content from collection operations.
- Add `Location`, ETag, `If-Match`, and `If-None-Match` behavior for creation, optimistic concurrency, and conditional reads.
- Keep Source creation independent from Memory, Experience, Skill, Handoff, and other generation commands.
- Enforce Artifact Family capabilities so generic writes cannot bypass Candidate review or other protected workflows.
- Reuse the existing Source journal, Artifact revision/head, and lineage tables rather than creating a second persistence model.
- Add the persistence fields required by the RFC:
 - `pc_sources.created_at`
 - `pc_artifacts.created_at`
 - `pc_artifact_heads.deleted_at`
- Update `openapi/powercontext.yaml` and regenerate the checked-in Python models, operations, and schema.
- Update the Python client and the DSH, OpenCode, and Pi integration clients.
- Add contract, generation, persistence, client, integration, and end-to-end coverage.
﻿
﻿
Yes. This is an additive API change.
﻿
The new endpoints expose:
﻿
- Source Create, Get, List, and Search.
- Artifact Create, Get, Get Revision, List, Search, Replace, and Delete.
- Server-generated Source and Artifact identifiers.
- Immutable Artifact revisions and optimistic concurrency through ETags.
- Logical Artifact deletion while retaining historical revisions.
﻿
List and Search share the same collection endpoints: an omitted or blank `query` performs List, while a non-empty `query` performs Search.
﻿
Existing domain-specific APIs and workflows remain available and are not replaced by these base APIs. No existing public endpoint is removed or changed.
﻿
# How was this change tested?
﻿
- `make check`
- `make contract-test`
- `make api-generate-check`
- `make js-api-generate-check`
- `make test` (`1047 passed, 12 skipped`)
- DSH plugin tests, build, and generation checks
- OpenCode plugin tests and type checking
- Pi plugin tests and type checking
- Strict Zensical documentation build
- `git diff --check`
- GitHub Actions across Python 3.11, 3.12, 3.13, and 3.14
- SQLite and OceanBase acceptance suites
﻿
# AI usage statement
﻿
OpenAI Codex was used to update the cross-layer contract and generated clients, add regression coverage, review implementation risks, and draft the PR description.


src/powercontext/builtin/persistence/records.py
```diff
@@ -0,0 +1,574 @@
...
+                if is_generation_eligible(stored.value):
+                    sources.append(source_ref)
+            draft = self._validated_draft(
+                family,
+                write.content,
+                sources=tuple(sources),
+                artifacts=current.lineage.artifacts,
+            )
+            try:
+                revised = await self._artifacts.revise(connection, scope_id, current, draft)
```
frf12: Each Revision created by this management API should reference a server-owned `lineage_only` Source that records the direct write. The Revision 1 Source is filtered above, but Replace does not create a replacement Source, so Revision 2 can return `sources: []`. Please create a Source bound to the next Revision and commit it in the same transaction as `revise()`.

src/powercontext/builtin/persistence/records.py
```diff
@@ -0,0 +1,574 @@
...
+                    family=cast(Any, family),
+                    artifact_id=artifact_id,
+                ),
+            ),
+        )
+        try:
+            async with self._database.transaction() as connection:
+                stored = await self._sources.add(connection, scope_id, source)
+                draft = draft.model_copy(update={"sources": (stored.ref,)})
+                artifact = await self._artifacts.create(
```
frf12: This path validates the Family content model and writes directly through `ArtifactRepository`, but it does not perform the complete Family write behavior. It skips Memory entry versions and projections, Handoff citation validation and fixed identity, Experience Search updates, and Skill package validation and Search updates. Please dispatch Create and Replace through Family-specific write handlers while retaining the direct management API behavior.

src/powercontext/builtin/persistence/records.py
```diff
@@ -0,0 +1,574 @@
...
+        source_type: str,
+        content: JsonValue,
+        /,
+    ) -> SourceRecord:
+        self._require_content_source(source_type)
+        source = ContentSource(
+            name=self._id_factory("source"),
+            materialization=SourceMaterialization.CAPTURED,
+            content=_canonical_source_text(content),
+            wire_content=None if isinstance(content, str) else _JSON_VALUE.validate_python(content, strict=True),
```
frf12: `wire_content=None` represents both a plain string Source and JSON `null`. As a result, `content: null` is stored and returned as the JSON string `"null"`. Please preserve the distinction between an absent `wire_content` value and an explicit JSON `null`.

frf12 (commented):
Please see the inline comments.

src/powercontext/builtin/persistence/records.py
```diff
@@ -0,0 +1,574 @@
...
+                    ARTIFACT_HEADS_TABLE.c.artifact_id > after,
+                )
+                .order_by(ARTIFACT_HEADS_TABLE.c.artifact_id)
+                .limit(limit + 1)
+            )
+            rows = (await connection.execute(statement)).all()
+            selected_rows = rows[:limit]
+            items = []
+            for row in selected_rows:
+                artifact = await self._artifacts.get(
```
frf12: This loop performs a separate Artifact and Source/Artifact reference read for every item after loading the heads. A 100-item page therefore produces many database queries, while RFC #1437 requires the response data to be read in batches. Please batch-load the selected Revisions and their `sources` and `artifacts`.

src/powercontext/builtin/persistence/records.py
```diff
@@ -0,0 +1,574 @@
...
+        if isinstance(cursor_ttl_seconds, bool) or cursor_ttl_seconds < 1:
+            raise ValueError("cursor_ttl_seconds must be a positive integer")  # noqa: TRY003
+        if cursor_secret is not None and not cursor_secret:
+            raise ValueError("cursor_secret must not be empty")  # noqa: TRY003
+        self._database = database
+        self._sources = sources
+        self._artifacts = artifacts
+        self._clock = _utc_now if clock is None else clock
+        self._id_factory = _resource_id if id_factory is None else id_factory
+        self._cursor_secret = secrets.token_bytes(32) if cursor_secret is None else cursor_secret
```
frf12, line 110: This default creates a new `cursor_secret` whenever the Runtime starts. A cursor issued before a server restart therefore becomes invalid before its recorded expiration time. Please provide a stable configured secret, or define server restart as part of the cursor validity contract.

src/powercontext/builtin/persistence/records.py
Zxf-xufeng: Agreed. The current implementation is an N+1 path: after loading the heads, each item performs one Revision query and two lineage queries.  
I’ll add a repository-level batch loader keyed by the complete Revision identity, batch-load Source and Artifact lineage, and assemble results in page and ordinal order. The list path will then use a bounded constant number of queries. I’ll add ordering, lineage, pagination, and query-budget regression tests.

src/powercontext/builtin/persistence/records.py
Zxf-xufeng: Agreed. Generating the signing secret in each Runtime instance makes cursor lifetime accidentally process-bound.  
I’ll inject a stable cursor-signing secret through the Server and Runtime configuration. Local installations will reuse a protected key under POWERCONTEXT_HOME, while replicated deployments can provide the same secret through their secret manager. Tests will verify cursor use across Runtime reconstruction, invalid signatures after key rotation, and the existing 410 cursor_expired behavior.

src/powercontext/builtin/persistence/records.py
Zxf-xufeng: I’ll add an explicit persisted presence discriminator for wire_content, keep missing values backward-compatible with existing text Sources, and update response/digest mapping to use it. Regression tests will cover JSON null, the string "null", persistence round-trips, and HTTP responses.

src/powercontext/builtin/persistence/records.py
Zxf-xufeng: Agreed. We will update Replace so that every Revision created through this management API has its own server-owned lineage_only Source recording the direct write.
For the next Revision N, Replace will:
- create a new Source with role=lineage_only, operation=artifact_replace, and a target bound to (scope_id, family, artifact_id, revision=N);
- use that new Source as the Source lineage of Revision N;
- preserve the previous Revision and its Source unchanged;
- inherit the previous Artifact lineage for now;
- commit the Source, new Artifact Revision, head update, and lineage rows in the same transaction.

src/powercontext/builtin/persistence/records.py
Zxf-xufeng: Thanks — fixed. The Base Artifact API now dispatches Create/Replace through a FamilyManagementWriter registry:
- Memory reuses MemoryService, maintaining Entry Versions/heads, changes and search projections.
- Experience and Skill maintain their existing search/package projections.
- Handoff reuses HandoffService, including singleton enforcement and citation lineage.
The lineage-only Source, Artifact Revision and Family-derived state are committed in one transaction and roll back together on failure. No LLM generation or Candidate workflow is triggered. Regression tests and all PR checks pass.

Teingi (approved):
LGTM

PR #1441, opened 2026-09-03
feat: add windows launch service

### LINKED ISSUE #1298: Tracking Issue: local Server availability and service installation

Alanxtl:

Closes #1298

Implements [RFC 1299](https://github.com/oceanbase/powercontext/blob/master/docs/zh/rfcs/1299_local_server_availability_and_service_installation.md).

# Rationale for this change

Complete the Windows implementation of the personal Server service lifecycle defined by RFC 1299. Windows users can run PowerContext as a persistent current-user service without administrator privileges or a visible console window.

- Add the Windows Task Scheduler adapter for the current user.
- Add ownership validation, lifecycle operations, restart policy, and foreign-task protection.
- Use `pythonw.exe` to avoid opening a visible CMD window.
- Redirect Server stdout and stderr to PowerContext-owned log files.
- Support enabling or disabling automatic startup when the current user logs in.
- Add Windows-native lifecycle tests and Windows CI coverage.
- Update README and deployment documentation for Windows.

Windows is now supported by `powercontext service install`, `status`, and `uninstall`.

The service runs as the current user through Task Scheduler and does not install a Windows Service or require administrator privileges. Existing service definitions remain compatible.

# How was this change tested?

- `make check`
- `make docs-test`
- `uv run pytest -q tests/test_service.py tests/test_service_environment.py tests/test_system_cli.py`
- Windows Task Scheduler native lifecycle tests on an isolated Windows environment:
  `4 passed, 1 skipped`
- Verified `powercontext service install --help`
- tested on my windows

# AI usage statement

OpenAI Codex

### DIFF
--- a/src/powercontext/service/adapters/windows.py
+++ b/src/powercontext/service/adapters/windows.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import csv
 import io
+import json
 import os
 import shutil
 import subprocess
@@ -95,6 +96,8 @@ class WindowsTaskSchedulerAdapter:
             return SupportState.UNSUPPORTED, "Task Scheduler personal services are available only on Windows"
         if shutil.which("schtasks.exe") is None:
             return SupportState.UNSUPPORTED, "schtasks.exe is not installed or is not on PATH"
+        if shutil.which("powershell.exe") is None:
+            return SupportState.UNSUPPORTED, "powershell.exe is not installed or is not on PATH"
         try:
             account, sid = self._user_identity()
         except ServiceError as error:
@@ -293,20 +296,26 @@ class WindowsTaskSchedulerAdapter:
         self.artifact_path.unlink(missing_ok=True)
 
     def manager_state(self) -> ManagerState:
-        result = self._run("/Query", "/TN", self.identifier, "/FO", "LIST", "/V", "/HRESULT", check=False)
-        if _is_task_not_found(result):
-            return ManagerState.INACTIVE
+        result = self._run_task_info(check=False)
         if result.returncode != 0:
             return ManagerState.UNKNOWN
-        values = _list_output(result.stdout)
-        status = values.get("status", "").casefold()
-        if status in {"running", "正在运行"}:
+        try:
+            values = json.loads(result.stdout)
+        except json.JSONDecodeError:
+            return ManagerState.UNKNOWN
+        if not isinstance(values, dict):
+            return ManagerState.UNKNOWN
+        status = values.get("State")
+        if not isinstance(status, str):
+            return ManagerState.UNKNOWN
+        status = status.casefold()
+        if status == "notfound":
+            return ManagerState.INACTIVE
+        if status == "running":
             return ManagerState.ACTIVE
-        if status in {"ready", "disabled", "queued", "not running", "就绪", "已禁用"}:
-            last_result = _last_result(values.get("last result", ""))
+        if status in {"ready", "disabled", "queued"}:
+            last_result = _last_result(values.get("LastTaskResult"))
             return ManagerState.FAILED if last_result not in {None, 0} else ManagerState.INACTIVE
-        if "could not start" in status or "failed" in status:
-            return ManagerState.FAILED
         return ManagerState.UNKNOWN
 
     def log_location(self, definition: ServiceDefinition | None) -> str | None:
@@ -338,6 +347,9 @@ class WindowsTaskSchedulerAdapter:
         settings = _child(root, "Settings")
         actions = _child(root, "Actions")
         execute = _child(actions, "Exec") if actions is not None else None
+        structure_mismatch = _task_structure_mismatch(root, definition)
+        if structure_mismatch is not None:
+            return structure_mismatch
         if definition.start_on_login:
             logon_matches = logon is not None and (
                 _text(logon, "Enabled").casefold() in {"", "true"}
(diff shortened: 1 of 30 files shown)


src/powercontext/service/adapters/windows.py
```diff
@@ -0,0 +1,536 @@
...
+        self.artifact_path.unlink(missing_ok=True)
+
+    def manager_state(self) -> ManagerState:
+        result = self._run("/Query", "/TN", self.identifier, "/FO", "LIST", "/V", "/HRESULT", check=False)
+        if _is_task_not_found(result):
+            return ManagerState.INACTIVE
+        if result.returncode != 0:
+            return ManagerState.UNKNOWN
+        values = _list_output(result.stdout)
+        status = values.get("status", "").casefold()
```
Teingi: [P1] On localized Windows installations, `/FO LIST` localizes the field names as well as the values, but these lookups only accept the English keys `status` and `last result`. For example, `状态: 正在运行` is stored under `状态`, so `manager_state()` returns `UNKNOWN`; the first install then aborts in `start(reload_definition=True)` before `/Run`. Please use a locale-independent Task Scheduler state API and add a non-English regression test.

src/powercontext/service/adapters/windows.py
```diff
@@ -0,0 +1,536 @@
...
+        account, sid = self._user_identity()
+        expected_arguments = _launcher_arguments(definition)
+        registration = _child(root, "RegistrationInfo")
+        triggers = _child(root, "Triggers")
+        logon = _child(triggers, "LogonTrigger") if triggers is not None else None
+        principals = _child(root, "Principals")
+        principal = _child(principals, "Principal") if principals is not None else None
+        settings = _child(root, "Settings")
+        actions = _child(root, "Actions")
+        execute = _child(actions, "Exec") if actions is not None else None
```
Teingi, line 349: [P1] This validates only the first `Exec` action. Appending a second `Exec` (for example, `cmd.exe`) still makes `loaded_registration()` return `OWNED`, so a modified task can execute extra commands while passing the status/install/uninstall ownership checks. Please require exactly the expected action, trigger, and principal structure, reject any extras as `FOREIGN`, and add regression tests for additional elements.

src/powercontext/service/environment.py
```diff
@@ -104,6 +109,91 @@ def _validate_protection(path: Path, status: os.stat_result) -> None:
         )
 
 
+def _validate_windows_protection(path: Path) -> None:
+    """Require a Windows ACL limited to the interactive user and trusted OS admins."""
+
+    account, sid = _windows_user_identity()
```
Teingi, line 118: [P1] This ACL check never verifies the file owner SID. On Windows `st_uid` is recorded as `0`, so a file owned by another account can pass as long as its current DACL contains the allowed principals; that owner can later change the DACL and regain access to the credential file. Please require the current user's owner SID, persist it in the installed identity, and revalidate it on launch.

src/powercontext/service/cli.py
```diff
@@ -49,17 +50,38 @@ def install(
...
+        ),
+    ] = None,
 ) -> None:
-    """Install, enable, and start the personal Server service."""
-
+    """Install the personal Server service and optionally start it at user login."""
+
+    if start_on_login is None:
+        start_on_login = (
+            typer.confirm("Enable automatic Server startup when you log in?", default=False)
```
Teingi, line 65: [P2] Pressing Enter here installs without login auto-start, while the README and English RFC still document bare `powercontext service install` as enabling startup for future logins; only the Chinese RFC describes this prompt and opt-out. Please choose one default and keep the implementation and both locales in sync.

src/powercontext/service/adapters/windows.py
Alanxtl: fixed, now use PowerShell Task Scheduler API

src/powercontext/service/adapters/windows.py
Alanxtl: fixed, now check action、trigger、principal only

src/powercontext/service/environment.py
Alanxtl: fixed, now Windows recheck owner SID

src/powercontext/service/cli.py
Alanxtl: fixed

Teingi (approved):
LGTM

PR #1444, opened 2026-09-03
fix: env file parser

### LINKED ISSUE #1443: bug: Environment file parser does not support multiline quoted values
### Describe the bug

PowerContext environment files do not accept a multiline quoted JSON value.

### Steps to reproduce

cat server.env
POWERCONTEXT_SERVER_HTTP_HOST=127.0.0.1
POWERCONTEXT_SERVER_HTTP_PORT=8888

POWERCONTEXT_SERVER_DASHBOARD_SCOPES='[
  {
    "scope_id": "git:github.com/oceanbase/obproxy",
    "display_name": "obproxy"
  },
  {
    "scope_id": "git:github.com/oceanbase/oceanbase",
    "display_name": "oceanbase"
  },
  {
    "scope_id": "git:github.com/oceanbase/powercontext",
    "display_name": "powercontext"
  }
]'

 powercontext config validate --env-file "$HOME/.config/powercontext/server.env"

### Expected behavior

PowerContext should support multiline quoted JSON values in env files.
If multiline values are intentionally unsupported, the documentation and error message should explicitly state that every assignment must be written on a single line and suggest converting the JSON value to single-line JSON.

### Actual behavior

A multiline POWERCONTEXT_SERVER_DASHBOARD_SCOPES value fails with:

Error: invalid assignment at /Users/xxx/.config/powercontext/server.env:5: No closing quotation

### Environment

 powercontext --version
0.0.3.dev66+g4213d32d6

thunguo:

Closes #1443.

# Rationale for this change

PowerContext parsed environment files one physical line at a time. As a result, a quoted value that continued onto subsequent lines was rejected at its opening line with `No closing quotation`, even when the quote was closed later in the file. This prevented readable multiline JSON values such as `POWERCONTEXT_SERVER_DASHBOARD_SCOPES` from being used with `--env-file`.

- Continue consuming physical lines when a quoted environment value has not closed yet.
- Preserve newlines and indentation inside multiline single-quoted and double-quoted values.
- Keep the parser shell-free and retain its existing validation for shell expansion, invalid assignments, duplicate names, and NUL characters.
- Add parser regression coverage for multiline JSON, double-quoted values, and assignments following a multiline value.
- Add CLI regression coverage proving that `powercontext config validate --env-file` accepts multiline dashboard scope JSON.

Yes. Commands that accept `--env-file` now support multiline quoted values. Existing single-line environment files continue to behave as before. There are no public API changes, persisted-format changes, breaking changes, or migration requirements.

# How was this change tested?

- Reproduced the original `No closing quotation` failure before applying the fix.
- `make check` — passed, including lock-file validation, pre-commit hooks, Ruff, and type checking.
- `uv run pytest tests/test_env_file.py tests/test_config_cli.py -q` — 49 passed.
- `uv run pytest tests/test_server.py tests/test_service_environment.py tests/test_service.py -q` — 148 passed, 2 skipped.
- `make test` — 1,288 passed and 19 skipped. Three `tests/client/test_receiver_service.py` tests failed locally because they exercise Linux-only systemd service installation while the suite was run on macOS; these failures are unrelated to the environment-file parser change.

# AI usage statement

### DIFF
--- a/src/powercontext/cli/env_file.py
+++ b/src/powercontext/cli/env_file.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 
 import os
 import re
-from collections.abc import Collection, Generator, Mapping, MutableMapping
+from collections.abc import Collection, Generator, Iterator, Mapping, MutableMapping
 from contextlib import contextmanager
 from pathlib import Path
 
@@ -44,7 +44,8 @@ def parse_environment(content: str, *, source: str = "environment") -> dict[str,
     """
 
     environment: dict[str, str] = {}
-    for line_number, line in enumerate(content.splitlines(), start=1):
+    lines = iter(enumerate(content.splitlines(), start=1))
+    for line_number, line in lines:
         stripped = line.lstrip(" \t")
         if not stripped or stripped.startswith("#"):
             continue
@@ -53,12 +54,7 @@ def parse_environment(content: str, *, source: str = "environment") -> dict[str,
             stripped = stripped[export.end() :]
         if "\x00" in stripped:
             raise EnvironmentFileError(f"invalid NUL character at {source}:{line_number}")  # noqa: TRY003
-        try:
-            tokens = _split_shell_words(stripped)
-        except ValueError as error:
-            raise EnvironmentFileError(  # noqa: TRY003
-                f"invalid assignment at {source}:{line_number}: {error}"
-            ) from error
+        tokens = _split_assignment(stripped, lines, source=source, line_number=line_number)
         if not tokens:
             continue
         if len(tokens) != 1 or "=" not in tokens[0]:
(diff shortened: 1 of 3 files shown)


src/powercontext/cli/env_file.py
```diff
@@ -76,6 +72,37 @@ def parse_environment(content: str, *, source: str = "environment") -> dict[str,
...
+    *,
+    source: str,
+    line_number: int,
+) -> list[str]:
+    """Split one assignment, consuming physical lines until its quotes close."""
+
+    assignment = first_line
+    while True:
+        try:
+            return _split_shell_words(assignment)
```
Teingi: Re-running `_split_shell_words()` on the entire accumulated assignment for every continuation makes valid multiline values quadratic. On this head, a pretty-printed 400-scope JSON value (29 KB, 1,602 lines) takes about 3.5s to parse versus about 0.004s when compact; a 500-to-1,000-line synthetic input grows from 2.8s to 11.5s. Because every `--env-file` command hits this path, larger readable configs can significantly delay validation and server startup. Please carry quote/escape state forward and scan each character once, with a large multiline regression or benchmark.

thunguo: got it,I'll take care of that later.

thunguo: Fixed in 72547b9. `_split_assignment()` now retains lexer state across physical lines and `_scan_shell_words()` processes each character once, eliminating repeated reparsing of the accumulated value. On the same 400-scope, 1,602-line (~29 KB) input, local parse time dropped from 1.116s to 0.0014s. I also added `test_large_multiline_value_parses_within_linear_time_budget`, covering an 800-scope (~3,202-line) value with a 1-second budget. `make check` passes; focused parser/CLI tests report 50 passed, and Server/service environment tests report 148 passed, 2 skipped.

Teingi (approved):
LGTM

PR #1445, opened 2026-09-03
fix: update validation to accept minimal server environment without inference models

### LINKED ISSUE #1442: bug: powercontext config validate rejects a server-runnable env file when inference models are not configured
### Describe the bug

powercontext config validate --env-file requires a complete inference configuration even when the env file is only intended to configure a minimal PowerContext server.
In this case, the env file configures seekdb, the HTTP endpoint, and the Dashboard scopes. The server can use these settings, but config validate rejects the file because POWERCONTEXT_SERVER_INFERENCE_GENERATION_MODEL is missing.
This behavior is confusing because users may reasonably expect config validate to validate the same minimal configuration accepted by powercontext server run. If a complete generated configuration is intentionally required, the command help, documentation, and error message should clearly explain that requirement.

### Steps to reproduce

cat server.env
POWERCONTEXT_SERVER_HTTP_HOST=127.0.0.1
POWERCONTEXT_SERVER_HTTP_PORT=8888
POWERCONTEXT_SERVER_DASHBOARD_SCOPES='[{"scope_id":"git:github.com/oceanbase/obproxy","display_name":"obproxy"},{"scope_id":"git:github.com/oceanbase/oceanbase","display_name":"oceanbase"},{"scope_id":"git:github.com/oceanbase/powercontext","display_name":"powercontext"}]'

powercontext config validate --env-file "$HOME/.config/powercontext/server.env"

### Expected behavior

powercontext config validate --env-file <file> should validate the configuration that is required for the selected server features. A minimal configuration that can be successfully consumed by powercontext server run, such as a server using seekdb with the Dashboard enabled, should not fail only because optional inference features are not configured.
Alternatively, the command and documentation should clearly state that config validate only accepts a complete generated configuration and cannot be used to validate a minimal server env file.

### Actual behavior

The server env file contains valid database, HTTP, and Dashboard scope settings. However, powercontext config validate fails immediately with:
Error: missing required environment value: POWERCONTEXT_SERVER_INFERENCE_GENERATION_MODEL

### Environment

 powercontext --version
0.0.3.dev66+g4213d32d6

jiannnnyyyyy:

Closes #1442.

# Rationale for this change

`powercontext config validate --env-file` incorrectly required a complete inference configuration, even when the environment file was only intended to configure a minimal Server.

This made valid configurations accepted by `powercontext server run --env-file` fail validation because no generation model was configured.

- Allow `config validate` to validate minimal Server-only environment files without inference models.
- Keep strict validation for complete configurations generated by `config init`.
- Validate provider models only when inference models are configured.
- Ensure env-file validation ignores stale `POWERCONTEXT_SERVER_*` values from the shell.
- Add a regression test covering the seekDB, HTTP, and Dashboard configuration from issue #1442.
- Update English and Chinese configuration documentation.

Yes. `powercontext config validate --env-file` now accepts valid minimal Server configurations that do not configure inference models.

There are no breaking API changes or persistence format changes.

# How was this change tested?

- `pytest -q tests/test_config_cli.py -k 'not init_validate_and_show_round_trip_managed_environment'`
  - 22 passed
- Manual validation using the exact issue scenario:
  - seekDB
  - `127.0.0.1`
  - port `8888`
  - multiple Dashboard scopes
  - no inference models configured
- Ruff linting and formatting checks
- Ty type checking
- `zensical build --strict`

The skipped existing test checks Unix file permissions and is not portable on Windows.

# AI usage statement

This PR was developed with OpenAI Codex using the GPT-5 model. The AI assisted with repository inspection, implementation, test updates, documentation updates, and validation.

### DIFF
--- a/src/powercontext/cli/config.py
+++ b/src/powercontext/cli/config.py
@@ -28,13 +28,16 @@ from contextlib import contextmanager
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from pathlib import Path
-from typing import Annotated, Never
+from typing import TYPE_CHECKING, Annotated, Never
 from urllib.parse import urlsplit
 
 import typer
 from pydantic import ValidationError
 
 from powercontext.cli.env_file import EnvironmentFileError, parse_environment
+
+if TYPE_CHECKING:
+    from powercontext.server.settings import ServerSettings
 
 HELP_OPTION_NAMES = ("-h", "--help")
 MANAGED_BEGIN = "# >>> powercontext managed configuration >>>"
@@ -287,13 +290,16 @@ def show_command(
 def validate_command(
     env_file: Annotated[Path, typer.Option(help="Environment file to validate.")] = Path(".env"),
 ) -> None:
-    """Validate syntax, model configuration, and Server settings."""
+    """Validate syntax, configured model adapters, and Server settings."""
 
     try:
         content = env_file.read_text(encoding="utf-8")
         values = parse_environment(content, source=str(env_file))
-        configuration = configuration_from_document(content)
-        _validate_operational_configuration(configuration, values=values)
+        if _has_complete_generated_configuration(values) or _managed_metadata(content):
+            configuration = configuration_from_document(content)
+            validate_configuration(configuration)
+        _validate_server_settings(values)
+        _validate_builtin_runtime(values)
     except (ConfigError, EnvironmentFileError, OSError, UnicodeError, ValidationError) as error:
         _fail(str(error))
     typer.echo(f"Configuration is valid: {env_file.resolve()}")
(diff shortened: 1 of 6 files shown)


src/powercontext/cli/config.py
```diff
@@ -291,13 +295,16 @@ def show_command(
...
     try:
         content = env_file.read_text(encoding="utf-8")
         values = parse_environment(content, source=str(env_file))
-        configuration = configuration_from_document(content)
-        _validate_operational_configuration(configuration, values=values)
+        if _has_complete_generated_configuration(values) or _managed_metadata(content):
+            configuration = configuration_from_document(content)
+            validate_configuration(configuration)
+        _validate_server_settings(values)
+        _validate_provider_models(values)
```
Teingi: Could we also validate the inference-dependent runtime features here? With a minimal env that only sets `POWERCONTEXT_SERVER_RUNTIME_SCHEDULE_SECONDS=60`, this command exits 0, but `open_builtin_runtime()` gets no candidate pipeline and raises `scheduled-pipeline` during server startup. The same gap exists for the experience scheduler and memory reranking. Please share a pure runtime preflight, or add equivalent cross-field checks, so successful validation cannot still fail during runtime assembly.

src/powercontext/cli/config.py
```diff
@@ -522,23 +529,51 @@ def _validate_operational_configuration(
...
+async def _construct_provider_models(inference: InferenceConfig) -> None:
+    if inference.generation_model is None and inference.embedding_model is None:
+        return
     from pydantic_ai.embeddings import infer_embedding_model
     from pydantic_ai.models import infer_model
 
-    async with infer_model(configuration.generation.model):
-        infer_embedding_model(configuration.embedding.model)
+    if inference.generation_model is not None:
+        async with infer_model(inference.generation_model):
```
Teingi: This preflight does not use the provider construction semantics used by the Runtime: it drops workload `base_url` and `headers`, and it never validates `rerank_model`. I reproduced both directions: a supported OpenAI-compatible endpoint using static `Authorization` headers is rejected for missing `OPENAI_API_KEY`, while `deepseek:*` plus a custom base URL passes here and then fails with `inference-endpoint-provider` at startup. Please extract or reuse the Runtime provider factory for generation, embedding, and active reranking.

Teingi (commented):
Thanks for addressing #1442. I reproduced the minimal success case, but the current head still has runtime-parity gaps and conflicts with the latest master (`259bee3`).

tests/test_config_cli.py
```diff
@@ -120,6 +120,25 @@ def test_init_validate_and_show_round_trip_managed_environment(
...
 
 
+def test_validate_accepts_minimal_server_environment_without_inference_models(tmp_path: Path) -> None:
+    environment = tmp_path / "server.env"
+    environment.write_text(
+        "\n".join((
+            "POWERCONTEXT_SERVER_DATABASE_KIND=seekdb",
+            "POWERCONTEXT_SERVER_HTTP_HOST=127.0.0.1",
+            "POWERCONTEXT_SERVER_HTTP_PORT=8888",
+            'POWERCONTEXT_SERVER_DASHBOARD_SCOPES=\'[{"scope_id":"git:github.com/oceanbase/powercontext","display_name":"powercontext"}]\'',
```
Teingi: `POWERCONTEXT_SERVER_DASHBOARD_SCOPES` was removed by current `master` in #1401, and extra settings are ignored. After resolving the merge conflict, this test can still pass without exercising any Dashboard scope configuration. Please update the regression to the current organization/scope contract, or drop the Dashboard claim, and assert the effective settings rather than only the exit code.

src/powercontext/cli/config.py
jiannnnyyyyy: Addressed.
I added a pure preflight_builtin_runtime() path and wired config validate to use it. The preflight reuses the Runtime composition logic without opening persistence or making inference requests.
It now validates:
- scheduled source processing
- scheduled Experience incubation
- Memory reranking prerequisites
- whether the required generation, embedding, and reranking pipelines can be assembled
I also added regression coverage for these Runtime features when no required inference model is configured.

src/powercontext/cli/config.py
jiannnnyyyyy: Addressed.
I removed the standalone CLI provider validation and now build ServerSettings from the environment file before running the Runtime preflight. The preflight reuses the same Runtime provider construction path for generation, embedding, and active reranking.

tests/test_config_cli.py
jiannnnyyyyy: Addressed.
I resolved this against the current master contract from #1401. POWERCONTEXT_SERVER_DASHBOARD_SCOPES was removed from the regression environment because current master ignores that setting.

Teingi (approved):
LGTM

Issue #1450, opened 2026-09-04
Tracking Issue: DSH reliability and cross-Agent PowerContext guidance


knqiufan:
## Feature description

Make the DSH installation-to-recall journey diagnosable and make PowerContext tool/Skill routing consistent across maintained Agent integrations.

This is a focused follow-up tracker for a reported user journey: install the plugin, connect to a compatible Server, observe automatic recall/capture, and use natural-language Memory or Handoff requests correctly. It records independently deliverable work and acceptance criteria. Child Issues should own implementation and closure.

### Reported user experience

A PowerContext `0.1.0` user reported that:

1. The DSH plugin loaded and exposed `pc_*` tools, but calls returned `Error: PowerContext Server returned HTTP 404`.
2. Automatic recall, capture, and injection were not apparent in the conversation, leading the user to suspect that the hook was not registered.
3. Tool descriptions and Skill discovery gave weak guidance. In the reported interaction, the Agent started with `pc_memory_list` as if listing stored entries were the normal way to restore context.

The user suggested a `using-powercontext` routing Skill with more focused domain Skills. This is useful as an additional guidance layer, while per-tool descriptions remain necessary because an Agent can call an exposed tool without loading a Skill.

**Evidence boundary:** these are reported symptoms, supplemented by source inspection. The original failing endpoint, response body, installed plugin revision, and exact DSH version have not been captured here. A wrong endpoint, proxy/base-path configuration, or mismatched Server/plugin revision are possibilities; the status code alone does not establish which occurred. An empty Memory store is also different from a missing HTTP route.

### Verified current gaps

Source baseline: [PowerContext master `f0f288a`](https://github.com/oceanbase/powercontext/commit/f0f288abecaccb97e1fe97d991b87b808bbebfbd), inspected on 2026-09-04. Released `0.1.0` and current master must be qualified separately.

| Area | Current behavior and consequence |
| --- | --- |
[...]

Issue #1449, opened 2026-09-04
feat: Automatically use the default Scope when no Scope is configured on first startup


jiannnnyyyyy:
### Feature description

When PowerContext is started for the first time without any Scope-related environment variables, automatically create and use the default Scope, and display it in the Dashboard.

### Problem and proposed solution

After installing and starting PowerContext:

```bash
uv tool install "powercontext[cli,server]"
powercontext server run
```

`powercontext doctor` reports that the package, server, runtime, and database are all healthy. However, opening `http://127.0.0.1:8000` shows:

> No dashboard scope configured.

The Dashboard remains empty, making it unclear whether the server started correctly or what additional configuration is required.

We suggest that PowerContext should:

- Automatically use the server’s default Scope when no Scope is explicitly configured.
- Display the default Scope in the Dashboard.
- Provide an onboarding flow if Scope configuration is required.
- Include actionable Scope configuration guidance in `powercontext doctor`.

### Alternatives considered

- Require users to manually set `POWERCONTEXT_CODEX_SCOPE_ID` or another Scope-related environment variable. This is explicit but creates unnecessary friction for first-time users.
- Add more documentation explaining the required configuration. This may help, but does not address the empty Dashboard or lack of in-product guidance.
- Keep the current behavior and only improve the Dashboard error message. This is easy to implement, but users would still need to configure the Scope manually.

### Additional context

Environment:

```text
powercontext 0.1.0 (PyPI)
macOS
http://127.0.0.1:8000
```

`powercontext doctor` output:

```text
package: ok - powercontext 0.1.0
server liveness: ok
server readiness: ok
runtime: ready
database: ready
```

Related documentation:

<https://github.com/oceanbase/powercontext/blob/master/docs/zh/docs/how-to/full-capability-runtime.md>

### Are you willing to contribute to this feature?

- [x] Yes, I am willing to contribute code, docs, or design feedback.

PR #1452, opened 2026-09-04
fix(dsh): handle Scope failures at tool and command boundaries

### LINKED ISSUE #1451: bug(dsh): handle Scope failures at tool and command boundaries
## Describe the bug

Parent tracking issue: #1450, work package A.

The DSH plugin resolves the Server-owned Scope before entering its direct-operation error boundary. A failure from `POST /v1/scope-bindings/resolve` therefore rejects a registered `pc_*` tool or `/pc` command instead of returning a controlled failure.

The same unconditional prerequisite blocks `/pc doctor`, `/pc capabilities`, bare `/pc`, and even local usage errors. In particular, Doctor cannot reach its health checks when Scope resolution is the failing layer.

After Scope resolution succeeds, direct HTTP 404 responses are normalized to `code: not_found`, but their specific public error code, such as `memory_not_found`, is discarded.

## Steps to reproduce

Use the current DSH plugin's registered tool/command handlers with its real `PowerContextClient` and a controlled fetch fixture:

1. Return HTTP 404 with `{"detail":"Not Found"}` for `/v1/scope-bindings/resolve`; return valid successful JSON for health and capabilities.
2. Execute registered `pc_search`, `/pc doctor`, `/pc capabilities`, bare `/pc`, or an unknown `/pc` subcommand.
3. Observe that each rejects with `PowerContext Server returned HTTP 404`, and that the health/capability routes are never reached.
4. Repeat with a valid Scope-domain response: HTTP 404 and `{"error":{"code":"scope_not_found","message":"The requested Scope was not found."}}`. It still escapes the registered entrypoint.
5. Return a valid Scope from the resolver, then return a Memory-domain 404 from the requested operation. The result is controlled, but contains only `code: not_found`, without the specific `memory_not_found` code.

A real-Server acceptance case should also use a nonexistent explicit Scope: scoped operations must fail without falling back to the Server default, while health and capability inspection remain usable.

## Expected behavior

- Registered tools and commands contain failures across Scope resolution and operation execution; no uncaught host exception or false success.
- Preserve existing error classifications and add optional `error_code` for recognized public 404 domain codes. Keep existing specific 409/422 codes compatible.
- A code-less 404 from an explicitly required endpoint is distinguishable from a valid resource/domain 404. A 404 alone must not be presented as proof of a wrong process or deployment version.
- `/pc doctor`, `/pc capabilities`, unknown commands, and usage validation do not require Scope resolution.
- Bare `/pc` remains inspectable when Scope resolution fails, reporting unresolved Scope and controlled recovery information.
- Other scoped operations retain Server-owned explicit/binding/default precedence. Scope failure or cancellation must not trigger a later write, new binding, or fallback to another Scope.
- Propagate the caller's cancellation signal through Scope resolution without changing timeout budgets or introducing retries.
- Reuse existing content-free native diagnostics and deduplication. Diagnostic delivery failures must not escape the operation boundary. Do not echo Server messages, URLs, credentials, or response bodies.

## Actual behavior

The source-level reproduction rejects at Scope resolution for all entrypoints listed above. A successful Scope resolution followed by a Memory 404 returns `{ok:false,code:"not_found",status:404,...}` without its domain code.

The existing five focused test files for tools, commands, invocation, Scope, and diagnostics pass all 23 tests, so the regression needs coverage through actual registered entrypoints rather than only direct calls to the lower-level invocation helper.

## Environment

- PowerContext source: master at [f0f288a](https://github.com/oceanbase/powercontext/commit/f0f288abecaccb97e1fe97d991b87b808bbebfbd).
- DSH plugin: `integrations/dsh/plugins/powercontext`.
- Reproduction: Node.js 24.14.1 on Windows, using registered handlers and controlled HTTP responses.
- Evidence is a source-level reproduction. The exact deployed endpoint, plugin revision, and DSH version from the original user report remain unverified.

## Additional context

Relevant source: [tool entrypoint](https://github.com/oceanbase/powercontext/blob/f0f288abecaccb97e1fe97d991b87b808bbebfbd/integrations/dsh/plugins/powercontext/src/tools.ts#L52-L61), [command entrypoint](https://github.com/oceanbase/powercontext/blob/f0f288abecaccb97e1fe97d991b87b808bbebfbd/integrations/dsh/plugins/powercontext/src/commands.ts#L128-L135), and [error mapping](https://github.com/oceanbase/powercontext/blob/f0f288abecaccb97e1fe97d991b87b808bbebfbd/integrations/dsh/plugins/powercontext/src/invoke.ts#L62-L80).

Reuse #1401's Scope contract and #1360's diagnostic infrastructure. #1438 owns HTTP request encoding changes; #1398 owns access-control behavior. #1449's default-Scope onboarding is independent.

This issue excludes automatic hook diagnostic wiring and snapshots, expanded CLI Doctor probes, the full status panel, and guidance changes; those remain in #1450 B–F. Include focused regression tests, one real-Server DSH acceptance case, maintained English/Chinese documentation, and rebuilt checked-in plugin artifacts.

AI assistance: OpenAI Codex assisted with source inspection, reproduction, and this report.

knqiufan:

Closes #1451.

Refs #1450 (work package A).

# Rationale for this change

Scope resolution currently runs outside the error boundary used for DSH HTTP operations. A failed resolver request can therefore escape as a host exception from a named tool or block `/pc doctor` before its health checks run. Operation-level 404 results also discard the public reason, making a missing Scope or Memory difficult to distinguish from an unavailable required route.

- Catch failures across Scope resolution and direct operation execution, and report them through the existing bounded diagnostics mechanism. Diagnostic writer exceptions and rejected callbacks do not replace the operation result.
- Preserve `not_found` for business 404s and add optional `error_code` for recognized public reasons. A 404 becomes `version_mismatch` only when the original business code is absent and the endpoint is required, including Scope resolution. Keep the existing authentication, availability, conflict, and validation codes, and map detected invalid responses explicitly.
- Replace Server-provided failure messages with fixed text and omit unrecognized error codes from results and diagnostics.
- Dispatch doctor, capabilities, and local usage before Scope resolution. Keep bare status usable with `scope=unresolved`, controlled error details, a recovery hint, and a redacted Server origin.
- Pass the direct caller's cancellation signal through Scope resolution and stop cancelled calls before subsequent business dispatch. Retain the existing per-request timeout and Server-owned Scope selection rules.
- Add registered-tool/command regression coverage and a live-Server test for a nonexistent explicit Scope, retain default-Scope coverage, update both DSH guides, and rebuild the checked-in bundle.

The implementation builds on #1401, #1360, and #1371. It leaves the transport/encoding work in #1438 and Scope onboarding in #1449 to their existing owners. Automatic-hook diagnostics, broader Doctor probes, and guidance/status work remain in #1450 B–F.

Yes. Scope failures return controlled failures instead of escaping to DSH. Doctor and capabilities work independently of Scope availability; usage help is local. Bare `/pc` reports an unresolved Scope honestly and displays only the Server origin.

Failure results retain available `status` and `request_id`; business 404s can additionally expose a safe `error_code`. Existing success results and specific public 409/422 codes remain compatible. Failed Scope resolution does not write Memory, create bindings, or select another Scope. No HTTP API contract or persisted-format change is required.

# How was this change tested?

The initial registered-entry regression suite reproduced the defect on the base revision: 24 failures and 3 passes. The completed suite passes.

GNU make is unavailable on the local Windows host, so the Makefile recipes were executed directly in PowerShell:

- `make js-test` recipe: frozen pnpm installation, unit tests, build, generated-operation/bundle `git diff --exit-code`, unit tests again, and live-Server e2e — **117 unit tests and 9 e2e tests passed**, with no generated-output drift.
- `make check` recipe: generated integration-manifest documentation check, **26 integration-manifest tests**, `uv lock --locked`, `uv run prek run -a`, `uv run ty check`, and the Pydantic AI integration type check — **passed**.
- `make docs-test` recipe: strict clean Zensical build with the canonical OpenAPI copy, API-page existence check, and published-contract byte equivalence — **passed**.
- `git diff --cached --check` — **passed**.

Regression coverage includes resolver and operation failures, required-route versus business 404s, authentication, timeout, cancellation, unresolved Scope, safe error output, diagnostic writer failures and cooldown, Scope-independent commands, and exact-Scope stats. The live-Server case verifies that a nonexistent explicit Scope fails direct operations while doctor/capabilities remain usable.

Upstream CI also passed `dsh-package` (`make js-test`), `quality` (`make check` and `make contract-test`), `check-docs` (`make docs-test`), and the license-header check on commit `c776cddf`.

# AI usage statement

OpenAI Codex assisted with investigation, implementation, regression tests, documentation, and validation.

### DIFF
--- a/docs/en/docs/how-to/configure-dsh.md
+++ b/docs/en/docs/how-to/configure-dsh.md
@@ -36,6 +36,35 @@ workspace therefore uses the Server default instead of the Harness process direc
 
 The plugin calls `POST /v1/context/prepare` once before the model analyzes the prompt. Explicit `remember_memory` calls do not require a model.
 
+## Diagnose direct tool and command failures
+
+Named tools and Scope-dependent `/pc` commands return a controlled failure if Scope resolution fails. They stop before
+the requested operation, without creating a binding or retrying with another Scope. Cancellation and the existing
+per-request timeout also apply to Scope resolution.
+
+Inside DeepSeek Harness:
+
+- `/pc doctor` checks liveness and readiness independently of Scope resolution and reports both results.
+- `/pc capabilities` queries the Server capabilities without resolving a Scope.
+- Unknown subcommands and missing arguments return local usage help without contacting the Server.
+- Bare `/pc` shows the resolved Scope and Server origin. If resolution fails, it returns an error while still showing
+  `scope=unresolved`, a controlled error, and the `/pc doctor` recovery hint. Configured Scope IDs are not reported as
+  resolved. The displayed origin omits credentials, paths, query strings, and fragments.
+- `search`, `remember`, `flush`, `review`, `skills scan`, and `stats` require a resolved Scope. `stats` queries that Scope.
+
+| Result code | Meaning |
+| --- | --- |
+| `not_found` | A business 404. The optional `error_code` preserves a recognized public reason, such as `scope_not_found` or `memory_not_found`. |
+| `version_mismatch` | A required endpoint returned 404 without a business code. Check the Server endpoint and plugin/Server compatibility; this does not establish a particular deployment cause. |
+| `authentication_failed` | The Server returned 401. Check the configured Authorization header. |
+| `unavailable` | Connection failure, timeout, cancellation, or HTTP 503. Native diagnostics use `server_unavailable`. |
+| `unscoped` | The resolver completed without a Scope. |
+| `invalid_response` | The client detected an invalid Server response. |
+
+Existing conflict and validation codes, such as `revision_conflict` and `invalid_request`, retain their meaning. Failure
+results preserve available HTTP status and request ID, but use fixed messages instead of Server-provided text. Unknown
+error codes are omitted from `error_code` and diagnostics; their presence alone does not imply a version mismatch.
+
 ## Control prompt capture
 
 Prompt capture is enabled by default. Disable it before starting DeepSeek Harness when the current work must not be recorded:
--- a/docs/zh/docs/how-to/configure-dsh.md
+++ b/docs/zh/docs/how-to/configure-dsh.md
@@ -36,6 +36,33 @@ Server 管理的 Scope。workspace 路径只会哈希为外部 binding key。缺
 
 插件在模型分析提示词前只调用一次 `POST /v1/context/prepare`。显式 `remember_memory` 不需要模型。
 
+## 排查工具和命令的直接调用失败
+
+Scope 解析失败时，具名工具和依赖 Scope 的 `/pc` 命令会返回受控失败，并在执行请求的操作前停止。
+插件不会因此创建 binding 或换用其他 Scope 重试。取消信号和现有的单请求超时也适用于 Scope 解析。
+
+在 DeepSeek Harness 内：
+
+- `/pc doctor` 不依赖 Scope 解析，继续检查 liveness 和 readiness，并保留两个检查结果。
+- `/pc capabilities` 直接查询 Server 能力，无需解析 Scope。
+- 未知子命令或缺少参数时，在本地返回用法说明，不访问 Server。
+- 裸 `/pc` 显示已解析的 Scope 和 Server origin。解析失败时返回错误，但仍显示 `scope=unresolved`、受控错误信息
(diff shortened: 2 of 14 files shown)


Teingi (approved):
LGTM

PR #1458, opened 2026-09-04
docs(rfc): standardize source access for artifact generation

Zxf-xufeng:
## Which issue or RFC does this close?

Follow-up design for #1437. The RFC defines the generation-side Source access rules required by the Source and Artifact REST API design.

## Why is this change needed?

Artifact Create and Replace persist a system-owned `lineage_only` Source for provenance. That Source must remain readable through management paths, but it must never become evidence for Memory, Experience, Skill, Handoff, Candidate review, or a future Artifact generator. Today those consumers combine repository reads and eligibility checks independently, which allows behavior to drift.

## What changed?

- add matching English and Chinese RFC documents;
- define separate semantics for explicit SourceRef admission and bounded Source journal windows;
- require Memory, Experience, Skill, Handoff, Candidate review, and future Source-consuming families to use the shared generation access path;
- document public Source reads, management writes, Recall, publication, and persistence target validation as non-generation boundaries;
- specify batching, cursor behavior, malformed payload handling, neutral errors, transaction boundaries, migration order, and validation coverage;
- add both RFC pages to the Zensical navigation.

## User-facing or breaking changes

This PR changes documentation only. It adds no HTTP endpoint, public Source type, role, database column, table, or index. Existing public Source Get remains a read operation and does not enforce generation eligibility.

## Validation

- `git diff --check`
- `uv run zensical build --clean -s`

## AI usage

AI was used to compare the design with the current repository, draft the bilingual RFC, and check terminology and navigation. The design decisions and final content were reviewed by the author.


Teingi (approved):
LGTM

PR #1461, opened 2026-09-04
feat(generation): standardize source access

Zxf-xufeng:
## Which issue or RFC does this close?

Implements #1458. This is a stacked change on #1438 and should be reviewed after the foundational Source and Artifact REST API implementation.

## Why is this change needed?

Artifact Create and Replace persist `lineage_only` Content Sources as durable provenance. Those records must remain available to public and management reads, but they must not feed Artifact generation. The existing Memory, Experience, Skill, Handoff, and Candidate paths read Sources independently, so eligibility and Source-window behavior can drift or be omitted by a new family.

## What changed?

- add `SourceRepository.get_many` for ordered, deduplicated exact reads and `list_window` for SQL-bounded `(after, through]` journal reads;
- add a lightweight generation Source access layer that combines strict repository decoding with the existing eligibility rule;
- migrate Memory Flush and Experience incubation to the same bounded-window filtering behavior while preserving full cursor advancement;
- migrate Experience and Skill generation, Candidate Propose/Revise/Approve, Handoff citation resolution, and Handoff activation to the shared explicit admission path;
- batch built-in Handoff Source and Artifact evidence reads while preserving compatibility with existing custom evidence resolvers;
- keep runtime/public Source Get independent from generation eligibility;
- allow the existing internal Content Source target to carry any valid Artifact family string without adding a new field or Source type;
- use a neutral `source_not_eligible` message and expose only the submitted SourceRef in error details;
- add repository, malformed-payload, explicit rejection, window skip, cursor advancement, Candidate atomicity, public read, and Handoff regression coverage.

## User-facing or breaking changes

- Explicit generation with a `lineage_only` Source now fails atomically with `422 source_not_eligible` before model or Artifact writes.
- Journal consumers silently exclude valid `lineage_only` records and still advance through the selected window; an all-filtered window is a no-op.
- Public and management Source reads can still return the Source. No HTTP success schema, Source type, database table, column, or index is added.

## Validation

- `make check`
- focused Source/Memory/Experience/Skill/Review/Handoff tests: 56 passed
- clean-environment full suite: 1327 passed, 19 skipped, 1 local-service-dependent test deselected; the Linux-only receiver-service test file was excluded on macOS

## AI usage

AI was used to trace existing Source consumers, implement the shared access path and repository methods, draft regression tests, and review compatibility boundaries. The code, behavior, and test results were reviewed by the author.


src/powercontext/builtin/runtime/relational.py
```diff
@@ -1035,9 +1039,7 @@ async def get(self, source: Source, /) -> Source:
         ref = self._as_ref(source)
         try:
             async with self._database.connection(self._bound_connection) as connection:
-                value = (await self._repository.get(connection, self._scope_id, ref)).value
-                require_source_eligible(ref, value)
-                return value
+                return (await self._repository.get(connection, self._scope_id, ref)).value
```
Teingi, line 1133: [P2] Keep generation admission on explicit Memory extraction

The same `SourceCatalog` is still used by `MemoryService._canonical_operation_evidence()`, so this change also removes admission from `context.artifacts.memory.remember(memory=None, sources=(lineage_source,), mode="extract")`. I reproduced this with the `lineage_only` Source created by Artifact Create: the base raises `SourceNotEligibleError` before extraction; this head passes the full Source to the extraction pipeline, and a pipeline returning `()` makes the call succeed. Please wire the explicit Memory extraction path through `GenerationSourceAccess` before invoking the pipeline, while keeping public Source reads independent.

src/powercontext/builtin/persistence/handoff.py
```diff
@@ -136,30 +136,73 @@ def __init__(
...
+) -> HandoffCitation:
+    identity = error.identity
+    missing = identity[-1] if isinstance(identity, tuple) and identity else None
+    for citation in citations:
+        if isinstance(citation, HandoffSourceCitation) and citation.source_ref == missing:
+            return citation
+        if isinstance(citation, HandoffArtifactCitation) and citation.artifact_ref == missing:
+            return citation
+    if citations:
+        return citations[0]
```
Teingi, line 207: [P2] Preserve the failing ArtifactRef in batch lookup errors

`ArtifactRepository.get_many()` can raise `RepositoryNotFoundError("artifact-family", family)` for an unregistered family. That identity is a string, so `_missing_citation()` falls back to `citations[0]`. With `(valid_source_citation, missing_artifact_citation)`, I reproduced `HandoffEvidenceUnavailableError.citation` pointing to the valid Source; the base correctly points to the missing Artifact. This gives SDK callers the wrong evidence to repair. Please retain or recover the failing `ArtifactRef` for this error case instead of choosing the first citation.

src/powercontext/builtin/runtime/relational.py
Zxf-xufeng: Fixed in bac14d78. Memory now uses a dedicated Source resolver that routes evidence reads through GenerationSourceAccess.require_for_generation, while ordinary Source catalog reads remain independent of generation admission.
Added regression coverage for both explicit extract and auto selecting extraction. A lineage-only Source remains readable, but extraction raises SourceNotEligibleError before the pipeline runs, including when the pipeline would return an empty result. No new Artifact revision is created. Both cases reproduce the failure on the reviewed commit and pass with the fix.

src/powercontext/builtin/persistence/handoff.py
Zxf-xufeng: Fixed in bac14d78. ArtifactRepository.get_many() now preserves (scope_id, ArtifactRef) when family validation fails, allowing Handoff to identify the failing Artifact citation instead of falling back to the first citation.
Added regression coverage with a valid Source citation followed by an unavailable Artifact citation, covering both an unregistered family and a missing Artifact in a registered family. The unregistered-family case reproduces the incorrect citation on the reviewed commit; both cases now report the correct Artifact citation.

Teingi (approved):
LGTM

PR #1459, opened 2026-09-04
fix(dsh): complete automatic-path diagnostics and recall presentation

knqiufan:
# Which issue or RFC does this PR close?

Closes #1457. Refs #1450 (B). Builds on the merged fixes #1452 (A) and #1464 (Windows Server cursor-key fix).

This PR is rebased on upstream master `8550f9f9`, which contains #1398, #1452, #1453, and #1464. It contains B's implementation and acceptance records plus the independently reviewable macOS LaunchAgent start fix from #1480. That fix is included here so the same native CI path that exposed the failure can validate it; once #1480 merges, rebasing will drop the equivalent commit.
[PR comparison](https://github.com/knqiufan/powercontext/compare/8550f9f939f2903cc22d265bf629be06d04e8dfa...0436943a8bfb155c63fd7855030447b450248194).
Live-model acceptance is pending: the configured official DeepSeek endpoint returned HTTP 401.

# Rationale for this change

An ordinary DSH conversation can silently lose automatic Scope diagnostics or discard valid prepared context when capture or a diagnostic writer fails. Successfully injected context also lacks the host's snapshot metadata. The new registered-entry regressions reproduce these defects on A: 20 failures out of 28 initial tests.

B reuses A's error classification, endpoint semantics and Scope cancellation support, and completes their automatic pre-step integration. This fixes existing host contracts and requires no RFC.

# What changes are included in this PR?

- Report Scope/prepare/capture/flush/injection failures at their own boundaries. Logger exceptions and rejected promises are best effort; existing safe fields and outcome-based 60-second cooldown remain.
- Stop automatic work after Scope failure or overall cancellation. Preserve prepare/capture independence, accepted Source results, and the complete downstream decision, including `startsRequestSeries`.
- Attach one native snapshot with a PowerContext section whose text exactly matches the injected message. Preserve the Server content, citations, byte validation and untrusted-history framing through normal host persistence.
- Fix two compatibility defects exposed by actual host acceptance: omit absent `request_id` fields so tool output is lossless JSON, and declare command input so Web routes `/pc doctor` and `/pc capabilities` as commands.
- Add registered-entry regressions and a pinned real-DSH runtime test package. CI now runs `make dsh-runtime-test` after `make js-test`, loading the built distributable with real host services and a real PowerContext Server.
- Rebuild `lib`, document behavior in English/Chinese, and add reproducible runtime instructions plus an acceptance record. Fix Windows e2e Server teardown to stop its owned child process tree.
[...]

PR #1464, opened 2026-09-04
fix(server): preserve binary cursor keys on Windows

### LINKED ISSUE #1463: bug(server): preserve binary cursor signing keys on Windows
### Describe the bug

On Windows, creating a persistent REST cursor signing key can silently change its bytes. The Server initially starts successfully, but a subsequent start fails with `ValueError: cursor signing key must contain 32 bytes`.

`src/powercontext/server/cursor_secret.py` opens the key with `os.open` without `O_BINARY`. Windows text-mode writes translate each LF byte into CRLF. Generated keys containing LF are therefore longer on disk than the 32-byte value used by the first process. The binary reader rejects the file on restart.

This is a Server persistence defect introduced in #1438, found by the native Windows lifecycle check on #1459. It is independent of the DSH automatic-path changes. It affects both the file-backed SQLite key and the locally persisted key used without an explicit secret for other database configurations.

### Steps to reproduce

On Windows at upstream commit `fc75f9270cf4b493411e1c3bef9965bd3b313edb`, with the development dependencies installed, run:

```python
from pathlib import Path
from unittest.mock import patch

from powercontext.server.cursor_secret import resolve_cursor_secret

    root = Path(directory)
    database = SQLiteConfig(url=f"sqlite+aiosqlite:///{root / 'powercontext.db'}")
    with patch("powercontext.server.cursor_secret.secrets.token_bytes", return_value=bytes(range(32))):
        resolve_cursor_secret(database, None)
    print((root / ".powercontext.db.cursor-key").stat().st_size)
    resolve_cursor_secret(database, None)
```

The fixed input includes LF, making the failure deterministic. It changes neither filesystem access nor repository permissions and does not require a model or external database.

### Expected behavior

The persisted key contains exactly the generated 32 bytes. The Server can restart and accept pagination cursors issued before the restart.

### Actual behavior

The reproduction prints `33` and the next read raises the length validation error. The existing Server restart test also fails deterministically when its generated key contains LF.

[Native Windows CI failure](https://github.com/oceanbase/powercontext/actions/runs/33895076873/job/101095690131), artifact `windows-native-service-logs`, contains:

```text
16:26:31 PowerContext Server is ready
16:26:42 Started server process [2248]
ValueError: cursor signing key must contain 32 bytes: .../.powercontext.db.cursor-key
Application startup failed. Exiting.
```

The test assertion only reported an unreachable service. Randomly generated keys without LF avoid the defect, so rerunning can hide it without fixing it.

### Environment

- PowerContext: upstream `fc75f927` after #1438; also present in #1459 at `e3ad78d7`.
- Failed CI: `windows-latest`, Python 3.12.10, fresh wheel installation with `cli,server` extras.
- Deterministic local reproduction: Windows, Python 3.12.13.
- Existing generated keys of the correct length remain valid. Already malformed key files should not be silently replaced or rotated by the fix.

knqiufan:

Closes #1463.

Found while investigating the Windows native-service check on #1459. This fix is based directly on `master` and is independent of the DSH changes.

# Rationale for this change

On Windows, `os.open` defaults to text mode. Writing a random 32-byte cursor key containing LF expands it to CRLF on disk. The first Server process uses the original key and starts successfully; the next process reads the expanded file and fails startup because its length is no longer 32 bytes. Retrying CI can avoid the random LF byte and conceal the defect.

- Open new cursor-key files with `O_BINARY` where supported, preserving exclusive creation and existing POSIX file permissions.
- Make both SQLite and other-database local key persistence tests deterministic with control bytes, and verify the exact persisted bytes. Keep POSIX mode assertions on POSIX systems.
- Exercise actual HTTP pagination across Server lifespans using a key containing LF, proving that previously issued cursors remain usable.
- Run these regressions against the installed wheel in Windows CI, alongside the existing real Task Scheduler lifecycle tests. Print captured Server logs in the failure output before querying a task that test cleanup may already have removed.

Newly generated cursor keys retain their original bytes on Windows, allowing subsequent Server starts. Existing valid keys and cursor formats are unchanged. This does not automatically replace malformed key files or rotate signing keys.

# How was this change tested?

- Before the fix, the deterministic persistence/restart checks produced **3 failed, 1 passed** on Windows with the same length-validation error found in the CI artifact.
- `python -m pytest -q tests/test_cursor_secret.py tests/test_server.py`: **51 passed** on Windows/Python 3.12.13.
- Built a wheel and installed `powercontext[cli,server]` plus pytest into a fresh `.native-venv`; the exact new Windows CI regression command passed **4 tests**, including HTTP cursor reuse after restart.
- `uv lock --check`, `uv run --no-sync prek run -a`, `uv run --no-sync ty check`, and `git diff --check`: passed.
- A local attempt at the real native suite was denied at `schtasks /Create` by this host. The existing GitHub-hosted Windows lifecycle job remains required to validate install/stop/start/uninstall; its checks are not skipped or relaxed.
- Upstream CI at `2fb02a2e`: **17/17 checks passed**, including Python 3.11–3.14, SQLite/OceanBase acceptance, and native Linux/macOS/Windows services. The [Windows job](https://github.com/oceanbase/powercontext/actions/runs/33898013675/job/101105173539) passed all four deterministic cursor checks and all four applicable real lifecycle tests; its one skip is the macOS-only scenario.

# AI usage statement

Implemented with OpenAI Codex. The failure was traced to the uploaded CI service logs, reproduced deterministically on Windows before changing production code, and checked against both source and the built distribution. No credentials or private model configuration are included.

### DIFF
--- a/.github/workflows/native-service.yml
+++ b/.github/workflows/native-service.yml
@@ -116,6 +116,13 @@ jobs:
           uv venv .native-venv
           uv pip install --python .native-venv\Scripts\python.exe "powercontext[cli,server] @ $wheelUri" pytest
 
+      - name: Verify binary cursor persistence and restart compatibility
+        shell: pwsh
+        run: >-
+          .native-venv\Scripts\python.exe -m pytest -q
+          tests/test_cursor_secret.py
+          tests/test_server.py::test_server_reuses_file_backed_cursor_secret_across_restarts
+
       - name: Exercise the real Task Scheduler lifecycle
         shell: pwsh
         run: >-
@@ -127,6 +134,13 @@ jobs:
         if: failure()
         shell: pwsh
         run: |
+          Get-ChildItem -LiteralPath "$env:RUNNER_TEMP\powercontext-native-tests" -Recurse -File -ErrorAction SilentlyContinue |
+            Where-Object { $_.Name -in @('server.stdout.log', 'server.stderr.log') } |
+            ForEach-Object {
+              Write-Output "::group::$($_.FullName)"
+              Get-Content -LiteralPath $_.FullName -Tail 100
+              Write-Output "::endgroup::"
+            }
           schtasks.exe /Query /TN "$env:POWERCONTEXT_NATIVE_SERVICE_IDENTIFIER" /XML /HRESULT
           if ($LASTEXITCODE -ne 0) { exit 0 }
           schtasks.exe /Query /TN "$env:POWERCONTEXT_NATIVE_SERVICE_IDENTIFIER" /FO LIST /V /HRESULT
--- a/src/powercontext/server/cursor_secret.py
+++ b/src/powercontext/server/cursor_secret.py
@@ -53,7 +53,9 @@ def _load_or_create(path: Path) -> bytes:
 
     generated = secrets.token_bytes(_CURSOR_SECRET_BYTES)
     try:
-        descriptor = os.open(path, os.O_WRONLY | os.O_CREAT | os.O_EXCL, 0o600)
+        # Windows text descriptors translate LF bytes even when using os.write.
+        flags = os.O_WRONLY | os.O_CREAT | os.O_EXCL | getattr(os, "O_BINARY", 0)
+        descriptor = os.open(path, flags, 0o600)
     except FileExistsError:
         return _read(path)
     try:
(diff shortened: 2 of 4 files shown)


Teingi (approved):
LGTM

PR #1467, opened 2026-09-04
docs(rfc): define unified artifact tags

### LINKED ISSUE #1466: RFC: Add unified tags for Artifacts and Memory entries
## Problem

Teams need to classify individual PowerContext-managed resources with custom tags such as a customer, product area,
scenario, or validation status, and then retrieve current resources by exact tag membership.

Encoding classification in Artifact content would create content Revisions and rebuild search projections even though
the reusable knowledge did not change. Existing Source and managed Skill `metadata` fields have different authorities
and version semantics. A tag on the single Memory Artifact for a Scope also cannot distinguish individual Memory
entries.

## Proposed direction

- Treat tags as mutable, Scope-local catalog attributes rather than Artifact content or lineage.
- Give managed Experience, Skill, Handoff, whole Memory Artifacts, and individual Memory entries one public tag-set
  model.
- Store all assignments in one polymorphic `pc_artifact_tags` table.
- Retain the owning `family + artifact_id`, then distinguish an Artifact target from a nested Memory-entry target.
- Bind tags to logical identities, not Artifact Revisions or Memory entry versions.
- Support exact `all` and `any` queries and apply tag eligibility before Memory FTS/vector top-k selection.
- Do not copy tags across Scope publication by default and never interpret tags as authorization or Agent
  instructions.

## First-delivery scope

- Read and compare-and-swap replacement of a complete tag set.
- Cross-family exact tag query with current exact Artifact or Memory citations.
- Optional tag filters for current Artifact listing, Memory-entry listing, and Memory search.
- A minimal Dashboard tag editor and filter.
- Deterministic Unicode normalization, bounded values, pagination, authorization filtering, and secret-safe
  observability.

## Non-goals

The first delivery does not include tag colors, hierarchy, aliases, a managed tag catalog, automatic model tagging,
historical tag snapshots, tag-based authorization, revision-specific tags, or arbitrary key/value Artifact metadata.

## Acceptance direction

The design must preserve tags across content revisions, isolate tags between Memory entries in the same Memory,
change no content digest or lineage, reject stale concurrent replacements, keep existing unfiltered behavior backward
compatible, and produce equivalent behavior on SQLite and OceanBase/seekDB profiles.

A bilingual RFC pull request will define the complete target, persistence, API, retrieval, security, migration, and
validation contracts.

Teingi:

Closes #1466.

# Rationale for this change

Users need stable custom labels for organizing and retrieving Memory, Experience, Skill, and Handoff content without
editing immutable Artifact revisions or overloading behavior-bearing metadata. Memory entries also need independent
labels even though the standard profile stores them inside one Memory Artifact per Scope.

- Add synchronized English and Chinese RFCs for unified Artifact tags.
- Define one `pc_artifact_tags` assignment table for whole Artifacts and logical Memory entries.
- Define exact tag normalization, `all`/`any` filtering, target identity, concurrency, authorization, publication,
  migration, and observability semantics.
- Define REST tag subresources and a cross-family query contract aligned with RFC 1437's Scope URI, ETag, and cursor
  conventions.
- Keep tag changes outside Artifact revisions, content digests, lineage, embeddings, authorization, and prompt input.

This PR contains design documentation only. It does not change the current API or persisted schema.

No runtime behavior changes in this PR. If implemented, the RFC introduces additive tag APIs and optional list/search
filters. The design declares no breaking change and requires existing unfiltered behavior and historical responses to
remain unchanged.

# How was this change tested?

- `make docs-test` with Node.js 22.22.3: passed, including lint, 545-page static generation, and export verification.
- Repository document hooks for case conflicts, merge conflicts, end-of-file, and trailing whitespace: passed.
- `git diff --cached --check`: passed.

The repository-wide `ty check` currently reports three pre-existing unresolved `scripts.scope_binding` imports in
Codex integration hooks; this documentation-only change does not modify those files.

# AI usage statement

OpenAI Codex (GPT-5) was used to inspect existing contracts, draft the bilingual RFC, and run validation. The author
reviewed the final design and delivery.

### DIFF
--- a/docs/en/rfcs/1467_artifact_tags.md
+++ b/docs/en/rfcs/1467_artifact_tags.md
@@ -291,10 +291,16 @@ The primary key supports loading one target's tags. The first secondary index su
 second supports a cross-family Scope query. Implementations must use binary identity comparison or application-built
 `tag_key` values rather than depending on database-default case or locale collation.
 
-One table cannot express conditional foreign keys to both `pc_artifact_heads` and `pc_memory_entry_heads`. The owning
-Artifact foreign key is enforced for every row. For `memory_entry`, the repository must additionally lock and validate
-the current `(scope_id, memory_artifact_id, entry_id)` head in the same transaction before changing assignments. The
-transaction rejects a missing entry. This is an explicit application invariant, not a best-effort cleanup rule.
+The owning Artifact foreign key is enforced for every row. For `memory_entry`, the repository must lock the owning
+`(scope_id, family="memory", artifact_id)` Artifact head, load the exact Revision it points to, and validate `entry_id`
+against that Revision's authoritative `MemoryContent.manifest.entries` in the same transaction before changing
+assignments. Both `active` and `inactive` manifest entries are valid targets. An entry absent from that manifest is
+rejected even if an older Revision or immutable entry-version row contains it.
+
+`pc_memory_entry_heads` contains only active search projections. Deactivation removes an entry's projection while
+retaining its logical identity and content in the authoritative manifest. Tag reads and mutations must not require a
+row in that table, and tag assignments must not reference it through a foreign key. Projection cleanup and rebuilding
+must leave tag assignments unchanged.
 
 Inactive Memory entries and deprecated or retired Artifacts retain their assignments. Tag queries apply the existing
 visibility and lifecycle selection before returning results. A caller with the target's write authority may reorganize
@@ -355,7 +361,8 @@ query(
 `replace` performs the following steps in one transaction:
 
 1. Resolve and authorize the target without returning hidden existence details.
-2. Lock the owning Artifact head. For a Memory entry, also lock and validate its current entry head.
+2. Lock the owning Artifact head and load the exact Revision it points to. For a Memory entry, validate `entry_id`
+   against that Revision's manifest, accepting either `active` or `inactive` state.
 3. Load the complete current tag set and calculate its digest.
 4. Reject a mismatched expected digest as a failed precondition.
 5. Validate and normalize the complete replacement set.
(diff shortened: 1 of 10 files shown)


docs/en/rfcs/1467_artifact_tags.md
```diff
@@ -0,0 +1,687 @@
...
+```
+
+The primary key supports loading one target's tags. The first secondary index supports family-specific filtering; the
+second supports a cross-family Scope query. Implementations must use binary identity comparison or application-built
+`tag_key` values rather than depending on database-default case or locale collation.
+
+One table cannot express conditional foreign keys to both `pc_artifact_heads` and `pc_memory_entry_heads`. The owning
+Artifact foreign key is enforced for every row. For `memory_entry`, the repository must additionally lock and validate
+the current `(scope_id, memory_artifact_id, entry_id)` head in the same transaction before changing assignments. The
+transaction rejects a missing entry. This is an explicit application invariant, not a best-effort cleanup rule.
```
Teingi: [P2] Validate entry existence against the current manifest

`pc_memory_entry_heads` only contains active search projections. I verified on this commit with SQLite that `forget()` removes the entry's head row while the entry remains readable in the current manifest; rebuilding projections does not restore that row. Requiring it here would therefore reject tag replacements for inactive entries, contradicting the following paragraph's promise that their tags can be reorganized without reactivation.

Please lock the owning Artifact head and validate `entry_id` against that Revision's authoritative manifest. The inactive-query path should also resolve citations from the manifest. Update both language versions and add an acceptance case for reading, replacing, and clearing tags after deactivation.

Teingi: Fixed in b962344. Both RFCs now require locking the owning Artifact head and validating `entry_id` against that Revision's authoritative manifest, accepting both active and inactive entries. Tag reads and mutations do not require a `pc_memory_entry_heads` row.

Inactive tag queries and entry listing resolve state and citations from the same manifest before pagination. The RFC acceptance criteria now cover reading, replacing, and clearing tags after deactivation, preserving assignments through projection rebuilds, and rejecting targets absent from the current manifest. The single-table design is unchanged.

Validation: document hooks, bilingual JSON examples and local links, and `make docs-test` passed. I also independently reproduced the deactivation and projection-rebuild behavior with SQLite on the reviewed commit.

Zxf-xufeng (approved):
LGTM

Issue #1465, opened 2026-09-04
feat: add scope-owned versioned prompt management


Teingi:
## Feature description

Add a Scope-owned Prompt Artifact family so each PowerContext Scope can customize the operational prompts used by Memory, Experience, Skill, and Handoff generation while retaining immutable revision history.

The initial registered prompt keys are:

- `memory.extract`
- `memory.rerank`
- `experience.incubate`
- `experience.generate`
- `skill.generate`
- `handoff.generate`

Each Prompt revision contains a mode, instructions, and typed demonstrations. A demonstration uses the formal fields `input` and `expected_output`; dashboard groupings such as positive and negative examples are presentation concerns rather than separate storage fields.

## Problem and proposed solution

The built-in Runtime currently binds versioned instructions when it is composed. Different Agents or users can be bound to different Scopes, but the Scope itself cannot select a custom prompt, inspect its history, or reproduce which revision controlled an operation.

Use the existing Scope and Artifact revision model instead of introducing prompt-specific persistence:

- add `prompt` as an Artifact family;
- use each registered prompt key as the stable `artifact_id`;
- reuse the existing create, current read, replace with `If-Match`, and exact revision read operations;
- add `GET /v1/scopes/{scope_id}/artifacts/{family}/{artifact_id}/revisions` for generic revision history;
- add `POST /v1/scopes/{scope_id}/prompts/{prompt_key}/demonstrations` to generate structured demonstration suggestions without saving them;
- implement rollback by reading an old immutable revision and replacing the current head with that content, producing a new monotonically increasing revision;
- resolve and freeze one exact Prompt revision at the start of each inference operation;
- preserve Server-owned safety rules, output schemas, credentials, tool authority, and model configuration outside editable Prompt content.

No Prompt revision in a Scope means the current built-in behavior remains unchanged. There is no implicit parent-Scope inheritance or Agent/User-specific storage in the core contract.

## Alternatives considered

- Dedicated Prompt, Prompt Version, and Prompt Binding tables duplicate ownership, revision, and head semantics already supplied by Artifact storage.
- Prompt-specific CRUD endpoints duplicate the generic Artifact API and make the surface harder to extend.
- Agent- or user-keyed Prompt configuration conflicts with Scope isolation; Agent and user integrations should resolve their existing Scope binding before calling the core API.
- Raw prompt replacement per inference request is difficult to authorize, audit, reproduce, and cache.
[...]

PR #1468, opened 2026-09-04
docs(rfc): define scope-owned prompt management

Teingi:
Tracks #1465.

# Rationale for this change

PowerContext binds seven built-in instruction variants across six inference operation types at Runtime composition.
Customers need Scope-specific operational prompts with immutable history, rollback, and traceable runtime selection.
This proposal reuses Scope and Artifact revision/head semantics: no new database tables and no parallel Prompt CRUD API.

- Add synchronized English and Chinese RFCs for Scope-owned operational Prompt management.
- Define `family=prompt`, registered prompt keys as stable Artifact IDs, and strict content containing `schema_version`,
  `mode`, `instructions`, and typed `demonstrations` (`input` plus `expected_output`).
- Reuse Artifact CRUD and exact reads; propose exactly two new HTTP operations: generic Artifact revision listing and
  non-persisting Prompt demonstration generation. Rollback copies an old revision into a new conditional replacement.
- Define compatible Definition upgrades, readable immutable history, explicit incompatible-head/rollback errors, and
  the v1 restriction to homogeneous deployment versions.
- Carry verified generation provenance across Handoff prepare/finalize/commit, retaining the selected Prompt revision
  after head changes and distinguishing edited drafts from unchanged generated content.
- Separate code-enforced schema, Scope, authority, and resource constraints from best-effort model guidance and secret
  exclusion, with explicit adversarial validation requirements.
- Extend existing capabilities with effective per-key support and deployed built-in identity. The Dashboard bundles
  the six keys and handles empty Scopes, disabled operations, and externally managed components without a discovery API.
- Exclude injected pipeline/generator/reranker implementations from v1 customization; reject unsupported Custom writes
  and block existing Custom heads after a component switch instead of silently ignoring them.
- Specify authorization, tracing, configuration lineage, errors, deployment requirements, and implementation acceptance
  scenarios, including the narrow amendments to RFC 1396 and RFC 1437.

This PR contains design documentation only. It does not implement the proposed APIs or Runtime behavior.

No released behavior changes in this PR. The proposed implementation retains current behavior for Scopes without
Prompt Artifacts, including injected components. It requires no new SQL tables or content backfill.

Implementing the RFC extends existing capabilities and Handoff request/response/content models, so strict readers,
generated contracts, mappers, and clients must be updated together. Old Handoff payloads without generation metadata
remain valid but do not carry verified Prompt provenance. V1 does not support mixed Definition/compiler versions
within one deployment.

# How was this change tested?

Passed:

- `git diff --check` and `git diff --cached --check`.
- Relative-link and Markdown fence validation for both RFCs.
- Parsing all 14 JSON examples and validating all 6 demonstration input/output pairs against the current
  `MemoryExtractionInput` / `MemoryExtractionOutput` models in strict JSON mode.
- Bilingual example-count/error-code checks and confirmation of exactly two new endpoint rows in each RFC.
- All repository `UV_PYTHON=3.11 uv run --locked prek run -a` hooks except the existing `ty-check` failure below.
- From `website/`, using its pinned pnpm 11.13.1: `corepack pnpm install --frozen-lockfile`, `corepack pnpm lint`, and
  `corepack pnpm build`.
- Static export verification: 78 HTTP API pages and 378 Python API pages.

The root `make docs-test` invocation selected pnpm 11.25.0 through Corepack and failed the project's version check;
the equivalent install/lint/build commands above passed from `website/`. The website does not render RFC files, so
RFC examples and links were checked separately.

Repository-wide `ty check` still reports three existing unresolved `scripts.scope_binding` imports in the Codex
integration hooks (`bind_tools.py`, `recall.py`, and `session_binding.py`). This Markdown-only PR does not change those
files. The proposed Runtime lifecycle and adversarial tests are implementation requirements, not tests implemented
or claimed as passing by this documentation PR.

# AI usage statement

OpenAI Codex assisted with repository and review inspection, drafting and revising the bilingual RFCs, and validation.

### DIFF
--- a/docs/en/rfcs/1468_scope_owned_prompt_management.md
+++ b/docs/en/rfcs/1468_scope_owned_prompt_management.md
@@ -60,10 +60,13 @@ small, explicit Runtime extension boundary.
 
 ## User model
 
-The Dashboard presents one configuration page for the current Scope. Each registered prompt key has two modes:
+The Dashboard presents one configuration page for the current Scope. Each supported built-in prompt key has two modes:
 
 - **Auto** uses the built-in instructions selected by the deployed Runtime; and
 - **Custom** uses Scope-owned instructions and demonstrations.
+
+Disabled operations and injected components that do not support customization are shown as unavailable, with a reason.
+They are not presented as configurable Auto/Custom operations merely because their keys are registered.
 
 For `memory.extract`, a custom editor can show three sections:
 
@@ -88,7 +91,8 @@ For example, the stored content is:
         "evidence": [
           {
             "evidence_id": "source-1",
-            "text": "Before every release I run the core smoke test; it takes about 20 minutes."
+            "evidence_type": "source",
+            "content": "Before every release I run the core smoke test; it takes about 20 minutes."
           }
         ],
         "current_entries": []
@@ -106,7 +110,13 @@ For example, the stored content is:
     },
     {
       "input": {
-        "evidence": [{"evidence_id": "source-2", "text": "Please send me the code from the previous message."}],
+        "evidence": [
+          {
+            "evidence_id": "source-2",
+            "evidence_type": "source",
+            "content": "Please send me the code from the previous message."
+          }
+        ],
         "current_entries": []
       },
       "expected_output": {"candidates": []}
@@ -172,8 +182,10 @@ Content-Type: application/json
 }
 ```
 
-The Server uses `prompt_key` as `artifact_id` and commits revision 1. Unknown or disabled prompt keys are rejected.
-Callers cannot allocate arbitrary operational Prompt identities.
+The Server uses `prompt_key` as `artifact_id` and commits revision 1. Unknown keys are rejected; Custom writes also
+require that the effective component supports customization. Auto writes remain available for registered keys so an
+operator can clear a custom selection even when its operation is unavailable. Callers cannot allocate arbitrary
+operational Prompt identities.
 
 Updating a Prompt uses the existing conditional replacement operation:
 
@@ -214,11 +226,11 @@ Content-Type: application/json
 ```json
 {
   "instructions": "Keep durable testing preferences and verified failure lessons.",
-  "demonstration_count": 6
+  "demonstration_count": 1
 }
 ```
 
-The response contains exactly six schema-valid demonstrations:
+The response contains exactly one schema-valid demonstration:
 
 ```json
 {
@@ -232,9 +244,9 @@ The response contains exactly six schema-valid demonstrations:
 }
 ```
 
-The shortened array above illustrates the item shape only. The endpoint does not create or replace an Artifact,
-advance a head, or silently combine its output with current content. The caller reviews and edits the suggestions, then
-uses the normal Artifact create or replace operation to save them.
+The endpoint does not create or replace an Artifact, advance a head, or silently combine its output with current
+content. The caller reviews and edits the suggestions, then uses the normal Artifact create or replace operation to
+save them.
 
 ## Inspect history and roll back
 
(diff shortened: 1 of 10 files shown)


docs/en/rfcs/1468_scope_owned_prompt_management.md
```diff
@@ -0,0 +1,688 @@
...
+`PromptContent` is strict and rejects unknown fields:
+
+```python
+class PromptDemonstration(BaseModel):
+    input: JsonValue
+    expected_output: JsonValue
+
+
+class PromptContent(BaseModel):
+    schema_version: Literal["powercontext.prompt.v1"]
```
Teingi, line 374: **[P2] Clarify compatibility across Definition changes.** Demonstrations are validated against the active Definition during both writes and resolution. If its input/output schema changes incompatibly, an existing custom head may fail validation, and copying an old revision for rollback may be rejected. Please define how existing heads and rollback behave across such changes, including mixed-version deployments if supported. Binding revisions to a Definition version is one option; a documented compatibility policy with explicit handling of incompatible content is also reasonable.

docs/en/rfcs/1468_scope_owned_prompt_management.md
```diff
@@ -0,0 +1,688 @@
...
+configured administrative Principal. A family-specific writer enforces this stronger mutation rule instead of
+weakening all generic Artifact writes.
+
+The Runtime's internal use of the current Scope Prompt is part of the already-authorized domain operation. It is not an
+implicit exact-resource share. Operational Prompt revisions are not shareable through `prompt.user` in this RFC, and
+the Prompt access profile advertises no grantable exact roles for them.
+
+Custom Prompt content is untrusted configuration. It cannot:
+
+- override evidence-as-data treatment, citation requirements, secret exclusion, or identity allocation rules;
```
Teingi: **[P2] Distinguish enforced constraints from model guidance.** Keeping invariant instructions immutable does not guarantee that the model always follows them. For example, the current Memory output schema validates structure but does not determine whether arbitrary candidate text contains secrets. Please clarify which guarantees are enforced by code and which depend on model compliance. Describe semantic secret exclusion as best-effort unless a concrete enforcement mechanism and its limits are specified. Adversarial tests would help assess this boundary; schema validation alone does not demonstrate a model bypass or credential leak.

docs/en/rfcs/1468_scope_owned_prompt_management.md
```diff
@@ -0,0 +1,688 @@
...
+- `powercontext.prompt.key`;
+- `powercontext.prompt.selection` as `built_in` or `artifact`;
+- exact Artifact family, ID, and revision when custom;
+- built-in and Definition versions;
+- compiled prompt digest; and
+- demonstration count.
+
+Prompt bodies and demonstration bodies are not logged or emitted as metric labels.
+
+When a durable generated Artifact already records Artifact inputs, its lineage includes the exact custom Prompt
```
Teingi, line 494: **[P2] Define Prompt provenance across the Handoff lifecycle.** The RFC promises the exact custom Prompt ref in generated Artifact lineage, but `prepare`, `finalize`, and `commit` are separate requests, and the current `HandoffDraft` and `PreparedHandoff` models have no place for that selection. Reading the Prompt head again at commit could record a different revision. Please specify how generation metadata travels through this lifecycle and what it means after the caller edits the draft. Cover a Prompt head change between prepare and commit; the RFC can leave the specific storage or transport mechanism to implementation.

docs/en/rfcs/1468_scope_owned_prompt_management.md
```diff
@@ -0,0 +1,688 @@
...
+- enable tools, network, filesystem access, publication, or another capability;
+- resolve or reference another Scope implicitly; or
+- cause generated demonstrations to be saved without an explicit Artifact write.
+
+The demonstration generator treats supplied instructions as data within a Server-owned meta-prompt. Generated
+examples undergo the same secret and size validation as manually supplied demonstrations.
+
+## Non-HTTP surfaces
+
+The Dashboard uses the HTTP operations above. Its Agent or user selector resolves to a Scope before it reads or writes
```
Teingi, line 655: **[P3] Clarify how the Dashboard presents effective configuration.** The six stable keys could be bundled with the Dashboard, so an empty Artifact list in a fresh Scope does not by itself require a discovery endpoint. It would still help to explain how the UI identifies unavailable operations and, if displayed, the deployment-selected built-in profile. Existing capabilities or bundled metadata may be sufficient for v1. A new discovery API is only needed if the intended UI requires information those sources cannot provide.

docs/en/rfcs/1468_scope_owned_prompt_management.md
```diff
@@ -0,0 +1,688 @@
...
+selection chosen today, including the configured `memory_extraction_profile` and rerank enablement.
+
+Implementation requires:
+
+1. splitting each current instruction constant into immutable invariants and replaceable default guidance without
+   changing Auto-mode compiled behavior;
+2. registering the six Prompt Definitions;
+3. adding `prompt` to the base Artifact family contract, generated HTTP models, mapper, repository type registry, and
+   family management writer registry;
+4. resolving Prompt state per Scope at the six inference entry points;
```
Teingi: **[P2] Clarify support for injected Runtime components.** `open_builtin_runtime` accepts injected candidate, Experience, Skill, Handoff, and reranker implementations whose current protocols do not take a resolved Prompt. Since this RFC targets built-in inference operations, please state whether those injected implementations are excluded and how Prompt availability is determined for their keys. An explicit unsupported status with rejection of custom configuration would be sufficient for v1. A prompt-aware extension protocol is only needed if these components are intended to participate; test the chosen behavior.

docs/en/rfcs/1468_scope_owned_prompt_management.md
Teingi: Added a Handoff lifecycle section in ce3ea949 in both RFCs. The proposed generation envelope carries a Server-verifiable receipt from prepare through finalize and commit, preserving the exact generation-time Prompt selection. A draft generated with revision 2 still records revision 2 if the head advances to revision 3 before commit. Edited drafts retain their origin but are marked edited, not presented as unchanged model output. The RFC also specifies missing/tampered receipt handling, provenance-aware no-op comparison, and acceptance coverage across separate requests, using existing Artifact JSON storage.

docs/en/rfcs/1468_scope_owned_prompt_management.md
Teingi: Clarified in ce3ea949 in both RFCs. The Dashboard bundles the six stable keys and uses an extension to the existing GET /v1/capabilities response for effective support status, unavailability reasons, and the deployed built-in profile/version. It combines that metadata with the Scope's Artifact list, so a fresh Scope can show Auto without stored Prompt rows. Unavailable components remain visible with disabled customization controls. No discovery endpoint is added; the proposed new endpoint count remains two.

docs/en/rfcs/1468_scope_owned_prompt_management.md
Teingi: Specified in ce3ea949 in both RFCs. Injected pipelines, generators, and rerankers are outside v1 managed customization; their existing protocols stay unchanged. Availability is based on the effective component, not just the presence of a model. Unsupported components reject Custom writes and demonstration generation. If a Scope already has a Custom head when the component is replaced, invocation fails until the operator explicitly selects Auto or restores a compatible built-in component. Absent/Auto configurations preserve injected behavior. These cases are now explicit implementation acceptance requirements.

docs/en/rfcs/1468_scope_owned_prompt_management.md
Teingi: Revised in ce3ea949 in both RFCs. The trust-boundary section now separates code-enforced schema, Scope, authority, and resource constraints from instructions that depend on model compliance. Semantic secret exclusion is explicitly best-effort; the Memory output schema is not described as a general secret detector. The validation requirements now include model-backed adversarial checks with synthetic secret-like evidence, without treating schema acceptance as proof of a bypass or claiming those tests have already been implemented.

docs/en/rfcs/1468_scope_owned_prompt_management.md
Teingi: Clarified in ce3ea949 in both RFCs. The Definition compatibility section now requires backward-compatible demonstration contracts under the same key and content schema. History remains readable without active-Definition revalidation. An incompatible rollback returns 422 without advancing the head; an incompatible existing custom head fails the operation with 503 rather than silently selecting Auto. The RFC also requires upgrade preflight and explicitly excludes mixed-version workers from v1. Restoring Prompt content does not promise to reproduce the old model or compiler output.

Zxf-xufeng (approved):
LGTM

PR #1470, opened 2026-09-05
fix(persistence): stop leaking sqlite-vec rows on Memory commits

### LINKED ISSUE #1469: bug: sqlite-vec embeddings leak on every Memory commit and crowd live entries out of vector search
### Describe the bug

I was testing vector search on the SQLite backend and noticed that a search in one scope started returning nothing after I had appended a few entries to a memory in a *different* scope. Nothing was deleted, no error, the hit just disappeared.

Digging in: every Memory commit leaves the old embeddings behind in `pc_memory_entry_vec`. `RelationalMemoryBackend._commit` deletes the `pc_memory_entry_heads` rows first and only then calls `index.replace(...)`. But `pc_memory_vector_entries` has a foreign key to the heads with `ondelete="CASCADE"` (and `foreign_keys` is on by default in the SQLite profile), so by the time `SQLiteMemoryVectorIndex.replace` runs, the metadata rows are already gone. It looks up the vec0 rowids through that metadata, finds nothing, and never deletes the old vectors. vec0 is a virtual table, so nothing else cleans it up.

Two effects:

- `pc_memory_entry_vec` grows on every commit. One entry plus four appends already leaves 10 orphan rows for 5 live entries.
- Search results become wrong. `search()` uses `count(pc_memory_vector_entries)` as the KNN `k`, but the KNN runs over the whole vec0 table including the orphans. If orphans are closer to the query they take the k slots, get dropped by the join, and the live entry that should have matched is gone. That is what I was seeing across scopes.

`rebuild_projections` doesn't fix an affected database either, since it also finds vec0 rows only through the metadata table. Interestingly it already does the ordering right (clears the index *before* deleting the heads); `_commit` just doesn't.

Related: #1321 is about storage growth from re-storing the manifest on every revision. This is a separate growth source that only shows up with an embedding model configured, and it also affects results, so I'm filing it separately.

### Steps to reproduce

From a checkout of master (`fc75f92`), `uv sync`, then save this as `repro_sqlite_vec_orphans.py` and run `uv run python repro_sqlite_vec_orphans.py`. It uses a fake keyword embedding model, so no provider is needed.

```python
#!/usr/bin/env python3
"""Reproduce sqlite-vec rows leaking on every Memory commit.

Run inside a PowerContext checkout with dependencies installed:

    uv run python repro_sqlite_vec_orphans.py
"""

from __future__ import annotations

import asyncio
import math

from powercontext.builtin.artifacts.memory import EmbeddingProfile, MemoryEntryInput
from powercontext.builtin.inference import EmbeddingResult
from powercontext.builtin.runtime import BuiltinConfig, open_builtin_contexts

PROFILE = EmbeddingProfile(profile_id="repro-v1", model="repro", dimension=3, distance="l2", normalization="unit")
_DELTA_NORM = math.sqrt(0.1**2 + 1.0)


class KeywordEmbeddingModel:
    profile = PROFILE

    async def embed(self, texts: tuple[str, ...], /) -> EmbeddingResult:
        def vector(text: str) -> tuple[float, float, float]:
            lowered = text.casefold()
            if "alpha" in lowered:
                return (1.0, 0.0, 0.0)
            if "gamma" in lowered:
                return (0.0, 0.0, 1.0)
            if "delta" in lowered:  # near gamma, but never an exact match
                return (0.0, 0.1 / _DELTA_NORM, 1.0 / _DELTA_NORM)
            return (0.0, 1.0, 0.0)

        return EmbeddingResult(vectors=tuple(vector(text) for text in texts))


async def row_counts(contexts) -> tuple[int, int, int]:
    async with contexts.database.transaction() as connection:
        live = (await connection.exec_driver_sql("SELECT count(*) FROM pc_memory_vector_entries")).scalar()
        vectors = (await connection.exec_driver_sql("SELECT count(*) FROM pc_memory_entry_vec")).scalar()
        orphans = (
            await connection.exec_driver_sql(
                "SELECT count(*) FROM pc_memory_entry_vec "
                "WHERE rowid NOT IN (SELECT vector_id FROM pc_memory_vector_entries)"
            )
        ).scalar()
        return int(live), int(vectors), int(orphans)


async def main() -> None:
        config = SQLiteConfig(url=f"sqlite+aiosqlite:///{directory}/repro.db")
        async with open_builtin_contexts(BuiltinConfig(database=config), embedding_model=KeywordEmbeddingModel()) as contexts:
            quiet = (await contexts.get("scope-quiet")).artifacts.memory
            busy = (await contexts.get("scope-busy")).artifacts.memory

            target = await quiet.remember(
                memory=None, entries=(MemoryEntryInput(kind="fact", text="Delta record."),), mode="append"
            )
            before = await quiet.search("gamma", memories=(target,), mode="vector")
            print("scope-quiet vector search before writes elsewhere:", [hit.text for hit in before.hits])

            churned = await busy.remember(
                memory=None,
                entries=(MemoryEntryInput(kind="fact", text="Gamma one."), MemoryEntryInput(kind="fact", text="Gamma two.")),
                mode="append",
            )
            for index in range(4):
                churned = await busy.remember(
                    memory=churned, entries=(MemoryEntryInput(kind="fact", text=f"Alpha record {index}."),), mode="append"
                )

            live, vectors, orphans = await row_counts(contexts)
            print(f"pc_memory_vector_entries rows: {live}")
            print(f"pc_memory_entry_vec rows:      {vectors}  (orphans without metadata: {orphans})")

            after = await quiet.search("gamma", memories=(target,), mode="vector")
            print("scope-quiet vector search after writes elsewhere: ", [hit.text for hit in after.hits])


if __name__ == "__main__":
    asyncio.run(main())
```

The script puts one entry ("Delta record.", a vector close to "gamma") in `scope-quiet`, does five appends in `scope-busy`, and searches `scope-quiet` for "gamma" before and after.

### Expected behavior

- `pc_memory_entry_vec` has exactly one row per row in `pc_memory_vector_entries`, no matter how many commits happened.
- The "gamma" search in `scope-quiet` returns `['Delta record.']` both times. Writes elsewhere should not change it.

### Actual behavior

```
scope-quiet vector search before writes elsewhere: ['Delta record.']
pc_memory_vector_entries rows: 7
pc_memory_entry_vec rows:      21  (orphans without metadata: 14)
scope-quiet vector search after writes elsewhere:  []
```

If I set `SQLiteConfig(..., foreign_keys=False)` the same script gives 7 / 7 / 0 and the second search still finds the entry, so it really is the cascade.

### Environment

- PowerContext master `fc75f92`
- Python 3.12.3, sqlite-vec 0.1.9, aiosqlite 0.22.1, SQLite 3.45.1
- Ubuntu 24.04

AbyssWhalen:

Closes #1469.

# Rationale for this change

`RelationalMemoryBackend._commit` deletes the active-head rows before asking the index to replace the projections. The sqlite-vec metadata table (`pc_memory_vector_entries`) cascades from those heads, and the vec0 rows in `pc_memory_entry_vec` are only reachable through that metadata, so every commit leaves the previous revision's embeddings behind as orphans. Growth is quadratic in the number of appends: one entry plus four appends leaves 10 orphans for 5 live entries.

The vector search passes the live metadata count as the KNN neighbor limit while the KNN scans the whole vec0 table, so orphans that were nearer to the query displaced live entries. In practice, appends in one scope made vector searches in another scope return nothing. `rebuild_projections` could not repair it because it also finds vec0 rows only through the metadata. The ordering issue predates the sqlite-vec migration in #1306; the Vec1 search used the same `k = total` limit.

- `RelationalMemoryBackend._commit` clears the index for the memory before deleting the heads, the same order `rebuild_projections` already uses, so the metadata is still present when the vec0 rows are removed.
- `SQLiteMemoryVectorIndex.initialize` deletes vec0 rows that have no metadata row, so databases written by earlier versions recover on the next start instead of carrying the orphans forever.
- Regression tests: `tests/e2e/test_sqlite_vec.py` checks one embedding per live entry across appends and that a vector search is unaffected by writes in other scopes; `tests/builtin/persistence/test_sqlite_memory_vector_index.py` checks that `initialize` drops embeddings without metadata.

No API, schema, persisted-format, or documentation changes. The OceanBase and seekdb backends are unaffected: their vector table has no foreign key to the heads and `replace` deletes by scope and artifact.

Behavioral fix only. Vector and hybrid Memory search on the SQLite backend no longer lose hits after unrelated writes, and the SQLite file stops growing with stale embeddings. The first start after upgrading runs one pass over `pc_memory_entry_vec` to drop existing orphans, which may take a moment on long-lived databases. No breaking changes.

# How was this change tested?

- Reproduced on `master` (`fc75f92`) with the script in the issue: 7 live entries left 21 vec0 rows (14 orphans) and a cross-scope vector search went from `['Delta record.']` to `[]`. With `foreign_keys=False` the leak does not occur, confirming the cascade as the cause. With the fix: 7 / 7 / 0 and the search still returns the entry.
- The three new tests fail on `master` and pass with the fix.
- `uv run python -m pytest --doctest-modules`: full suite green (1324 passed, 21 skipped).
- `uv run ruff check`, `uv run ruff format --check`, `uv run ty check`: clean.

# AI usage statement

AI-assisted; the change and tests were reviewed and validated by the author.

### DIFF
--- a/src/powercontext/builtin/persistence/memory.py
+++ b/src/powercontext/builtin/persistence/memory.py
@@ -469,6 +469,10 @@ class RelationalMemoryBackend:
                 insert(MEMORY_ENTRY_VERSIONS_TABLE),
                 [_entry_values(self._scope_id, entry) for entry in value.entry_versions],
             )
+        # Clear the index before the heads go, as rebuild_projections does: index
+        # metadata may cascade from the heads, and an index can only find its
+        # rows through that metadata.
+        await self._index.replace(connection, self._scope_id, value.memory.as_ref(), ())
         await connection.execute(
             delete(MEMORY_ENTRY_HEADS_TABLE).where(
                 MEMORY_ENTRY_HEADS_TABLE.c.scope_id == self._scope_id,
--- a/src/powercontext/builtin/persistence/sqlite/memory_index.py
+++ b/src/powercontext/builtin/persistence/sqlite/memory_index.py
@@ -152,6 +152,9 @@ _INSERT_FTS_SQL = text(
     """
 )
 _DELETE_VECTOR_SQL = text("DELETE FROM pc_memory_entry_vec WHERE rowid = :vector_id")
+_DELETE_ORPHAN_VECTORS_SQL = (
+    "DELETE FROM pc_memory_entry_vec WHERE rowid NOT IN (SELECT vector_id FROM pc_memory_vector_entries)"
+)
 _INSERT_VECTOR_SQL = text("INSERT INTO pc_memory_entry_vec (rowid, embedding) VALUES (:vector_id, :embedding)")
 _SELECT_VECTOR_SQL = text("SELECT embedding FROM pc_memory_entry_vec WHERE rowid = :vector_id")
 _VECTOR_SEARCH_SQL = text(
@@ -347,6 +350,9 @@ class SQLiteMemoryVectorIndex:
             raise CapabilityNotSupportedError("vector", detail) from error
         if row is None or int(row[0]) != -1:
             raise CapabilityNotSupportedError("vector", "sqlite-vec probe returned an invalid row")
+        # Embeddings are only reachable through their metadata rows. Drop any that
+        # lost theirs, so stale neighbors cannot take the place of live entries.
+        await connection.exec_driver_sql(_DELETE_ORPHAN_VECTORS_SQL)
 
     async def replace(
         self,
--- a/tests/builtin/persistence/test_sqlite_memory_vector_index.py
+++ b/tests/builtin/persistence/test_sqlite_memory_vector_index.py
@@ -62,6 +62,28 @@ def test_vector_index_probe_clears_a_leftover_probe_row(tmp_path) -> None:
     asyncio.run(scenario())
 
 
+def test_vector_index_initialize_drops_embeddings_without_metadata(tmp_path) -> None:
+    async def scenario() -> None:
+        async with (
+            SQLiteProfile.open(
+                SQLiteConfig(url=f"sqlite+aiosqlite:///{tmp_path / 'memory.db'}"),
+                tables=SQLITE_MEMORY_VECTOR_TABLES,
+                load_vector_extension=True,
+            ) as profile,
+            profile.database.transaction() as connection,
+        ):
+            await connection.exec_driver_sql("CREATE VIRTUAL TABLE pc_memory_entry_vec USING vec0(embedding float[3])")
+            await connection.execute(
+                _INSERT_VECTOR_SQL,
+                {"vector_id": 7, "embedding": _pack_vector((0.0, 0.0, 1.0))},
+            )
+            await SQLiteMemoryVectorIndex(_profile(3)).initialize(connection)
+            remaining = (await connection.exec_driver_sql("SELECT count(*) FROM pc_memory_entry_vec")).scalar()
+            assert int(remaining) == 0
+
+    asyncio.run(scenario())
+
+
 def test_vector_index_probe_reports_a_table_dimension_mismatch(tmp_path) -> None:
     async def scenario() -> None:
         async with (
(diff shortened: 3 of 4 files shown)


Teingi (approved):
LGTM

PR #1480, opened 2026-09-07
fix(service): kickstart newly bootstrapped LaunchAgents

### LINKED ISSUE #1479: bug(service): kickstart newly bootstrapped LaunchAgents
## Describe the bug

On macOS, `powercontext service install` and the restart path can register the per-user LaunchAgent successfully but leave the Server process stopped. The controller then waits for the HTTP endpoint until it times out, even though no Server process was started.

`LaunchdUserAdapter.start()` currently treats a successful `launchctl bootstrap` as an immediate start. `bootstrap` registers the job; `launchctl kickstart` is the operation that explicitly asks launchd to run it immediately. Relying only on `RunAtLoad` and `KeepAlive` makes the existing start contract nondeterministic.

This has appeared repeatedly in the native macOS lifecycle CI, both during the initial install and after an explicit stop/start:

- https://github.com/oceanbase/powercontext/actions/runs/34075169684/job/101599749920
- https://github.com/oceanbase/powercontext/actions/runs/33935919038
- https://github.com/oceanbase/powercontext/actions/runs/33865926348

The same #1459 commit passed the same runner image shortly before the latest failure, which rules out a DSH behavior regression and points to this launchd start race.

## Steps to reproduce

1. On macOS, ensure the PowerContext LaunchAgent is not loaded.
2. Run `powercontext service install`, or stop an installed service and start it again.
3. Observe the native lifecycle until the configured local HTTP endpoint becomes live.
4. On affected runs, `launchctl bootstrap` returns successfully but the endpoint never becomes reachable and the service produces no startup logs.

The repository's `tests/native/test_personal_service_lifecycle.py::test_native_personal_service_lifecycle` exercises both paths and has reproduced the failure intermittently in CI.

## Expected behavior

A successful install or explicit start must ask launchd to start the registered job immediately. If launchd rejects the start, the command should surface that error directly instead of waiting for an HTTP timeout.

## Actual behavior

A newly bootstrapped job is not explicitly kickstarted. Some runs remain registered without a running process until the controller reports that the service did not become live. In the initial-install failure, both Server stdout and stderr logs are empty because the process never starts.

## Environment

- PowerContext: current `master` and PR #1459 merge commit
- Python: 3.13 in native-service CI
- OS: macOS 26.6.2 arm64, GitHub Actions image `20260831.0337.3`

knqiufan:

Closes #1479.

# Rationale for this change

The macOS service adapter currently relies on `launchctl bootstrap` to both register and immediately start a LaunchAgent. Registration can succeed without a process becoming active, so installs and explicit restarts intermittently wait for the HTTP endpoint until they time out. This has reproduced in several native macOS workflow runs, including #1459.

- Explicitly call `launchctl kickstart` after bootstrapping a new or replaced LaunchAgent.
- Preserve the existing `kickstart -k` behavior for a loaded but inactive job.
- Add a regression test that verifies a newly bootstrapped job is immediately kickstarted.

The plist's `RunAtLoad` and `KeepAlive` settings remain responsible for login startup and recovery. This change makes an explicit install/start request deterministic and surfaces a launchd start error immediately.

On macOS, `powercontext service install` and explicit service restart now reliably start the personal Server immediately after registration. There are no public API, configuration, or persisted-format changes.

# How was this change tested?

- `uv run pytest -q tests/test_service.py -k launchd` — 23 passed, 1 skipped
- `uv run pytest -q tests/test_service.py tests/test_service_bootstrap.py -k 'not unsafe_retry_state'` — 96 passed, 5 skipped
- `uv run prek run -a`
- `git diff --check`

The repository's native-service workflow will exercise the complete install, stop, restart, and uninstall lifecycle on a real macOS LaunchAgent.

# AI usage statement

OpenAI Codex was used to inspect the failing CI artifacts and historical runs, identify the launchd lifecycle race, implement the change, and run the listed validation. The resulting diff and test evidence were reviewed before submission.

### DIFF
--- a/src/powercontext/service/adapters/launchd.py
+++ b/src/powercontext/service/adapters/launchd.py
@@ -205,6 +205,10 @@ class LaunchdUserAdapter:
             is_loaded = False
         if not is_loaded:
             self._run("bootstrap", self._domain, str(self.artifact_path))
+            # bootstrap registers the job, while kickstart is the launchd
+            # operation that guarantees an immediate start regardless of the
+            # configured launch conditions.
+            self._run("kickstart", self._target)
         elif self.manager_state() is not ManagerState.ACTIVE:
             self._run("kickstart", "-k", self._target)
 
--- a/tests/test_service.py
+++ b/tests/test_service.py
@@ -1086,6 +1086,27 @@ def test_launchd_stop_waits_until_bootout_removes_the_loaded_job(
     run.assert_called_once_with("bootout", "gui/501/com.oceanbase.powercontext")
 
 
+def test_launchd_start_kickstarts_a_newly_bootstrapped_job(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    adapter = LaunchdUserAdapter(home=tmp_path, uid=501)
+    monkeypatch.setattr(
+        adapter,
+        "loaded_registration",
+        lambda: ManagerRegistration(ManagerOwnershipState.NOT_LOADED),
+    )
+    run = Mock()
+    monkeypatch.setattr(adapter, "_run", run)
+
+    adapter.start(reload_definition=False)
+
+    assert run.call_args_list == [
+        (("bootstrap", "gui/501", str(adapter.artifact_path)), {}),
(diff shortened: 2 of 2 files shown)


Teingi (approved):
LGTM

PR #1485, opened 2026-09-07
docs(rfc): propose profile artifacts and subject source writes

Zxf-xufeng:

RFC-only proposal; no tracking issue.

# Rationale for this change

Applications need to keep a user's message in a shared conversation Scope while accumulating the same evidence in an ordinary Scope addressed by a business subject key. Each Scope should be able to generate its own Profile and use existing Artifact APIs.

- Add matching English and Chinese RFCs for the `profile` Artifact Family, with one Profile snapshot per Scope.
- Resolve `subject_key` through existing Scope Bindings (`integration=subject`, `kind=user`). Subject writes create or reuse a binding and reject an explicit Scope mismatch; management APIs retain rebind/unbind behavior.
- Define one atomic subject Source endpoint that writes matching Source IDs and content to the business and subject Scopes, while preserving existing single-Scope endpoints.
- Add only `pc_profile_policies`; reuse existing Artifact, Candidate, Source Cursor, Scope Binding, and scheduler tables, with concrete row examples and server-owned BLOB metadata.
- Define configurable daily generation at 02:00 Asia/Shanghai, bounded Source windows, concurrent commit checks, review recovery, and rejection that consumes the window without retry.
- Reuse Artifact CRUD and Candidate review. Add only subject Source creation, Policy GET/PUT, and Profile Flush. Extend existing Binding Resolve with backward-compatible `allow_default` to prevent subject lookups falling back to a default Scope.
- Specify Python types, authorization, transaction boundaries, compatibility, limitations, and acceptance criteria.

This is a design proposal only; it changes no runtime code, OpenAPI contract, or persisted database schema. If implemented, it adds opt-in Profile behavior and a subject dual-write convenience endpoint. Existing binding management and single-Scope Source behavior are preserved. Multiple subjects may share a Scope; rebinding does not migrate historical data.

# How was this change tested?

- `git diff --check`
- Bilingual heading, fence, table-count, and local-link checks; Python snippet syntax and JSON example parsing.
- `CI=true make docs-test` (website lint, build, and static export verification).

# AI usage statement

Codex was used to draft, translate, inspect repository contracts, and validate the RFC. The design incorporates the author's explicit decisions on Scope Binding behavior, subject writes, table reuse, and review semantics.

### DIFF
--- a/docs/en/rfcs/0000_profile_artifact.md
+++ b/docs/en/rfcs/0000_profile_artifact.md
@@ -1,0 +1,2054 @@
+- Proposal Name: `profile_artifact`
+- Start Date: 2026-09-07
+- RFC PR: [oceanbase/powercontext#0000](https://github.com/oceanbase/powercontext/pull/0000)
+- Related RFCs: [RFC 0014](0014_memory_layer_design.md), [RFC 0019](0019_local_source_memory_runtime.md),
+  [RFC 0050](0050_artifact_candidate_review_inbox.md), [Topic Memory RFC](0000_topic_memory.md),
+  [RFC 1345](1345_scope_organization_and_agent_integration.md),
+  [RFC 1396](1396_handoff_access_control.md), and [RFC 1437](1437_source_artifact_rest_api.md)
+
+# Summary
+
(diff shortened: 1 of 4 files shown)


Teingi (approved):
LGTM

PR #1487, opened 2026-09-07
feat(profile): implement scoped profiles and subject source writes

Zxf-xufeng:

Implements #1485. This stacked PR targets `codex/profile-artifact-rfc`; retarget to master after the RFC merges.

# Rationale for this change

Provide Scope-based Markdown profiles and an atomic subject-addressed Source convenience without introducing a special
Root Scope model or duplicating existing Artifact, Candidate, lineage, and cursor storage.

- Add only `pc_profile_policies`; retain existing shared-table columns, indexes, and constraints.
- Resolve `subject_type=user` and caller-owned `subject_key` through existing Scope Bindings.
- Add atomic business/subject Source writes with matching Source IDs, insert-only binding resolution, explicit conflicts,
  transactional new-Scope access bootstrap, and retry-safe target selection.
- Keep existing Source APIs single-scope; add exact Binding resolution with `allow_default=false`.
- Register the `profile/profile` singleton in generic Artifact Create/Get/List/Get Revision/Replace.
- Store normalized Markdown and server-owned generation metadata in Artifact BLOBs; preserve immutable revisions,
  system provenance, ETag preconditions, and rollback through Replace.
- Add bounded Source-window generation, no-change detection, policy/cursor/head CAS, and lineage-only filtering.
- Reuse Candidate APIs for review. Approval atomically commits the profile and consumes its window; rejection consumes
  the window without retry. Manual edits invalidate stale approval but do not prevent rejection.
- Schedule daily at 02:00 Asia/Shanghai with configurable cron/timezone, startup catch-up, bounded concurrency,
  and existing APScheduler sidecar storage. Generation owners and Candidate attestations share the commit transaction.
- Add four HTTP operations and Python Client methods; regenerate Python and JavaScript contracts.
- Add Chinese/English guides and storage, concurrency, lifecycle, authorization, HTTP, scheduler, and opt-in OceanBase tests.

The four new operations are subject Source Create, Profile Policy GET/PUT, and Profile Flush.
User scopes are ordinary scopes; other Artifact families keep their existing APIs and generation behavior.
Binding management retains rebind/unbind and many-to-one behavior; no automatic migration or cross-Scope deletion is added.
Newly created subject scopes default to automatic Profile generation. Existing scopes require explicit policy changes.

This implementation replaces the unpublished Subject Root prototype: special Root/projection/processing/metadata tables
are no longer created, the singleton ID is `profile`, and subject writes have their own endpoint.
Existing experimental databases must be backed up and migrated separately; this PR does not delete tables or rewrite
old experimental Profile payloads automatically. Published single-scope Source semantics are preserved.

# How was this change tested?

- `make check`: passed; lock consistency, pre-commit hooks, main and Pydantic AI integration type checks.
- `make contract-test`: 43 passed.
- `CI=true make docs-test`: passed; 91 HTTP API pages and 384 Python API pages verified.
- Profile storage/scheduler/HTTP tests: 21 passed, 1 skipped (live OceanBase URL not configured).
- Full suite with doctests and external service environment variables removed: 1422 passed, 20 skipped,
  4 deselected.
- The four exclusions are existing host-dependent tests: three Linux systemd user-service cases on macOS and
  `test_default_doctor_uses_client_server_url_from_environment`, which detects this machine's failed personal-service
  registration. The implementation does not modify that service.
- MySQL DDL/restore-layer tests passed. A live OceanBase contract test is provided but was not run without
  `POWERCONTEXT_TEST_OCEANBASE_URL` for a disposable test database.
- No live LLM/provider calls were used for validation.

# AI usage statement

Codex was used to implement the change and run validation. Human code review is still required.


src/powercontext/builtin/artifacts/profile/service.py
```diff
@@ -0,0 +1,356 @@
...
+                candidate_id = None
+                candidate = None
+                artifact = None
+                if markdown is not None and policy.activation_mode == "review_required":
+                    candidate_id = self.id_factory("candidate")
+                    candidate = await self.candidates.create(
+                        connection,
+                        scope_id,
+                        candidate_id,
+                        "profile",
```
Teingi, line 243: [P2] Support Profile candidates in the Dashboard review flow

With `activation_mode=review_required`, this creates a pending Profile candidate that the Dashboard lists but cannot edit, approve, or reject. I reproduced this in Firefox: `review.js:isSupportedCandidate()` only accepts Experience and Skill, so the page shows "This Candidate shape is not supported by this page. Decisions are disabled." The pending candidate then blocks subsequent Profile windows until someone uses the API directly. Please add Profile support to the review page's candidate validation, Markdown display/editor, and decision controls.

src/powercontext/server/authz/profiles.py
```diff
@@ -102,6 +102,17 @@ def subject_compatibility(self) -> dict[AccessRole, frozenset[str]]:
         selector="forbidden",
         mutation_semantics=frozenset({AccessAction.ARTIFACT_WRITE}),
     ),
+    "profile": ArtifactFamilyAccessProfile(
+        family="profile",
+        enabled=True,
+        share_unit="artifact",
+        shareable_states=frozenset({"committed"}),
```
Teingi, line 109: [P2] Wire Profile into HTTP sharing and resource discovery

The HTTP sharing/discovery paths still reject this newly enabled family. With a committed Profile, granting `artifact.viewer` through `/v1/access/bindings/create` returns 422 because `_validate_shareable_resource()` falls through to `artifact-family-disabled`. A Scope viewer can read the Profile directly, but `/v1/access/resources/list` omits it; specifying `family=profile` returns 422 because `_discover_scope_artifact_resources()` also lacks a Profile branch. Please add Profile to both helpers and the default family enumeration so sharing and discovery match the advertised access profile.

src/powercontext/server/app.py
```diff
@@ -1924,6 +1941,116 @@ async def create_source(
...
+        principal = _require_principal()
+        context = _access_audit_context(CREATE_SUBJECT_SOURCE.operation_id)
+        if new_binding:
+            await access.require(
+                principal, AccessAction.SERVER_ADMIN, ResourceRef.server(access.deployment_id), context=context
+            )
+        if new_scope:
+            await access.bootstrap_subject_scope(connection, principal, target, context=context)
+        else:
+            await access.require(principal, AccessAction.SCOPE_CONTRIBUTE, ResourceRef.scope(target), context=context)
```
Teingi, line 2036: [P2] Bootstrap the static preset for the actual subject Scope

In a static-bearer deployment, creating a Scope and immediately supplying it as `subject_scope_id` returns 403 here. Reading that target's `/profile-policy` first, even when the read returns 404, makes the identical subject write succeed with 201. The normal authorization dependency materializes the static preset only for the business Scope, while this target check skips that initialization. This makes valid writes depend on whether the target was previously accessed. Please apply the existing static-preset bootstrap to the actual target before checking contribution permission.

src/powercontext/builtin/artifacts/profile/service.py
Zxf-xufeng: Thanks for catching this. Fixed: the Review Dashboard now recognizes Profile candidates, includes Profile in the family filter and deep-link handling, displays the Markdown content, and supports content-only revisions as well as approve/reject decisions. Editing remains gated by the server-provided can_revise permission, so this change does not broaden the generic Candidate authorization model.

src/powercontext/server/authz/profiles.py
Zxf-xufeng: Fixed. Profile is now included in both explicit and default Artifact resource discovery, and committed Profile artifacts are accepted by the HTTP resource validation used when creating artifact.viewer bindings. I also added E2E coverage for granting access and discovering Profile resources with and without an explicit family filter. This only enables access sharing; Profile copy/publication remains explicitly rejected.

src/powercontext/server/app.py
Zxf-xufeng: Fixed. When an existing or explicitly supplied subject Scope is used in static-bearer mode, the static preset is now materialized through the transaction-bound access repository. AccessControlService.with_connection() also preserves the configured static principal, preventing the bootstrap from becoming a no-op. This avoids an independent nested transaction and preserves the atomicity of the dual Source write. A regression test now covers creating a Scope and immediately using it as subject_scope_id without first accessing that Scope.

Teingi: [P2] Recheck target contribution after bootstrapping the static preset

The first-use case now works, but `78fba9e` introduces a revocation regression. After establishing the subject binding, revoke the static principal's `scope.contributor` binding on the target Scope through `/v1/access/bindings/revoke`. With the same token, a direct Source POST returns 403, while `subject-sources` returns 201 and persists a Source in that target. On `8416af6`, both requests returned 403.

`bootstrap_static_scope()` replays the existing revoked binding without reactivating it, and the new static-preset branch then skips `access.require()`. Please still check `scope.contribute` for the actual target after bootstrap, with authorization reads using the same transaction so they can see newly created bindings. Add a regression case that revokes the target contribution grant and verifies that the dual write is rejected without persisting either Source.

Teingi (approved):
LGTM

PR #1489, opened 2026-09-07
feat(context): add configurable prepared text assembly

### LINKED ISSUE #1488: feat: add configurable standard text assembly for prepared context
### Feature description

Allow callers of `POST /v1/context/prepare` to request standard Markdown context with explicit Memory/Experience selection, section order, per-family limits, and optional metadata. Preserve the existing `PreparedContext` response envelope and default output for callers that omit the new `assembly` option.

### Problem and proposed solution

Prepared context currently interleaves Memory and Experience and embeds a JSON payload in its content string. A troubleshooting agent cannot prioritize Experience, a coding agent cannot select only Memory, and users cannot easily inspect the exact text delivered to a host.

Add an optional request configuration, for example:

```json
{
  "assembly": {
    "format": "markdown",
    "sections": [
      {"family": "experience", "limit": 2},
      {"family": "memory", "limit": 5}
    ],
    "show": ["confidence", "recall_rank"]
  }
}
```

Section order also determines byte-budget priority. Entries retain their existing retrieval order within a family. The server renders the final text with exact Scope, Artifact revision, and MemoryEntry version citations, visible truncation markers, and an untrusted-history boundary. Hosts validate and inject the returned text without trimming it again.

The current data does not provide assessed confidence scores. Display confidence as `unknown (not assessed)` when requested; do not equate recall rank with confidence or introduce confidence-based filtering or sorting.

### Acceptance criteria

- Omitting `assembly` preserves legacy output; explicit options select only the requested families. An empty section list returns empty context after authorization.
- Family order, limits, complete citations, Unicode handling, and the final UTF-8 byte budget are enforced by the server.
- Referenced Scopes retain read authorization and owner-readiness requirements.
- Existing integrations keep their defaults and can explicitly opt into standard text through their settings.
- No new database table or HTTP endpoint is required.
- Include English and Chinese RFCs and usage documentation, contract and regression tests, and an opt-in acceptance scenario using a configured database, LLM, embedding model, and native Codex plugin.

### Alternatives considered

- Formatting independently in each plugin duplicates selection and byte-budget behavior and makes delivered context inconsistent across hosts.
- Replacing the default format would change existing integrations without an explicit opt-in.
- Adding persisted presets or numeric confidence scoring expands the scope beyond a configurable, inspectable text assembly contract.

Teingi:

Closes #1488.

RFC 1489: [English](https://github.com/Teingi/powercontext/blob/codex/prepared-context-text-assembly/docs/en/rfcs/1489_prepared_context_text_assembly.md) / [中文](https://github.com/Teingi/powercontext/blob/codex/prepared-context-text-assembly/docs/zh/rfcs/1489_prepared_context_text_assembly.md).

# Rationale for this change

Prepared context currently interleaves Memory and Experience in an embedded JSON payload. Callers cannot request only Memory, prioritize Experience, or control per-family output limits. This change adds an explicit `assembly` option to the existing prepare request so a host can receive readable, cited Markdown with the requested section order and metadata.

- Add the OpenAPI assembly contract, regenerated Python models, shared client support, and runtime selection/rendering. Preserve the existing `PreparedContext` envelope and legacy output when `assembly` is omitted.
- Support Memory and Experience sections, per-family limits, existing retrieval order within each family, and optional recall rank and confidence labels. Confidence is explicitly `unknown (not assessed)`; numerical confidence scoring, filtering, and sorting are outside this contract.
- Enforce the final UTF-8 byte budget while retaining exact Scope, Artifact revision, and MemoryEntry version citations, explicit truncation, and a literal untrusted-history boundary.
- Check referenced Scope read access and owner readiness against the same Scope snapshot used for recall. Keep empty selection and denied-reference behavior covered by regressions.
- Add opt-in settings to the Codex, Claude Code, WorkBuddy, DSH, Pi, OpenCode, Hermes, OpenClaw, LangChain, LangGraph, Pydantic AI, and Bub integrations. Preserve complete standard text through host validation and include assembly settings in relevant recall cache keys.
- Include bilingual RFCs and usage guides, behavioral and compatibility tests, and an opt-in configured-service/native-Codex acceptance test.

Callers can supply `assembly.sections` to select families, set their order and limits, and supply `assembly.show` to display confidence or recall rank. An empty section list returns empty context after authorization. Invalid options, including explicit `assembly: null`, are rejected.

Existing plugin configurations retain their output format. Using the new option requires a server that supports it. No database tables, migrations, or HTTP endpoints are added.

# How was this change tested?

- `uv run python -m pytest --doctest-modules -q --maxfail=3 --tb=short`: **1,447 passed, 17 skipped**.
- `make check`, `make contract-test` (**43 passed**), `make api-generate-check`, and `make docs-test`: passed.
- Integration checks: DSH **119 unit + 9 native CLI E2E**, OpenCode **24**, Pi **49**, OpenClaw **31**, and Bub harness **28** passed; associated build/type checks passed.
- `uv run python -m pytest tests/e2e/real_experience_skill/test_context_text_assembly.py --run-real-e2e --real-e2e-env-file=.env --real-codex-timeout=240 -q -s --tb=short`: **1 passed**, covering configured OceanBase vector recall and embedding, real LLM Experience generation and approval, exact lineage, section selection/order, byte budgets, legacy compatibility, Scope isolation, and native Codex consumption.
- The native Codex positive case used the installed plugin to retrieve versioned context, repair a deployment configuration, and pass the unchanged release-gate test. A separate session with empty sections returned no historical evidence.

The configured database had an older access-table schema, so real acceptance used an isolated temporary database on the same OceanBase service. Cleanup verified that the temporary database was removed and the user's Codex configuration was unchanged. Credentials and generated run artifacts are excluded from the PR.

# AI usage statement

Implemented, documented, and validated with OpenAI Codex (GPT-6), including code generation, test execution, and real-service acceptance testing.


integrations/langchain/src/powercontext_langchain/client.py
```diff
@@ -25,7 +25,7 @@
 from pydantic import SecretStr
 
 from powercontext.client import PowerContextClient
-from powercontext.http import ResolveScopeBindingRequest
+from powercontext.http import ContextAssembly, ResolveScopeBindingRequest
```
Teingi: [P2] Align adapter dependencies with the new core API

This unconditional import breaks standalone adapter installs even when assembly is disabled. With the published `powercontext==0.2.0` wheel, which satisfies the adapters' declared dependencies, the base versions of the LangChain, LangGraph, and Pydantic AI adapters all import successfully; all three PR versions fail with `ImportError: cannot import name 'ContextAssembly' from 'powercontext.http'`. Please require a core version that provides this API and update the source-install instructions to install a compatible core alongside the adapter, or preserve compatibility with older core versions when assembly is disabled.

integrations/hermes/plugins/powercontext/provider.py
```diff
@@ -633,6 +634,51 @@ def _trace_command(self, args: list[str]) -> str:
...
+    def _prepare_options(self) -> dict[str, Any]:
+        options: dict[str, Any] = {
+            "max_bytes": _as_int(
+                _config_value(self._config, "max_bytes", "POWERCONTEXT_HERMES_MAX_BYTES", _DEFAULT_MAX_BYTES),
+                _DEFAULT_MAX_BYTES,
+                minimum=512,
+                maximum=32768,
+            )
+        }
+        raw = _config_value(self._config, "context_assembly", "POWERCONTEXT_HERMES_CONTEXT_ASSEMBLY", None)
```
Teingi, line 646: [P2] Apply Hermes assembly settings to manual prepare calls

These options are only used by the prefetch paths. The `powercontext_prepare_context` tool goes through `commands.request_operation()`, which adds `max_bytes` but omits the configured `assembly`. With `context_assembly={"sections":[]}`, I verified the resulting requests against this PR's SQLite-backed HTTP app: automatic recall returns `empty`, while the manual tool's request returns `ready` with Memory content. A single-family selection is likewise lost. Please reuse `_prepare_options()` in the manual prepare path and cover empty and single-family selections there.

integrations/langchain/src/powercontext_langchain/client.py
Teingi: Fixed in 134b94ca. The three adapters now remain compatible with `powercontext==0.2.0` when assembly is unset, and reject unsupported assembly settings with an explicit upgrade message. Added regression tests and documented installing the core and adapter together from the same checkout. Verified imports, legacy request construction, and unsupported-assembly errors against the published 0.2.0 wheel.

integrations/hermes/plugins/powercontext/provider.py
Teingi: Fixed in 134b94ca. `powercontext_prepare_context` and `/pc call prepare_context` now reuse `_prepare_options()`, preserving configured assembly and byte-budget defaults while allowing explicit request overrides. Regression tests cover provider/environment configuration, empty sections, and Memory-only selection. Verified empty, Memory-only, and Experience-only selections against the SQLite-backed HTTP app: manual tools, slash commands, and automatic recall return consistent results.

integrations/bub/src/powercontext_bub/plugin.py
```diff
@@ -35,7 +35,7 @@
 
 from powercontext.client import InvalidResponseError, PowerContextClient, ServerResponseError, TransportError
 from powercontext.client.capture import render_capture_event
-from powercontext.http import CaptureContentSourceRequest, FlushMemoryRequest, PrepareContextRequest
+from powercontext.http import CaptureContentSourceRequest, ContextAssembly, FlushMemoryRequest, PrepareContextRequest
```
Teingi: [P2] Keep Bub usable when text assembly is disabled

This import makes the Bub plugin fail to load with `powercontext==0.2.0`, which still satisfies its `powercontext[client]` dependency. I tested the base and PR versions against that published core wheel: the base plugin imports successfully, while the PR version fails when loading the `bub` entry point with `ImportError: cannot import name 'ContextAssembly' from 'powercontext.http'`, even with assembly unset. `tools.py` has the same unconditional import.

Please use the optional-import handling from the other Python adapters in both places, reject assembly settings with an upgrade message when the type is unavailable, and add Bub to the standalone compatibility regression coverage.

Teingi: Fixed in fcf311d5. Bub now uses optional assembly-type handling in both import paths, so the plugin and tools load with `powercontext==0.2.0` when assembly is unset. Explicit assembly settings on that core produce a clear upgrade message.

Added standalone regression coverage for the Bub entry point, automatic recall, manual context calls, and argument/environment configuration with and without assembly support. Verified against the published 0.2.0 wheel with Bub 0.4.3; all 30 harness tests and `make check` pass.

Zxf-xufeng (approved):
LGTM

PR #1417, opened 2026-09-02
docs(rfc): propose topic memory artifacts

frf12:
# Which issue or RFC does this PR close?

This PR proposes a new RFC and does not close an existing tracking issue.

# Rationale for this change

PowerContext needs a topic-oriented memory artifact that can evolve from Sources while exposing compact search hits and loading complete detail only on demand. The RFC also defines the background processing and concurrency model required for long-running generation in SQLite and multi-replica OceanBase deployments.

# What changes are included in this PR?

- Add the English Topic Memory RFC.
- Add the equivalent Chinese Topic Memory RFC.
- Define the content model, progressive retrieval, generation pipeline, indexing barrier, Pending/Cursor semantics, Supervisor fencing, APIs, configuration, and explicitly deferred boundaries.

# Are there any user-facing changes?

Documentation only. This PR proposes future HTTP, MCP, retrieval, and deployment behavior; it does not implement or enable Topic Memory.

# How was this change tested?

- `make check`
- `uv run --no-sync zensical build --clean -s`
- Verified `site/api/index.html` exists and generated `site/api/openapi.yaml` matches `openapi/powercontext.yaml`.
- Cross-checked both locales for matching section structure, identifiers, API paths, defaults, and deferred boundaries.

# AI usage statement

Codex was used to organize the previously reviewed design, draft the bilingual RFC, translate the English version, and run validation. The design decisions were reviewed interactively by the author.

PR #1490, opened 2026-09-07
feat(topic-memory): add end-to-end topic memory

frf12:

Implements the Topic Memory RFC introduced in #1417.

# Rationale for this change

PowerContext captures detailed Sources, but long-running agents also need durable, scoped knowledge that can be evolved, searched, and reused across sessions. Topic Memory turns accepted Sources into long-lived Topic Artifacts while preserving evidence lineage, Scope isolation, publication consistency, and bounded model/resource use.

This change implements that lifecycle end to end rather than exposing an isolated storage or UI layer: Source capture raises durable processing intent, background workers generate and atomically publish Topic revisions, retrieval makes the active projection available to agents, and operators can inspect the same published state from the Dashboard.

- Add the `topic-memory` Artifact family, Markdown chunking, immutable revisions, evidence lineage, current-head projections, and SQLite/OceanBase lexical plus optional vector/hybrid indexes.
- Persist processing Pending state, Cursors, frozen automatic waves, Leases, fencing generations, retries, and per-binding scheduling. Source capture and Pending updates share one transaction; publication validates Lease, Cursor generation, and Topic head compare-and-swap guards in one short transaction.
- Add the bounded generation pipeline for Probe, planning, CREATE/UPDATE/NOOP evolution, related-topic coordination, and conflict-safe retries. Model calls run outside database transactions, provider input/output is budgeted, and server-owned identities and evidence are never delegated to model output.
- Expose scoped flush, search, and exact-get operations through OpenAPI, generated clients, HTTP, MCP capabilities, and the Codex project-context skill. Prepared context can include retrieved Topic Memory without weakening existing Scope or authorization boundaries.
- Add the read-only `/topics` Dashboard with Scope selection, search, sorting, pagination, list/detail views, source and publication explanations, localized navigation, responsive layouts, and stale-response protection. The page does not create, edit, or delete Topic Memory.
- Add configuration, readiness, deployment-role, troubleshooting, and bilingual RFC/documentation updates. Optional retrieval channels activate only when their complete configuration and indexes are ready.
- Add repository, runtime, API, browser, and product-chain coverage for persistence semantics, generation, atomic publication, retrieval bounds, credential isolation, and a real local Codex-to-Dashboard workflow.

- Agents can call `search_topic_memory` and `get_topic_memory` to reuse current Topic knowledge for the active Scope. Operators can request a flush through `POST /v1/topic-memory/flush`.
- The Dashboard now includes a read-only Topics page. Results remain Scope-isolated and expose current publication metadata and evidence without providing mutation controls.
- Topic processing can run with the API process or the supported background deployment role. Invalid model budgets, incomplete vector configuration, or unsupported scheduler combinations fail readiness instead of silently degrading processing.
- Database initialization adds the Topic Memory processing, publication, projection, and index structures. Existing Artifact families and unfiltered prepared-context behavior remain supported.

# How was this change tested?

Local validation covered the complete feature branch at `4189638ba6851b2686c0bb94ce5138fb5e70663e`. These are local results, not a claim that remote CI has passed.

- `make test`: **1263 passed, 11 skipped**.
- `make check`: passed, including manifest, lock, formatting, lint, and type checks.
- `make contract-test`: **30 passed**.
- `make build`: passed.
- Focused Topic Memory product-chain suite: **15 passed**.
- Real local acceptance passed from Codex Source capture through background generation, search/exact retrieval, prepared context, and the `/topics` Dashboard.
- Python 3.11 and 3.13 credential-isolation acceptance passed; browser requests were verified not to receive the generation or embedding provider credentials.
- Cumulative `git diff --check` passed.

Known validation limitations:

- Remote CI is pending.
- Live OceanBase/SeekDB and a real external embedding provider were not available for the final acceptance run; their dialect, configuration, fail-closed, and adapter contracts are covered by local tests.

# AI usage statement

OpenAI Codex assisted with implementation, review-driven fixes, regression tests, documentation, and local product acceptance. The results above come from executed tests and real local PowerContext/Codex/browser processes; AI assistance is not a substitute for maintainer review or CI.


src/powercontext/builtin/persistence/topic_memory.py
```diff
@@ -0,0 +1,648 @@
...
+                        TOPIC_MEMORY_RETRIEVAL_SHAPE_TABLE.c.shape,
+                        TOPIC_MEMORY_RETRIEVAL_SHAPE_TABLE.c.profile_fingerprint,
+                    ).where(TOPIC_MEMORY_RETRIEVAL_SHAPE_TABLE.c.singleton == 1)
+                )
+            ).one_or_none()
+            if stored is None:
+                raise TopicMemoryStorageInvariantError("missing-retrieval-shape", 1)
+        stored_shape = str(stored[0])
+        stored_fingerprint = None if stored[1] is None else str(stored[1])
+        if (stored_shape, stored_fingerprint) != (shape, fingerprint):
```
Teingi, line 593: [P2] Allow embedding configuration changes before any Topic exists

This rejects an embedding configuration change even when the database contains no Artifacts. I reproduced it by opening an empty file-backed SQLite database without an embedding model, closing it, and reopening it with an embedding model: the second startup fails with `retrieval-shape (database requires fts; runtime configured hybrid:...)`. The same sequence succeeds on the PR base. Because Topic initialization runs for every built-in runtime, this also prevents users who have never used Topic Memory from enabling embeddings for ordinary Memory. Please allow the shape to change while no Topic data exists, and cover this empty-database transition with a regression test.

src/powercontext/builtin/runtime/artifact_processing.py
```diff
@@ -0,0 +1,1451 @@
...
+                await self._lose_leadership(ArtifactProcessingSupervisorStatus.STANDBY)
+                return
+            if completion.outcome in {
+                ArtifactProcessingWorkerOutcome.CURSOR_CONFLICT,
+                ArtifactProcessingWorkerOutcome.HEAD_CONFLICT,
+            }:
+                self._defer_control_conflict(running.work)
+                continue
+            self._retries.pop(key, None)
+            assignment = await self._prepare_assignment(running.work)
```
Teingi: [P2] Contain successor-window selection errors to the affected Scope

Unlike `_dispatch_workers`, this call does not catch `_WindowSelectionError`. If a Scope completes its first Window but selecting the next Window fails, the exception reaches `_run`, which drops leadership and cancels every other running Worker. I reproduced Scope A advancing its Cursor to 1, then failing selection of its second Window; an unrelated in-flight Scope B was terminated and the Supervisor became `degraded`. Please apply the same per-key failure/backoff handling used during initial dispatch, or requeue the work and let that path select the next Window. A regression test should fail selection after one successful Window while another Scope is running.

src/powercontext/builtin/runtime/composition.py
```diff
@@ -373,6 +442,89 @@ async def open_builtin_runtime(
...
+        contexts.token_estimator,
+        context_window_tokens=config.inference.generation_model_context_window_tokens,
+    )
+    automatic = config.runtime.topic_memory_schedule_seconds
+    return (
+        *configured,
+        ArtifactProcessingBinding(
+            binding_name=TOPIC_MEMORY_SOURCE_WINDOW_BINDING,
+            source_window_limit=config.runtime.topic_memory_source_window_limit,
+            launcher=SpawnArtifactProcessingWorkerLauncher(entrypoint),
```
Teingi, line 518: [P2] Handle in-memory SQLite before selecting the spawn launcher

`SQLiteConfig()` defaults to `sqlite+aiosqlite:///:memory:`, but this binding always reconstructs the database in a fresh child process. That child gets an empty database instead of the parent's Source Journal and Lease. I reproduced this with the actual spawn launcher: a Source captured in the parent consistently fails in the Worker with `source_window_changed`, so processing cannot advance and retries cannot recover it. Please provide an execution path that can access the same in-memory database, or reject this configuration before advertising Topic processing and accepting flush requests. Cover the default SQLite configuration as well as the file-backed case.

src/powercontext/builtin/persistence/sqlite/topic_memory_index.py
```diff
@@ -0,0 +1,673 @@
...
+    async def initialize(self, connection: AsyncConnection, /) -> None:
+        if connection.dialect.name != "sqlite":
+            raise TopicMemoryCapabilityError("sqlite-fts")
+        await connection.exec_driver_sql(_CREATE_TOPIC_FTS_SQL)
+        await connection.exec_driver_sql(_CREATE_CHUNK_FTS_SQL)
+        marker = await connection.scalar(select(SQLITE_TOPIC_MEMORY_FTS_MARKER_TABLE.c.singleton))
+        if marker is None:
+            await connection.execute(insert(SQLITE_TOPIC_MEMORY_FTS_MARKER_TABLE).values(singleton=1, schema_version=1))
+        await connection.exec_driver_sql("DELETE FROM pc_topic_memory_topic_fts")
+        await connection.exec_driver_sql("DELETE FROM pc_topic_memory_chunk_fts")
```
Zxf-xufeng, line 257: [P2] Avoid rebuilding all SQLite Topic FTS indexes on every Worker startup

Every spawned Topic Worker calls `open_builtin_contexts()`, which reaches this initializer and unconditionally deletes and rebuilds both FTS indexes across all Scopes. I reproduced this by opening a Worker for Scope A and observing it delete the entire indexes and reinsert Scope B's rows before processing any Source. Each bounded Source Window therefore incurs database-wide indexing work and holds SQLite's write lock while rebuilding, creating contention with Source capture and other Workers as the database grows. Please separate index creation/recovery from normal Worker initialization so Workers reuse existing indexes.

src/powercontext/builtin/runtime/topic_memory_processing.py
```diff
@@ -0,0 +1,1298 @@
...
+        evidence: tuple[TopicMemoryEvidence, ...],
+        stage: str,
+    ) -> Iterable[tuple[TopicMemoryEvidence, ...]]:
+        batch: list[TopicMemoryEvidence] = []
+        for item in evidence:
+            candidate = (*batch, item)
+            if self._evidence_batch_fits(candidate, stage):
+                batch.append(item)
+                continue
+            if not batch:
```
Zxf-xufeng: [P2] Provide a recovery path for a single Source exceeding the input budget

The new token budget and window shrinking handle oversized multi-Source Windows, but a single oversized Source still permanently blocks subsequent Sources in the same Scope under unchanged configuration. The selector retains that Source as a one-item Window, this check raises `input_budget_exceeded`, and the Supervisor retries indefinitely without advancing the Cursor. The existing oversized-Source test only reaches the normal tail after increasing the input budget from 100,000 to 300,000; it does not demonstrate recovery with the original configuration. Please provide either evidence-preserving splitting or an explicit blocked state with an actionable operator recovery path, rather than indefinitely retrying a deterministic budget failure. A regression test should cover an oversized Source followed by a normal Source without increasing the configured budget or silently skipping evidence.

src/powercontext/builtin/persistence/topic_memory.py
Teingi: Fixed in 586c20dc. The retrieval shape can change while no Topic data exists, and publication rechecks the shape so a stale runtime cannot write incompatible data after the first Topic. The regression reopens the same SQLite database through FTS, hybrid, a different embedding profile, and FTS again; databases with Topic revisions still reject incompatible changes.

src/powercontext/builtin/runtime/artifact_processing.py
Teingi: Fixed in 586c20dc. Successor-window selection now uses the same per-Scope failure and backoff handling as initial dispatch. The regression completes Scope A's first window, fails selection of its next window while Scope B is running, and verifies that the lease, Scope B's worker, and Scope A's durable pending work are preserved.

src/powercontext/builtin/runtime/composition.py
Teingi: Fixed in 586c20dc. Runtime composition rejects configured Topic processing with in-memory SQLite before exposing processing capability or accepting flush requests, with an actionable error requiring file-backed SQLite. The default configuration rejection and file-backed binding are covered, alongside an actual spawned-worker test. The bilingual configuration reference documents the requirement.

src/powercontext/builtin/persistence/sqlite/topic_memory_index.py
Teingi: Fixed in 586c20dc. Normal initialization reuses both existing FTS indexes when the schema marker is present. Rebuilding is limited to initial creation or recovery of a missing index/marker. Reopen tests verify zero index writes, preserved cross-Scope search results, and recovery when either FTS table or the marker is missing.

src/powercontext/builtin/runtime/topic_memory_processing.py
Teingi: Fixed in 586c20dc. Oversized Sources are processed as contiguous, budget-checked fragments in Probe and temporary-Topic requests, retaining every character and the original evidence ID/SourceRef. Cursor advancement and publication remain atomic after all fragments succeed. Regressions process the oversized Source and its normal successor at the original budget, verify complete text coverage and lineage, and confirm that a failed final fragment publishes nothing and leaves the Cursor unchanged. Both RFC locales now describe this behavior.

Teingi (approved):
LGTM

PR #1492, opened 2026-09-07
fix(prompts): reject generic artifact publication

### LINKED ISSUE #1491: bug: Prompt publication creates unusable artifacts
### Describe the bug

`POST /v1/artifact-publications` accepts Prompt Artifacts and returns `201`, but assigns the target a generated `pub_*` ID. Runtime resolves Prompts by registered operation keys such as `memory.extract`, so the published copy is never selected. Updating the copy fails with `422 / unknown_prompt_key`.

The publication path only excludes Memory and calls `ArtifactRepository.copy_exact` directly. This bypasses `PromptManagementWriter`, which validates the Prompt's logical ID and content. The identity check involved here is the registered Prompt key, not caller authentication.

### Steps to reproduce

1. Start the standard Server with SQLite and `generation_model="test"`, and create two Scopes. Use their IDs as `SOURCE_SCOPE` and `TARGET_SCOPE` below. With access control enforced, use a principal authorized to administer both Scopes.
2. Create a custom Prompt with `POST /v1/scopes/SOURCE_SCOPE/artifacts`:

   ```json
   {
     "family": "prompt",
     "prompt_key": "memory.extract",
     "content": {
       "schema_version": "powercontext.prompt.v1",
       "mode": "custom",
       "instructions": "Keep stable preferences.",
       "demonstrations": []
     }
   }
   ```

3. Publish revision 1 with `POST /v1/artifact-publications`:

   ```json
   {
     "source": {
       "scope_id": "SOURCE_SCOPE",
       "artifact": {
         "family": "prompt",
         "artifact_id": "memory.extract",
         "revision": 1
       }
     },
     "target_scope_id": "TARGET_SCOPE",
     "idempotency_key": "publish-prompt"
   }
   ```

4. Read `GET /v1/scopes/TARGET_SCOPE/prompts/memory.extract`.
5. Use the returned publication's target ID in `PUT /v1/scopes/TARGET_SCOPE/artifacts/prompt/pub_*`, with `If-Match: "revision:1"` and the original Prompt content under the `content` field.

### Expected behavior

Reject Prompt publication with `422 / artifact_publication_unsupported` before creating any target Artifact. Prompt configuration in the target Scope should use the regular Artifact Create/Replace APIs and their dedicated Prompt writer.

### Actual behavior

- Publication returns `201` and creates a Prompt whose ID starts with `pub_`.
- The target Scope's `memory.extract` configuration still reports `mode: auto` and `artifact: null`.
- Updating the published copy returns:

  ```json
  {"error":{"code":"unknown_prompt_key","message":"Prompt key is not registered","details":null}}
  ```

This reports a successful publication while leaving an unusable, unupdatable Prompt in the target Scope. Existing custom configuration under a registered key would also remain unaffected by the copy.

### Environment

- PowerContext: `master` at `e4ebdcdff64a9793aa30f5d087cc71cd7e9ba87c`
- Python: 3.11.13; Linux
- SQLite with the standard Server and in-process HTTP transport; no external model calls
- The incorrect `201` response is reproduced with access control both disabled and enforced.

Teingi:

Closes #1491.

# Rationale for this change

Publishing a Prompt through `/v1/artifact-publications` returns `201` with a generated `pub_*` ID. Runtime resolves Prompts by registered keys, so the target Scope never selects the copy, and updating it fails with `unknown_prompt_key`. Generic publication bypasses the dedicated Prompt writer's identity and content validation.

- Reject Prompt publication in the shared publication application before copying any target state, using the existing `artifact_publication_unsupported` error.
- Add HTTP regression coverage with access control disabled and enforced: repeated publication attempts leave no copy, registered-key creation and updates remain usable, and rejected publication preserves an existing target configuration.
- Document the rejection and supported configuration path in OpenAPI and both API reference locales; regenerate the checked-in schema.

Prompt publication now returns `422 / artifact_publication_unsupported` with `details.family=prompt`. Configure the target Scope through Artifact Create/Replace using a registered `prompt_key`. Previously published `pub_*` Prompt copies remain unusable and require separate cleanup.

# How was this change tested?

- Reproduced the original `201` response, target Auto configuration, and `422 / unknown_prompt_key` update failure on SQLite. Both new regression cases failed before the fix.
- `uv run --locked --group dev python -m pytest tests/e2e/test_prompt_management_api.py tests/builtin/test_publication.py tests/builtin/artifacts/prompt tests/builtin/persistence/test_prompt_records.py tests/e2e/test_access_control_regressions.py tests/test_access_http.py -q --tb=short` — 93 passed.
- `make api-generate` and `make contract-test` — generated files are current; 43 passed.
- `uv run --locked --group dev prek run -a` — all hooks passed, including Ruff and repository-wide type checking.
- `CI=true UV_PYTHON=/usr/bin/python3.11 make docs-test` with Node 22.22.3 and pnpm 11.13.1 — lint, build, and export verification passed for 650 public pages.
- `git diff --cached --check` — passed.

# AI usage statement

Codex (GPT-6) was used to investigate and reproduce the defect, implement the fix and regression tests, update documentation, and run validation.


Zxf-xufeng (approved):
LGTM

PR #1496, opened 2026-09-08
fix(auth): enforce revoked subject scope contribution

### LINKED ISSUE #1495: fix(auth): subject source dual writes bypass revoked target contribution
## Problem

After #1487 was merged, the subject-source endpoint can write to an existing target Scope after the static principal’s scope.contributor grant has been revoked. Ordinary Source creation correctly returns 403, but subject-source creation returns 201 and persists both copies.

Related review: https://github.com/oceanbase/powercontext/pull/1487#discussion_r3954128708

## Reproduction

1. Enable enforced access control with static bearer authentication.
2. Create a subject binding through POST /v1/scopes/{scope_id}/subject-sources.
3. Revoke the target Scope contribution grant through POST /v1/access/bindings/revoke.
4. A direct Source POST to the target returns 403.
5. Repeat the subject-source POST: it incorrectly returns 201 and writes both Sources.

Reproduced on master at c9ab81199f8a3c0ab208bd1b28e3f9f68cc2eba1, with both automatically created and explicitly supplied subject Scopes.

## Expected behavior

Subject dual writes must check the effective contribution permission on the resolved target Scope. A denied write must persist neither Source and must not advance either Source journal. First use of a fresh explicit Scope must continue to succeed.

## Cause and proposed fix

Static preset bootstrap is idempotent and does not reactivate a revoked binding. The subject-source endpoint currently skips the target contribution check after bootstrap. Recheck target contribution after bootstrap, binding builtin authorization reads to the same transaction so newly created grants are visible. Preserve external authorization providers and add regression tests.

No API or database migration is needed.

Zxf-xufeng:

Closes #1495.
Follow-up fix for merged PR #1487 and https://github.com/oceanbase/powercontext/pull/1487#discussion_r3954128708.

## Rationale for this change

Static preset initialization is idempotent: replaying a revoked grant does not reactivate it. The subject-source endpoint nevertheless skipped its target contribution check after initialization, allowing a dual write where an ordinary Source write correctly returned 403.

- Require scope.contribute on the resolved existing subject Scope after static preset initialization.
- Bind builtin authorization reads, relationship writes, and audit writes to the same transaction so the permission check sees newly created grants on first use.
- Preserve externally supplied authorization providers.
- Add HTTP regression coverage for automatically created and explicitly supplied subject Scopes: after revocation, direct and dual writes return 403; neither Source storage nor either journal advances.
- Retain existing first-use explicit Scope coverage.

Subject-source writes now correctly return 403 when target contribution has been revoked. First-use initialization remains supported. No API shape, schema, migration, RFC, or dependency changes.

- Reproduced the pre-fix dual-write response of 201 after revocation.
- uv run pytest tests/e2e/test_profile_api.py tests/e2e/test_access_control_regressions.py tests/e2e/test_access_control_http.py tests/test_access_control.py -q — 50 passed.
- uv run prek run -a — all hooks passed, including Ruff and ty.
- git diff --check — passed.

## AI usage statement

Implemented and tested with OpenAI Codex. AI assistance was used for diagnosis, code changes, regression tests, and this PR description.

### DIFF
--- a/src/powercontext/server/app.py
+++ b/src/powercontext/server/app.py
@@ -2031,7 +2031,10 @@ async def create_subject_source(
         elif access.uses_static_preset(principal):
             # The subject Scope is resolved inside the atomic dual-write transaction,
             # so materialize the fixed static preset through the same connection.
-            await access.with_connection(connection).bootstrap_static_scope(principal, target, context=context)
+            bound = access.with_connection(connection)
+            await bound.bootstrap_static_scope(principal, target, context=context)
+            # Bootstrap is idempotent and must not bypass a previously revoked grant.
+            await bound.require(principal, AccessAction.SCOPE_CONTRIBUTE, ResourceRef.scope(target), context=context)
         else:
             await access.require(principal, AccessAction.SCOPE_CONTRIBUTE, ResourceRef.scope(target), context=context)
 
--- a/src/powercontext/server/authz/service.py
+++ b/src/powercontext/server/authz/service.py
@@ -340,6 +340,10 @@ class BuiltinAuthorizationProvider:
         self._deployment_id = deployment_id
         self._clock = clock or (lambda: datetime.now(UTC))
 
+    def with_repository(self, repository: AccessRepository) -> BuiltinAuthorizationProvider:
+        """Use transaction-local relationships without changing policy semantics."""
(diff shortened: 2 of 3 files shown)


Teingi (approved):
LGTM

PR #1502, opened 2026-09-08
docs(rfc): add scope and source discovery

Zxf-xufeng:

Proposes a new RFC; no tracking issue.

# Rationale for this change

Clients need to discover a Scope from a partial `scope_id` and then enumerate the public Sources owned by that Scope. The existing API only supports an unfiltered Scope list and exact Source reads.

- Extend `GET /v1/scopes` with optional literal `scope_id` substring filtering.
- Define paginated `GET /v1/scopes/{scope_id}/sources` using the existing public `SourceRecord` model.
- Specify authorization, public Source type boundaries, snapshot pagination, cursor binding, persistence reuse, errors, and client compatibility.
- Add synchronized English and Chinese RFC documents.

This is a design-only PR. If implemented, it adds one optional query parameter and one new read-only endpoint without changing existing no-argument behavior or persisted schemas.

# How was this change tested?

- `uv run prek run --files docs/en/rfcs/0000_scope_source_discovery.md docs/zh/rfcs/0000_scope_source_discovery.md`

# AI usage statement

OpenAI Codex was used to inspect the existing API and persistence model and to draft the RFC. The author reviewed the resulting design.


docs/en/rfcs/1502_scope_source_discovery.md
```diff
@@ -0,0 +1,221 @@
...
+must encode the complete `scope_id` as one path segment.
+
+# Reference-level explanation
+
+## Scope query contract
+
+`GET /v1/scopes` accepts one optional `query` string of at most 256 Unicode characters. The server trims leading and
+trailing whitespace and then:
+
+- matches only `scope_id`, never title, summary, references, Source content, or Artifact content;
```
Teingi: **[P2] Align Scope discovery with the current opaque IDs**

`ScopeApplication` now generates opaque `scp_...` IDs; repository names and URLs live in metadata or bindings. I verified this with the current SQLite application: creating a Scope titled `PowerContext` with a repository external reference produces an opaque ID, and applying the proposed `powercontext` substring filter to its ID yields no match. The repository discovery example above therefore does not work for newly created Scopes. Please either define matching against the relevant metadata/binding fields, or explicitly limit the motivation and examples to discovery by a known fragment of a generated Scope ID.

Zxf-xufeng: Thanks for catching this. I’ve updated the RFC to reflect the current opaque scp_... IDs. Scope discovery now requires an explicit query_field and supports searching title, summary, External Reference value, Binding external_id, or scope_id. The examples now search the PowerContext title or repository External Reference instead of assuming repository-derived Scope IDs. The corresponding Chinese RFC has been updated as well. Addressed in 8075190.

Teingi (approved):
LGTM

PR #1503, opened 2026-09-08
feat(api): add scope and source discovery

Zxf-xufeng:

Implements RFC #1502.

# Rationale for this change

Clients need to find Scopes from partial IDs and enumerate the public Sources owned by a selected Scope without downloading and reimplementing persistence logic themselves.

- Add optional case-sensitive literal `scope_id` filtering to `GET /v1/scopes`.
- Add snapshot-bounded, cursor-paginated `GET /v1/scopes/{scope_id}/sources` for public Content Sources.
- Reuse the Scope repository, Source repository and adapters, `SourceRecord` mapping, signed cursor codec, and existing authorization resolvers.
- Preserve Handoff Receipt identity attestations in Source list responses.
- Add typed Python client methods and regenerate Python/TypeScript API bindings.
- Add contract, persistence, HTTP, SDK, pagination, snapshot, and cursor-isolation tests.

No database migration is required. The endpoint is read-only and does not invoke generation or advance Source consumer cursors.

Yes. `GET /v1/scopes` accepts optional `query`; existing no-argument behavior is unchanged. A new `GET /v1/scopes/{scope_id}/sources` endpoint and `PowerContextClient.list_sources()` method are available. Only the existing public `content` Source type is listed.

# How was this change tested?

- `make check`
- `make contract-test` — 43 passed
- Focused behavior suite — 82 passed
- Full portable suite with host PowerContext environment overrides removed and Linux-only systemd tests excluded — 1543 passed, 26 skipped

The unmodified `make test` command on this macOS host is affected by pre-existing `POWERCONTEXT_SERVER_*` environment variables and three Linux-only systemd tests; the isolated command confirms all other tests pass.

# AI usage statement

OpenAI Codex was used to inspect the existing architecture, implement the RFC, and run validation. The author reviewed the resulting API and code changes.


src/powercontext/server/app.py
```diff
@@ -1993,6 +2021,44 @@ async def get_handoff_report(
...
+    )
+    items = [_source_record_response(item) for item in result.items]
+    if http_request.app.state.access_mode == "enforced":
+        access = _require_access_control(http_request)
+        for item, source in zip(items, result.items, strict=True):
+            if not _is_handoff_receipt_content(source.content):
+                continue
+            identity = await access.receipt_identity(scope_id, source.source_id)
+            if identity is None:
+                raise AccessUnavailableError("receipt_identity_pending")
```
Teingi: [P2] Avoid blocking the Source collection on user-supplied receipt markers

In enforced mode, a `scope.contributor` can POST `{"content":{"schema":"powercontext.handoff-receipt.v1"}}` to this collection and receive 201. That ordinary Source has no receipt identity, so this branch makes subsequent list requests return `503 receipt_identity_pending` for viewers and administrators. With `limit=1`, traversal stops at that Source without a continuation cursor, although exact reads of later normal Sources still return 200. Please distinguish receipts using server-owned provenance rather than just `content.schema`, or reject the reserved marker on generic ingestion paths and handle already-stored values. Keep the fail-closed identity check for actual receipts.

src/powercontext/builtin/persistence/records.py
```diff
@@ -197,6 +199,62 @@ async def get_source(self, scope_id: str, source_type: str, source_id: str, /) -
...
+                through, after = high_watermark, 0
+            stored = await self._sources.list(
+                connection,
+                scope_id,
+                after=after,
+                through=through,
+                limit=limit + 1,
+                source_type=CONTENT_SOURCE_NAME,
+            )
+            available = tuple(_source_record(scope_id, value) for value in stored)
```
Teingi: [P2] Apply the byte budget before loading the full candidate page

`_sources.list()` fetches and decodes all `limit + 1` payloads, and `_source_record()` hashes each before the 4 MiB budget is checked. With 51 valid ~1 MiB Sources in SQLite, the default `limit=50` returned only ~3 MiB but raised process peak RSS from ~99 to ~311 MiB; `limit=3` returned the same three items with a ~123 MiB peak. Larger payloads or concurrent list requests amplify this memory cost despite the small response. Please read incrementally and stop loading/decoding content when the byte budget is reached, retaining only the lookahead needed to issue `next_cursor`.

src/powercontext/server/app.py
Zxf-xufeng: Agreed. content.schema should not be treated as trusted receipt provenance.
I’ll use the persisted, server-owned receipt identity as the authoritative classification, reject the reserved powercontext.handoff-receipt.v1 marker on the generic Source creation endpoint, and treat already-stored marker-only Sources without server provenance as ordinary Sources so they cannot block collection traversal.
Actual handoff receipts will continue to record identity before creating the Source and fail closed when the identity store is unavailable. I’ll add regressions covering forged markers, legacy stored values, pagination past them, and genuine receipt identity enrichment.

src/powercontext/builtin/persistence/records.py
Zxf-xufeng: Agreed. The current implementation applies the response budget after loading and decoding the entire limit + 1 candidate set, so the response-size limit does not provide a corresponding memory bound.
I’ll change Source traversal to read rows incrementally and stop decoding as soon as either the item limit or 4 MiB response budget is reached. Cursor continuation will be based on the last returned journal position, with only bounded lookahead used to determine whether another page exists.
I’ll add regression coverage using multiple large Sources to verify the page boundary and that subsequent pages remain reachable.

src/powercontext/server/app.py
Zxf-xufeng: Fixed in 9cf7529b. Generic Source creation and capture now reject the reserved handoff receipt schema. Existing marker-only Sources without server-owned identity are treated as ordinary Sources and no longer block exact reads or pagination. Actual receipts still fail closed when the identity store is unavailable. Regression coverage was added for all of these cases.

src/powercontext/builtin/persistence/records.py
Zxf-xufeng: Fixed in 9cf7529b. Source collection traversal now streams rows incrementally and stops decoding when the item limit or 4 MiB response budget is reached. It retains only one lookahead and continues from the last returned journal position. The regression test verifies bounded decoding and subsequent-page reachability.

openapi/powercontext.yaml
```diff
@@ -281,8 +343,12 @@ paths:
           $ref: "#/components/responses/Unauthorized"
         "403":
           $ref: "#/components/responses/Forbidden"
+        "422":
```
Teingi, line 353: [P3] Declare the Scope pagination cursor errors

Scope pagination returns `400 invalid_cursor` for invalid or mismatched cursors and `410 cursor_expired` for expired ones; I verified both through the HTTP endpoint. Neither response is declared here or in the served OpenAPI schema. Please add the existing `BadRequest` and `CursorExpired` responses, as `list_sources` does, and regenerate the bindings so the published contract includes these pagination failure cases.

Zxf-xufeng: Fixed in c33b3944. GET /v1/scopes now declares the existing 400 BadRequest and 410 CursorExpired responses. The generated bindings were refreshed, and a contract regression test was added to prevent these cursor errors from being omitted again. make contract-test and make check both pass.

Teingi (approved):
LGTM

PR #1506, opened 2026-09-08
fix(cli): make default deployment model-free

### LINKED ISSUE #1504: bug: default deployment leaves generation and embedding capabilities disabled
## Describe the bug

A fresh PowerContext deployment with SeekDB, authentication, and Dashboard configured does not configure a generation model or an embedding model. As a result, model-backed capabilities such as Memory extraction, Experience generation, managed Skill generation, and generated Handoff are disabled by default.

The Server starts and the Dashboard is reachable, so the missing inference configuration is easy to overlook.

## Steps to reproduce

1. Deploy the latest `master` version of PowerContext.
2. Configure SeekDB storage, Server authentication, and Dashboard only.
3. Start the Server and run `powercontext capabilities`.
4. Try to generate a managed Skill or Experience from existing Source evidence.

## Expected behavior

The deployment flow should either:

- prompt for generation and embedding provider/model configuration, or
- show a prominent, actionable warning that Skill, Experience, Memory extraction, vector search, and generated Handoff are unavailable until these models are configured.

It would also help if the Dashboard surfaced the missing-model state next to the affected capabilities.

## Actual behavior

The Server is healthy, but the capabilities remain disabled because no inference model is configured. Generation requests cannot produce Candidates until a generation model is added; vector/hybrid search is unavailable until an embedding model plus profile ID and dimension are configured.

## Environment

- PowerContext: `0.2.1.dev33+g80d618bd6` (latest master at the time of testing)
- Python: 3.14
- OS: macOS
- Storage: embedded SeekDB (`pylibseekdb 1.4.0`)
- Service: per-user launchd

No API keys, access tokens, or local database paths are included here.

## Additional context

The current configuration documentation describes the model variables, but the default deployment path can appear complete while model-backed artifact generation is silently unavailable. A first-run diagnostic or explicit capability banner would make this easier to discover.

Zxf-xufeng:
Closes #1504

# Rationale for this change

The default deployment flow prompted for generation and embedding models even though PowerContext can start and provide basic Memory/Source capabilities without an inference provider. This made the default deployment appear incomplete and prevented users from discovering which capabilities require model configuration.

- Make `powercontext config init` create a runnable model-free deployment without asking for provider, model, credential, embedding profile, or schedule settings.
- Print a post-deployment inference capability notice describing degraded or unavailable generation, extraction, reranking, artifact-generation, vector/hybrid retrieval, and FTS fallback behavior.
- Cover all currently supported coding agents: Codex, Claude Code, DeepSeek Harness, OpenCode, Pi, OpenClaw, Hermes, and WorkBuddy; also explain the shared Server state for Python adapters.
- Accept and validate model-free environment files while preserving explicit full-capability configuration.
- Update English and Chinese deployment/configuration documentation and add CLI regression coverage.

Yes. Default setup no longer prompts for model configuration. Users who need automatic extraction, semantic/vector retrieval, reranking, Experience/Skill/Handoff generation, or Prompt demonstration can add the documented model settings and run `powercontext config validate`.

# How was this change tested?

- `.venv/bin/ruff check src/powercontext/cli/config.py tests/test_config_cli.py`
- `.venv/bin/ty check src/powercontext/cli/config.py`
- `env -u POWERCONTEXT_CLIENT_API_TOKEN -u POWERCONTEXT_SERVER_AUTH_ENABLED -u POWERCONTEXT_SERVER_AUTH_TOKEN -u POWERCONTEXT_SERVER_DASHBOARD_SCOPES .venv/bin/pytest tests/test_config_cli.py tests/test_server.py tests/test_capabilities.py -q` — 85 passed
- Full suite with the same environment cleanup — 1541 passed, 26 skipped, 3 failed; the failures are Linux-only systemd service tests running on macOS.

# AI usage statement

Built with Codex (GPT-5); reviewed the resulting diff and test output locally.


src/powercontext/service/cli.py
```diff
@@ -70,18 +72,32 @@ def install(
         typer.echo("--no-start-on-login is currently supported only on Windows.", err=True)
         raise typer.Exit(code=2)
     try:
+        with server_settings_context(env_file=env_file) as settings:
```
Teingi: [P2] Preserve tilde expansion for the service environment file

This pre-read bypasses the `expanduser()` in `load_protected_environment_file()`. A valid protected file passed as `--env-file "~/.config/powercontext/server.env"` now fails with `No such file or directory` before reaching `ServiceController.install()`. I reproduced this with the same file and a fake service manager: the base exits 0, while this head exits 2 without touching the manager. Please expand the path before reading it, or reuse the existing protected loader, and add a CLI regression test for a literal `~` path.

Zxf-xufeng: Thanks for the review.

The `~` path issue has been fixed in the latest commit . `powercontext service install` now expands `--env-file` with `Path.expanduser()` before loading or installing the service definition. A regression test for a literal `~/...` path was also added.

The unified missing-model notice now uses the agreed wording and fixed documentation URL:

未配置或未完整配置 PowerContext Server 推理模型（generation model 和 embedding model），
可能影响部分制品功能。具体影响范围及配置方式请参考 PowerContext 官网配置说明：
https://powercontext.oceanbase.io/en/docs/reference/configuration/

Related local tests pass: 170 passed, 1 skipped.

There are still two follow-ups before merge:

1. The macOS LaunchAgent CI job is failing because the registered service does not become live. The service stdout/stderr logs are empty, so this needs to be rerun and investigated.
2. `config init --force` currently replaces an existing managed configuration with the model-free configuration, which may remove existing model and provider settings without an additional confirmation.

Teingi (approved):
LGTM

PR #1511, opened 2026-09-08
fix(skill): validate raw archive entry names

### LINKED ISSUE #1498: bug(windows): Skill ZIP validation loses raw entry names
### Describe the bug

Skill archive validation can accept a raw ZIP entry containing a backslash on Windows even though the package validator explicitly rejects backslashes. The same input therefore has different acceptance behavior across platforms.

[`_archive_files()`](https://github.com/oceanbase/powercontext/blob/80d618bd61432d1bb4eae85433fa22f2bbd50bce/src/powercontext/builtin/artifacts/skill/package.py#L276) validates `ZipInfo.filename`. On Windows, Python has already changed backslashes in that property to forward slashes. The original name remains in `ZipInfo.orig_filename`, but the application does not validate it. By the time [`_validate_relative_path()`](https://github.com/oceanbase/powercontext/blob/80d618bd61432d1bb4eae85433fa22f2bbd50bce/src/powercontext/builtin/artifacts/skill/package.py#L449) runs, the evidence needed for its backslash rejection is gone.

### Steps to reproduce

In a Windows checkout of `80d618bd61432d1bb4eae85433fa22f2bbd50bce`, run the following through the project's Python environment:

```python
import io
import zipfile
from powercontext.builtin.artifacts.skill.package import capture_skill_archive

buffer = io.BytesIO()
with zipfile.ZipFile(buffer, "w") as archive:
    archive.writestr("SKILL.md", b"---\nname: safe-skill\ndescription: Safe.\n---\n")
    entry = zipfile.ZipInfo("nested/windows")
    # Assign after construction so the test ZIP really contains a backslash
    # on every platform, including Windows.
    entry.filename = r"nested\windows"
    archive.writestr(entry, b"fixture")

with zipfile.ZipFile(io.BytesIO(buffer.getvalue())) as archive:
    print([(item.orig_filename, item.filename) for item in archive.infolist()])

snapshot = capture_skill_archive(buffer.getvalue())
print([entry.path for entry in snapshot.entries])
```

The existing regression also fails:

```powershell
uv run pytest -q tests/builtin/artifacts/skill/test_package.py::test_archive_rejects_unsafe_paths
```

### Expected behavior

`capture_skill_archive()` rejects the raw backslash-containing entry with `SkillPackageError` consistently across platforms. The archive's original names should be validated before any host-dependent normalization, and canonicalization must not erase invalid input before validation.

Keep traversal, duplicate/collision, forbidden-name, size, and special-file protections intact. The regression fixture must preserve the actual bytes of the invalid name: `writestr("nested\\windows", ...)` by itself normalizes the name while constructing the test on Windows and does not prove the production behavior.

### Actual behavior

The constructed archive contains a real backslash, but reading it on Windows reports:

```text
('nested\\windows', 'nested/windows')
```

The public capture API succeeds and produces paths `['SKILL.md', 'nested/windows']`. The existing rejection test fails with `DID NOT RAISE SkillPackageError`.

The demonstrated consequence is inconsistent validation and silent path reinterpretation. The reproduction does not demonstrate an out-of-directory write or code execution.

### Environment

- Windows x64, CPython 3.12.13; no external service required.
- Related package contract: RFC 1351, Standard Skill Package Lifecycle. This is a current implementation defect in archive input validation.
- No corresponding open issue or implementation PR was found when checking the repository's open items and changed files.
- AI assistance: OpenAI Codex assisted with source inspection, reproduction, and this report.

Praying:
## Summary

Fixes #1498.

`_archive_files()` validated `ZipInfo.filename`, but the host `zipfile` implementation normalizes that property before the check runs:

- on **Windows**, backslashes are rewritten as forward slashes while constructing the `ZipInfo`, so a raw `nested\windows` entry passed validation and was silently reinterpreted as the POSIX path `nested/windows`;
- on **every platform**, the name is truncated at the first NUL byte, so a `nested\x00evil` entry was accepted under its truncated name — which also made the existing `"\x00"` rejection in `_validate_relative_path` unreachable on the archive path (verified on Linux before this fix).

The raw name is only preserved in `ZipInfo.orig_filename`, so archive validation now checks that instead, before any host-dependent normalization can hide hostile input. Traversal, duplicate/collision, forbidden-name, size, and special-file protections are unchanged.

## Regression coverage

- The existing `test_archive_rejects_unsafe_paths` case for `nested\\windows` was removed: `writestr("nested\\windows", ...)` normalizes the name while building the entry on Windows, so the fixture did not prove the production behavior (as noted in the issue).
- The new `test_archive_rejects_hostile_raw_entry_names` assigns the hostile names (`nested\windows`, `nested\x00windows`) after `ZipInfo` construction so the raw bytes reach the archive on every platform, asserts those bytes are present in the built archive, and verifies both names are rejected. The NUL byte case reproduces the silent reinterpretation on Linux, where the CI runs.

## Testing

- `uv run pytest -q tests/builtin/artifacts/skill/test_package.py` — 13 passed
- `uv run pytest -q tests/builtin/` — 578 passed, 4 skipped
- Full suite `uv run python -m pytest --doctest-modules --ignore=tests/e2e` — 1626 passed, 11 skipped, plus one pre-existing failure of `tests/test_topics_dashboard.py::test_topics_browser_state_regressions_execute_in_node`, which fails identically on unmodified `master` in this environment (Node.js dependent).
- `uv run ruff check`, `uv run ruff format --check`, and `uv run ty check` on the changed files all pass.

### DIFF
--- a/src/powercontext/builtin/artifacts/skill/package.py
+++ b/src/powercontext/builtin/artifacts/skill/package.py
@@ -280,7 +280,12 @@ def _archive_files(archive_bytes: bytes) -> tuple[_PackageFile, ...]:  # noqa: C
         entries: list[tuple[zipfile.ZipInfo, str, int]] = []
         total_bytes = 0
         for info in archive.infolist():
-            path = _validate_relative_path(info.filename.rstrip("/") if info.is_dir() else info.filename)
+            # ZipInfo.filename truncates the name at NUL bytes on every platform
+            # and rewrites backslashes as separators on Windows; only
+            # orig_filename keeps the raw archive name, so validate that before
+            # host-dependent normalization can hide hostile input.
+            name = info.orig_filename
+            path = _validate_relative_path(name.rstrip("/") if info.is_dir() else name)
             collision_key = _path_collision_key(path)
             if collision_key in seen_paths:
                 raise SkillPackageError(f"Agent Skill archive contains duplicate or colliding paths: {path}")
--- a/tests/builtin/artifacts/skill/test_package.py
+++ b/tests/builtin/artifacts/skill/test_package.py
@@ -98,7 +98,6 @@ def test_different_zip_order_converges_on_the_same_canonical_package(tmp_path: P
     [
         ("../outside", b"bad"),
         (".env", b"SECRET=value"),
-        ("nested\\windows", b"bad"),
     ],
 )
 def test_archive_rejects_unsafe_paths(name: str, content: bytes) -> None:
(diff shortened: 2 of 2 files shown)


Teingi (approved):
LGTM

PR #1512, opened 2026-09-08
fix(topic-memory): guard runtime profiles and reuse worker indexes

frf12:

Follow-up to #1490 and RFC #1417. No additional issue is closed.

# Rationale for this change

Empty Topic stores can change retrieval configuration, but already-running runtimes and spawned workers must not silently reuse or overwrite a different persisted shape/profile. Topic workers also do not need to recreate ordinary Memory/Experience indexes for every Source Window.

- Check the persisted retrieval shape/profile before and after Topic search, exact get, and current-head browse. Serialize empty-store reconfiguration with publication.
- Filter SQLite and OceanBase vector candidates by embedding-profile fingerprint before applying candidate budgets.
- Make worker bootstrap validate the existing Topic shape instead of configuring it. A stale worker specification fails without modifying the shape or Cursor; normal runtime startup retains empty-store reconfiguration.
- Skip ordinary Memory/Experience index initialization in Topic workers, while preserving normal runtime recovery and Topic validation/publication guards.
- Include the reviewed bilingual RFC #1417 filename/ID migration and reference corrections. The RFC bodies and headings are otherwise unchanged; the example templates remain.

Stale runtimes fail with a typed retrieval-shape error instead of returning results from the wrong embedding space or overwriting current configuration. Topic worker startup avoids unrelated index rebuild work.

No public API or database migration is added. The Source fragmentation behavior merged in #1490 is unchanged. File-backed SQLite remains required for spawned Topic processing. The old `0000_topic_memory.md` documentation paths are replaced by `1417_topic_memory.md`, without redirects.

# How was this change tested?

Validation on head `be68e31bd73e9a7afcd48223d2d28b683fe51de0`:

- Focused persistence/runtime/processing suite: **139 passed, 1 skipped**.
- `make test`: **1790 passed, 26 skipped, 6 warnings**.
- `make check`: passed, including manifest tests, lock consistency, formatting, lint and type checks.
- Cumulative `git diff --check` from the #1490 merge: passed.
- Real spawn regression covers an old FTS worker specification after switching the database to hybrid: bootstrap fails, persisted configuration and Cursor remain unchanged, and the new runtime can publish and search/get/browse.
- SQLite tests cover read-only shape checks and profile filtering before candidate limits; OceanBase SQL-contract tests cover both vector channels.

On parent `93495aecd781f6fc36b70b0a9eac4b64c9642272`, documentation validation passed (`make docs-test`, 663 exported pages, 47 anchors per RFC locale). Independent reliability validation on that parent also exercised OceanBase CE 4.3.5.6 vector/hybrid/detail-vector and stale-profile filtering. The final worker-bootstrap amendment has not been rerun against live OceanBase/SeekDB or real embedding providers.

# AI usage statement

Implemented with OpenAI Codex and AI-assisted review. The changes were checked with automated regression tests, full local tests and quality gates; limitations are stated above.

### DIFF
--- a/src/powercontext/builtin/persistence/topic_memory.py
+++ b/src/powercontext/builtin/persistence/topic_memory.py
@@ -104,7 +104,7 @@ class TopicMemoryRepository:
         if missing_publication is not None:
             raise TopicMemoryStorageInvariantError("missing-publication", tuple(missing_publication))
 
-        await self._ensure_retrieval_shape(connection)
+        await self._ensure_retrieval_shape(connection, allow_empty_change=True)
         await self.index.initialize(connection)
 
         orphan_active = (
@@ -257,6 +257,7 @@ class TopicMemoryRepository:
 
         if ref.family != TopicMemory.family:
             raise InvalidRepositoryArgumentError("artifact_ref", "must reference topic-memory")
+        await self._check_retrieval_shape(connection)
         topic = await self.artifacts.get(connection, scope_id, ref)
         if not isinstance(topic, TopicMemory):
             raise TopicMemoryStorageInvariantError("artifact-type", ref)
@@ -280,6 +281,7 @@ class TopicMemoryRepository:
         if active_revision is None:
             raise TopicMemoryStorageInvariantError("missing-active-head", (scope_id, ref.artifact_id))
         current = ArtifactRef(family=TopicMemory.family, artifact_id=ref.artifact_id, revision=int(active_revision))
+        await self._check_retrieval_shape(connection)
         return PublishedTopicMemory(
             topic=topic,
             published_at=_aware_utc(published_at),
@@ -300,6 +302,7 @@ class TopicMemoryRepository:
 
         if not 1 <= limit <= 100:
             raise InvalidRepositoryArgumentError("limit", "must be between 1 and 100")
+        await self._check_retrieval_shape(connection)
         statement = (
             select(
                 TOPIC_MEMORY_ACTIVE_TOPICS_TABLE.c.artifact_id,
@@ -356,6 +359,7 @@ class TopicMemoryRepository:
                 ).limit(limit)
             )
         ).mappings()
+        await self._check_retrieval_shape(connection)
         return tuple(
             TopicMemoryCurrentItem(
                 artifact_ref=ArtifactRef(
@@ -406,6 +410,7 @@ class TopicMemoryRepository:
                 f"must contain at most {MAX_TOPIC_MEMORY_QUERY_TERMS} distinct Analyzer terms",
             )
         query_vector = self._canonical_query_vector(used_mode, query_vector)
+        await self._check_retrieval_shape(connection)
         if not analyzed and used_mode == "fts":
             return TopicMemorySearchResult(mode=used_mode, hits=())
         request = TopicMemorySearchRequest(
@@ -417,6 +422,7 @@ class TopicMemoryRepository:
             embedding_profile=embedding_profile,
         )
         channels = await self.index.search(connection, scope_id, request)
+        await self._check_retrieval_shape(connection)
         return TopicMemorySearchResult(
             mode=used_mode,
             hits=fuse_topic_memory_rankings(query, channels, limit, mode=used_mode),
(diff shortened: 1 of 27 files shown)


src/powercontext/builtin/persistence/topic_memory.py
```diff
@@ -104,7 +108,10 @@ async def initialize(self, connection: AsyncConnection, /) -> None:
         if missing_publication is not None:
             raise TopicMemoryStorageInvariantError("missing-publication", tuple(missing_publication))
 
-        await self._ensure_retrieval_shape(connection)
+        if configure_retrieval_shape:
+            await self._ensure_retrieval_shape(connection, allow_empty_change=True)
+        else:
+            await self._check_retrieval_shape(connection)
```
Teingi, line 114: [P2] Keep worker integrity checks on a consistent snapshot

The worker path now checks the shape without acquiring the publication lock, but the subsequent integrity scan still reads heads and chunks in separate queries. If another worker commits revision 2 after `active_rows` reads revision 1, the chunk count for revision 1 becomes zero and bootstrap raises `missing-active-chunks` against a healthy database. Since this scan covers all scopes, a publication in another scope can fail an unrelated worker and trigger retry backoff.

I reproduced this on this head with real connections to file-backed SQLite and OceanBase CE 4.3.5.4 (`READ-COMMITTED`, `autocommit=0`). The same SQLite interleaving succeeds on the base commit. On OceanBase, changing only the test worker's connections to `REPEATABLE-READ` makes bootstrap succeed while the concurrent revision still commits.

Please give the integrity scan a consistent read snapshot, or make the related invariant checks atomic, while retaining the shape/publication guards. Add a concurrent-revision regression for both backends. A transaction boundary alone is insufficient under OceanBase's `READ-COMMITTED` isolation.

frf12: Thanks for the deterministic reproducer. Fixed in [5235867fb85b154baf03cfd1ffe58f8f252d0fdc](https://github.com/oceanbase/powercontext/commit/5235867fb85b154baf03cfd1ffe58f8f252d0fdc), now included in PR head [088a8e7014efd60afaf458dabe14d8d09e0e4556](https://github.com/oceanbase/powercontext/commit/088a8e7014efd60afaf458dabe14d8d09e0e4556).

- Head/active-revision/chunk-count validation now uses one SQL statement, rather than carrying a cached revision into a later chunk query.
- Current-vector completeness is also checked within a single statement snapshot, including full Scope/Artifact/Revision identity, profile fingerprints and exact ordinal membership. SQLite additionally checks the physical vector rows.
- Read-only worker shape validation and the existing fenced publication guards remain in place. This does not change global isolation levels, acquire a new publication lock, or suppress genuine corruption errors.

`test_topic_memory_integrity.py` uses real connections and controlled before/after-commit barriers on file-backed SQLite and Docker OceanBase CE 4.3.5.6 (`READ-COMMITTED`, `autocommit=0`). It covers concurrent revision publication, OceanBase hybrid bootstrap, and negative cases for missing chunks/vectors, mismatched profiles and extra/substituted ordinals.

Verification on the exact PR head:

- Combined SQLite/OceanBase integrity, live retrieval and budget-integration matrix: **39 passed, 0 skipped**. The OceanBase image was pinned to digest `sha256:31086a6900c21c479c2bcd942b6a28c53b17a51f4e9b9eb8eafcc596adfcd2e3`; the disposable instance was removed afterward.
- Budget, restart-fault-injection, fragment-reduction and processing regressions: **71 passed**.
- Full local suite: **1,839 passed, 43 skipped, 6 deprecation warnings**; `make check` and `make docs-test` passed. These are local results, not a claim that remote CI has completed.

The successor commit also retains [f0e6e0bbb8767d506af3e05c2efef2a0a348404b](https://github.com/oceanbase/powercontext/commit/f0e6e0bbb8767d506af3e05c2efef2a0a348404b)'s durable work-budget protection. New cross-backend tests verify that concurrent integrity checks, engine reopen and publication rollback preserve those reservations.

Teingi (approved):
LGTM

PR #1513, opened 2026-09-08
fix(server): use square symbol favicon in dashboard

### LINKED ISSUE #1507: bug: dashboard favicon uses a wide wordmark and appears distorted
### Describe the bug

The Dashboard uses the full horizontal PowerContext wordmark as its browser-tab favicon. In the attached screenshot, the icon beside **PowerContext 审核** is compressed into blurred strokes and is difficult to recognize, while the page-header logo renders normally.

The same favicon declaration is shared by Overview, Prompts, Shared with me, Skills, Review, and Handoff Reports, so the underlying asset-selection problem is not limited to the Review page.

### Steps to reproduce

1. Run PowerContext with the Dashboard enabled, for example with `powercontext server run`. The investigation below uses upstream `master` at `80d618bd61432d1bb4eae85433fa22f2bbd50bce`.
2. Open `http://127.0.0.1:<server-port>/reviews` in a browser. The original report uses port `8766`, Chinese UI, and a Review URL with `scope=...&status=approved`.
3. Inspect the favicon beside the tab title. Use vertical tabs to match the attached screenshot.
4. Inspect the page's `<link rel="icon">`, then open the referenced `/static/powercontext-color.png`: it is a wide logo containing both the symbol and the full product name.

The favicon is declared unconditionally in the shared base template; selecting a particular scope or review status is not required to expose this asset choice.

### Expected behavior

Dashboard tabs should display a clear, recognizable PowerContext symbol, with appropriate proportions at small favicon sizes such as 16×16 and 32×32.

### Actual behavior

The browser-tab icon appears squeezed and blurred. The red rectangle in the original screenshot highlights the affected tab; the full-size page-header branding remains legible.

Original screenshot: the PowerContext Review tab favicon appears distorted inside the red rectangle

#### Investigation

- [`base.html:23`](https://github.com/oceanbase/powercontext/blob/80d618bd61432d1bb4eae85433fa22f2bbd50bce/src/powercontext/server/templates/base.html#L23) sets the favicon to `/static/powercontext-color.png` with `type="image/png"`.
- [The referenced PNG](https://github.com/oceanbase/powercontext/blob/80d618bd61432d1bb4eae85433fa22f2bbd50bce/src/powercontext/server/static/powercontext-color.png) is **1696×240 pixels**, approximately **7.07:1**, and includes the entire wordmark. It is also used appropriately as the header's [`brand-wordmark`](https://github.com/oceanbase/powercontext/blob/80d618bd61432d1bb4eae85433fa22f2bbd50bce/src/powercontext/server/templates/components/header.html#L25).
- The favicon reference changed from `oceanbase-logo.svg` to this PNG in [the branding change #1350](https://github.com/oceanbase/powercontext/commit/efb2ff2f702b9b0e474bdae060be62ba1de3e11c). This identifies when the current asset choice was introduced; the earlier UI was not visually retested.
- On the latest source checkout, FastAPI `TestClient` confirmed that `/`, `/prompts`, `/shared`, `/skills`, `/reviews`, and `/handoff-reports` all return HTML containing the same favicon link when those pages are enabled. The PNG returns **HTTP 200**, **`image/png`**, a valid PNG signature, and the dimensions above. The normal server factory was also checked with an isolated temporary SQLite database for the default Dashboard pages.

The source and response checks confirm that a horizontal wordmark is being supplied for a small tab icon. This explains the screenshot's loss of legibility; the checked resource is present and served successfully. The screenshot is the reporter's original visual evidence, while the latest-master verification above covers the served HTML and asset.

#### Proposed solution

1. Add a dedicated square favicon containing only the PowerContext brand symbol, using an appropriate brand asset and sufficient padding. Provide a square PNG or a square-viewBox SVG; PNG/ICO fallbacks can be supplied for the supported browsers as needed.
2. Point the shared base template's favicon declaration to that asset with the correct media type and, for raster variants, matching `sizes`. A distinct asset URL will also help existing installations obtain the replacement favicon.
3. Keep the horizontal wordmark for the page header and footer.
4. Verify recognition and proportions at 16×16 and 32×32 in horizontal and vertical tabs, in light and dark browser themes, including after upgrading an existing installation. A focused regression check can verify that the rendered page references a successfully served, square favicon rather than the wordmark.

### Environment

- PowerContext source investigated: upstream `master`, commit `80d618bd61432d1bb4eae85433fa22f2bbd50bce`, checked on 2026-09-08.
- Investigation OS: Windows 11, build 26200, x64.
- Investigation Python: 3.12.13.
- Original screenshot: local Dashboard at `127.0.0.1:8766`, Chinese UI, vertical browser tabs. The exact browser version and running PowerContext revision at screenshot capture were not recorded.

Lkx-JY:
Closes #1507

## Rationale for this change

The Dashboard tab favicon pointed at /static/powercontext-color.png, a 1696×240 horizontal wordmark. Browsers squeeze it into a 16×16 / 32×32 square, which renders as a blurry, unrecognizable smear. The header/footer keep the wordmark legitimately; only the favicon slot misused it.

- Add a single square, symbol-only favicon (`favicon.svg`) cropped from the wordmark's symbol region, drawn on a fixed white rounded-square background with a slate outline so it stays readable on both light and dark browser tab bars regardless of the page theme toggle.
- Point the shared `base.html` `<link rel="icon">` at `favicon.svg` (`image/svg+xml`).
- Remove the dark variant (`favicon-reverse.svg`) and all theme-swapping logic (`data-favicon-*` attributes, the `page-ui.js` `applyTheme()` swap, and the head-script swap) — a single fixed-background icon needs no theme switching.
- Add a regression test asserting the icon references the square symbol asset (not the wordmark) and that it is served as a square SVG.

Yes. The browser-tab icon is now a clear square PowerContext symbol that stays readable in light and dark tab bars. Existing installations pick up the new asset automatically (new URL, no favicon-cache staleness). No API or data-format changes.

- `make check` — passed (lock consistency, `prek` incl. ruff, `ty check`, integration manifest).
- `pytest tests/test_dashboard.py tests/test_topics_dashboard.py` — 26 passed.
- Rendered the SVG at 16px and 32px on white and dark backgrounds to verify the symbol and outline remain readable.

## AI usage statement

Built with assistance from an AI coding agent (deepseek-v4-pro); the favicon assets were cropped/resized programmatically from the existing wordmark PNGs.


src/powercontext/server/static/page-ui.js
```diff
@@ -85,6 +85,10 @@ export function createPageUi(translations, onLocaleChange = () => {}) {
 
   function applyTheme(theme, persist = true) {
     document.documentElement.dataset.theme = theme;
+    const favicon = document.getElementById("favicon");
+    if (favicon) {
+      favicon.setAttribute("href", theme === "dark" ? favicon.dataset.faviconDark : favicon.dataset.faviconLight);
```
Teingi: [P2] Keep the favicon readable when page and browser themes differ

Switching the Dashboard to dark mode does not change the browser tab bar, but this selects the white `favicon-reverse.svg` anyway. With a light tab background, the white main stroke disappears and the 16px icon is reduced to a few gray strokes. The saved page theme also preserves this choice across reloads and navigation.

I verified the asset selection in Firefox with a light color preference, then rendered the actual SVG at 16px and 32px on a white background to confirm the contrast loss; this was not a capture of the native tab bar.

Could we use an icon that stays readable on both backgrounds, for example with an outline or a fixed background, and cover mismatched page/browser themes in validation?

Teingi (approved):
LGTM

PR #1515, opened 2026-09-08
docs(rfc): evolve the artifact processing supervisor

frf12:
# Which issue or RFC does this PR close?

Amends Topic Memory RFC 1417 and builds on the implementation delivered by #1490.

Builds on #1512 (merged), which assigned Topic Memory its RFC number and delivered the Topic Memory follow-up fixes.

# Rationale for this change

Topic Memory introduced `ArtifactProcessingSupervisor` as reusable background-processing infrastructure. This RFC evolves that design for multiple Artifact Families while keeping the Supervisor responsible only for scheduling. It preserves the existing binding identity, single-flight, Worker, retry, Lease, and fencing foundations, then adds per-artifact scheduling and capacity isolation.

# What changes are included in this PR?

- Define `global` as one Supervisor with one Lease and `dedicated` as one Supervisor and one Lease per registered Artifact Family.
- Keep `(binding_name, scope_id)` as the stable dispatch, persistence, and single-flight identity.
- Add per-artifact interval or cron policy, Worker budget, and Scope-invocation timeout.
- Move Source selection, Cursor advancement, generation, Review, and publication semantics behind each artifact processor.
- Replace cross-Scope automatic-wave completion barriers with bounded artifact-type admission and Scope discovery.
- Specify durable scheduling intent separately from artifact-owned business progress.
- Define compatibility and staged migration for the shipped Topic Memory Supervisor and the existing Memory, Experience, and Profile schedulers.
- Add equivalent English and Chinese RFC documents.

# Are there any user-facing changes?

This is a design-only RFC. It does not change current runtime behavior, public APIs, or persisted formats. The RFC proposes future configuration and persistence changes that require separate implementation and migration work.

# How was this change tested?

- `UV_INDEX_URL=https://pypi.tuna.tsinghua.edu.cn/simple UV_HTTP_TIMEOUT=120 make docs-test`
- Repository commit hooks, formatting checks, link validation, type checking, and static export verification passed.
- The documentation build generated and verified 665 public pages, including 66 RFC pages.

# AI usage statement

OpenAI Codex using GPT-6 was used to inspect the existing RFC and implementation, compare preserved and revised contracts, draft the English and Chinese documents, and run repository validation. The design decisions were reviewed interactively with the author.

PR #1523, opened 2026-09-09
feat: implement artifact processing supervisors

frf12:

Implements the Artifact Processing Supervisor RFC in #1515, building on #1512.

# Rationale for this change

Memory, Topic Memory, Experience, and Profile use separate background execution paths. A shared supervisor gives all four families durable requests, bounded scheduling, isolated Scope workers, and transactional completion so accepted work can recover after interruption.

- Persist processing intents, source dirtiness, scan checkpoints, and acknowledgments. Each family has independent concurrency limits, timeouts, bounded queues, and fair discovery. Support a global lease or dedicated leases per family.
- Freeze finite Topic Memory targets at worker creation and preserve the cumulative budgets and retrieval integrity from #1512 through partial completion and replay.
- Reconstruct worker identity and supported custom prompts; commit ownership, attestation, and processing completion atomically. Record Memory embedding usage.
- Check enabled Profile Policy before automatic admission and authorization. Missing or disabled policies retain dirty Sources without creating automatic requests, retries, or Worker slot consumption; enabling Policy makes retained Sources eligible.
- Keep Server authorization optional for SDK Topic workers. Validate unsupported custom Source registries before database bootstrap, so rejecting startup does not persist a deployment manifest that prevents a corrected restart. Preserve synchronous/API and custom-binding paths.
- Add resumable maintenance migration, metrics, lifecycle tracing, regression coverage, and English/Chinese documentation.
- Make processing regressions reliable under CI terminal formatting and asynchronous scan completion while preserving generation and checkpoint assertions.

Existing databases require offline migration through `powercontext server processing-migrate` plan/apply/verify. Stop old workers and pause writes and explicit triggers first. Switching supervisor mode also requires maintenance; mixed-mode rolling deployment is unsupported.

Family-specific concurrency and timeout settings replace shared settings. Legacy aliases remain supported; conflicting explicit values fail startup. SQLite and embedded SeekDB require a single-host `all` process; OceanBase supports separate API/background roles.

Built-in workers require reconstructible model configuration. Workers with Server security configured also require reconstructible authorization configuration. Built-in background workers require the supported built-in Source Definitions; unsupported custom registries fail at startup. Synchronous/API roles and fully custom processing bindings retain custom registry support. Synchronous API contracts, explicit Profile requests, and Profile/Experience review requirements are preserved. Configuration and migration documentation follows the current `operate/` layout with redirects from old migration URLs.

# How was this change tested?

Current HEAD: `2d3da2047714d77d722af4e2483a5336237dd49c`. Local and real-backend checks ran directly on this revision; GitHub Actions tested its merge with `master` at `4043710abc5410f17c50d7e2a11d889732fbccd9`.

- `make test`: 1,983 passed, 60 optional tests skipped; `make check`: passed.
- Startup registry/configuration regressions: 33 passed. Across four families and both supervisor modes, an unsupported registry leaves no database behind, and a corrected configuration opens the same path and reads/writes custom Sources. Successful deployment manifest changes still require maintenance.
- Required real storage matrix: 80 passed, zero skipped (18 OceanBase, 16 embedded SeekDB, 26 SQLite, 20 composition contracts).
- All 16 GitHub Actions checks and CLA passed, including Python 3.11–3.14, packages, native services, 47 contract tests, and documentation covering 777 public pages. [Main CI](https://github.com/oceanbase/powercontext/actions/runs/34339573006), [native services](https://github.com/oceanbase/powercontext/actions/runs/34339572997).
- Real generation/embedding used `deepseek-v4-flash-0731` and `text-embedding-v4` with the supplied credentials:

| Backend / mode | Fresh four-family processing | Confirmed-request replay |
| --- | --- | --- |
| SQLite / global | PASS 110.40s | PASS 118.55s |
| embedded SeekDB / global | PASS 163.65s | PASS 128.34s |
| OceanBase / dedicated | PASS 175.89s | PASS 115.56s |

All six runs verified persistent artifacts/candidates, custom prompts, ownership, Cursor/G acknowledgment, hybrid retrieval, and two Profile cron fires with missing/disabled Policy controls. Replays left all five recorded processing-purpose counters unchanged; explicit retrieval performs a separate recall embedding. Optional skips do not count toward required real-backend/provider acceptance. All 1,377 tracked files remained frozen during local validation.

Real-provider checks use reconstructible static-admin identity; enforced authorization ordering is covered by deterministic security regressions. Existing databases already initialized by an older failed startup retain the maintenance recovery requirement; this fix prevents unsupported registries from initializing new databases.

# AI usage statement

OpenAI Codex (GPT-6) assisted with implementation, tests, documentation, and preparation of this PR.


src/powercontext/builtin/runtime/topic_memory_processing.py
```diff
@@ -1593,14 +1640,22 @@ async def report(purpose: ModelUsagePurpose, operation: ModelUsageOperation, usa
                 datetime.now(UTC).date(),
             )
 
+        from powercontext.builtin.runtime.topic_memory_scope import TopicMemoryScopeProcessor
+        from powercontext.server.processing_security import open_worker_security
```
Teingi: [P2] Keep Server authorization optional for SDK Topic workers

This import runs even when `spec.worker_security` is `None`. In a clean environment containing only the locked `builtin` dependencies, a spawned Topic worker fails here with `ModuleNotFoundError: No module named 'casbin'`, leaving its request at `requested_generation=1, handled_generation=0`. `pycasbin` belongs to the `server` extra, so SDK-only Topic processing fails on every retry. Please guard the import and security initialization with `spec.worker_security is not None`, as `family_processing.py` already does.

src/powercontext/builtin/runtime/family_processing.py
```diff
@@ -0,0 +1,201 @@
...
+        _generation_pipelines,
+        _prompt_registry,
+        _usage_reporting_embedding_model,
+        open_builtin_contexts,
+    )
+
+    config = spec.config
+    async with AsyncExitStack() as resources:
+        pipelines = await _generation_pipelines(
+            config.inference, config.runtime, resources, None, BUILTIN_SOURCE_REGISTRY
```
Teingi, line 86: [P2] Preserve or reject custom Source registries before dispatch

`open_builtin_runtime(..., source_registry=...)` accepts custom Source Definitions, but this child builds its pipelines and contexts with the built-in registry. I reproduced this with a persisted custom `conversation` Source: Runtime startup succeeds, then the spawned Memory worker raises `RepositoryNotFoundError` for `source-definition 'conversation'`. The accepted request stays unhandled and the cursor cannot advance. Please carry reconstructible Source registration into workers, or reject unsupported registries at startup when built-in background workers are enabled.

Teingi: The registry check fixes the original worker failure, but there is still a P2 recovery issue on `22fef2f1`:

1. Start with a fresh SQLite database, a custom Source registry, and `artifact_processing_families=("memory",)`. Startup rejects the registry as expected.
2. The failed startup has already committed a `phase=complete` manifest with `capabilities=["memory"]`, despite having zero processing requests.
3. Follow the error message and set `ARTIFACT_PROCESSING_FAMILIES=[]`. Restarting against the same database now raises `ProcessingSchemaNotReadyError: Processing configuration differs from the completed maintenance manifest`. The same corrected configuration starts successfully against a fresh database.

`open_builtin_contexts()` commits the manifest before `_artifact_processing_bindings()` validates the registry. Please move this configuration-only validation ahead of database bootstrap so a rejected startup does not require a migration to correct its configuration. A regression test should reject the first configuration, disable built-in families, and successfully reopen the same database.

Teingi (approved):
LGTM

PR #1525, opened 2026-09-09
fix(cli): load local dotenv for server startup

### LINKED ISSUE #1524: bug: server run ignores model settings in local .env
## Describe the bug

`powercontext server run` does not automatically load a `.env` file from the current working directory. A user can configure both the generation and embedding models in `.env`, start the server normally, and still receive the inference capability notice saying that the models are missing or incomplete.

The command already supports `--env-file`, but requiring it in this common case is easy to miss and makes the result look like model detection is broken.

## Steps to reproduce

1. Create a `.env` file in the current directory:

   ```dotenv
   POWERCONTEXT_SERVER_INFERENCE_GENERATION_MODEL=<generation-model>
   POWERCONTEXT_SERVER_INFERENCE_EMBEDDING_MODEL=<embedding-model>
   POWERCONTEXT_SERVER_INFERENCE_EMBEDDING_PROFILE_ID=<embedding-profile-id>
   POWERCONTEXT_SERVER_INFERENCE_EMBEDDING_DIMENSION=<dimension>
   ```

2. Start PowerContext without explicitly passing the file:

   ```shell
   powercontext server run
   ```

3. Observe that PowerContext reports that the generation model or embedding model is not configured.

4. Start it again with `--env-file .env`; the model configuration is then detected.

## Expected behavior

Reuse `pydantic-settings` dotenv support and make env-file selection explicit and consistent at the CLI boundary:

- `powercontext server run` loads `./.env` when it exists.
- `powercontext server run --env-file <path>` loads only the specified file and does not also merge `./.env`.
- Provide `--no-env-file` so callers can explicitly disable automatic dotenv loading.
- Preserve the precedence `CLI options > process environment > selected env file > defaults`.
- When a file is loaded, print its resolved path without printing values or credentials.
- If no env file exists, continue using process environment and defaults without failing.

Prefer passing the selected path to `ServerSettings` through the runtime `_env_file` argument instead of copying dotenv values into `os.environ`. Do not set `env_file=".env"` globally on `ServerSettings.model_config`, because `ServerSettings()` is also used outside this CLI path and global current-working-directory discovery would make tests and library/service callers nondeterministic.

For `service install`, a background service must not discover configuration from its runtime working directory. If `./.env` is discovered during installation, resolve and persist its absolute path at install time and retain the existing ownership, permission, and file-identity validation. An explicit `--env-file` should continue to use the same protected-file behavior.

## Actual behavior

Only process environment variables are considered unless `--env-file` is passed explicitly. A `.env` beside the command invocation is ignored, so the missing/incomplete model notice can be shown even though the user has configured both model types in that file.

## Acceptance criteria

- Add regression coverage for automatic `./.env` discovery, explicit override, explicit disablement, missing `.env`, and configuration precedence.
- Confirm that dotenv values do not leak into the parent process environment.
- Confirm that service installation persists a deterministic absolute env-file path and retains existing security checks.
- Update CLI documentation and examples to describe the behavior.

## Additional context

This was observed while validating the inference capability notice introduced around #1504 and #1506.

Zxf-xufeng:

Closes #1524.

# Rationale for this change

`powercontext server run` did not automatically load a model or Server configuration stored in
the current directory's `.env` unless the user repeated that path with `--env-file`. This could
produce the missing or incomplete inference-model notice even though both models were configured.

The foreground Server should discover a local `.env` by default while keeping service installation
deterministic and preserving the existing protected environment-file behavior.

- Discover `./.env` only for `powercontext server run` when no explicit file is selected.
- Keep `ServerSettings()` free from implicit current-working-directory discovery.
- Support `--env-file <path>` to select one environment file and `--no-env-file` to disable both
  discovery and environment-file loading. These options are mutually exclusive.
- Preserve the foreground precedence `CLI options > process environment > selected env file > defaults`,
  including nested Pydantic settings represented either as JSON or exploded variables.
- Parse the selected file with PowerContext's strict, shell-free environment-file parser so values
  such as literal `${...}` tokens are not expanded.
- Expose non-Server provider variables from the selected file only for the bounded foreground
  Server runtime, without leaking them into the parent process. Provider variable names retain
  their platform-specific case semantics.
- Keep `powercontext service install` behavior unchanged: it does not discover the current
  directory's `.env`; an explicit `--env-file` remains the protected, persisted service
  configuration, while omitting it continues to use the process environment.
- Update English and Chinese configuration and deployment documentation.
- Add regressions for local-file discovery, precedence, nested settings, strict literal values,
  provider environment handling, override/disable behavior, environment restoration, and
  personal-service behavior.

Yes. `powercontext server run` now reports the selected environment file when it loads one. It
automatically discovers `.env` in the current directory, supports an explicit `--env-file`, and
supports `--no-env-file` when the process environment should be the only external configuration
source.

`powercontext service install` retains its existing explicit-file and process-environment behavior;
it does not implicitly use a current-directory `.env`.

# How was this change tested?

- `pytest tests/test_config_cli.py tests/test_cli.py tests/test_service.py -q` — 180 passed, 1 skipped
- `ruff check` on changed Python files
- `ruff format --check` on changed Python files
- `ty check`
- `git diff --check`
- Manual CLI help and configuration-precedence verification

# AI usage statement

OpenAI Codex (GPT-5) was used to inspect the configuration flow, implement the change, add tests
and documentation, and run validation. The resulting diff was reviewed before submission.


src/powercontext/server/configuration.py
```diff
@@ -60,28 +81,47 @@ def server_settings_context(
...
-    with loaded_context:
+    with loaded_context, data_dir_context:
         http_overrides: dict[str, Any] = {}
         if host is not None:
             http_overrides["host"] = host
         if port is not None:
             http_overrides["port"] = port
         settings_kwargs: dict[str, Any] = {"http": http_overrides} if http_overrides else {}
+        if env_file is not None:
+            settings_kwargs.update(_env_file=env_file, _env_file_encoding="utf-8")
```
Teingi: [P2] Preserve the existing env-file parsing semantics

This reparses the file with python-dotenv after `read_environment_file()` has already parsed it, but the two parsers disagree. For example, `POWERCONTEXT_SERVER_AUTH_TOKEN='literal-${PR1525_ABSENT}'` passes `config validate` but starts with the token `literal-` when that variable is unset. A token containing an apostrophe, serialized with the project's `shlex.quote()` format, also passes validation but makes startup exit with code 2. Both work on the base commit. Please keep validation, generated configuration, and startup on the same parsing semantics so credentials retain their exact values.

src/powercontext/server/configuration.py
```diff
@@ -60,28 +81,47 @@ def server_settings_context(
...
+    server_environment = {name for name in os.environ if name.startswith(_SERVER_ENVIRONMENT_PREFIX)}
+    if env_file is not None:
+        # ServerSettings reads POWERCONTEXT_SERVER_* through pydantic-settings' dotenv source.
+        # Provider SDKs and path helpers still consume their own variables from os.environ, so
+        # expose only those non-Server assignments for the bounded runtime context. Existing
+        # process values win over dotenv values and everything is restored on exit.
+        runtime_environment = {
+            name: value for name, value in loaded.items() if not name.startswith(_SERVER_ENVIRONMENT_PREFIX)
+        }
+        loaded_context = environment_context(runtime_environment, override=False)
```
Teingi: [P2] Keep service-install preflight consistent with the launcher

`service install` also calls `server_settings_context(env_file=...)` for its preflight, so this changes its behavior too. With `POWERCONTEXT_SERVER_HTTP_HOST=0.0.0.0` in the shell and `127.0.0.1` in a valid protected file, `service install --env-file ...` now exits with the unauthenticated non-loopback error before reaching the controller. The same file succeeds through `ServiceController`, and the CLI succeeds on the base commit. Please keep foreground precedence scoped to `server run` and make service preflight use the same authoritative file configuration as the controller and launcher.

src/powercontext/server/configuration.py
```diff
@@ -60,28 +81,47 @@ def server_settings_context(
...
-        environment_context(loaded, override=True, clear=server_environment)
-        if env_file is not None or environment is not None or data_dir is not None
+    server_environment = {name for name in os.environ if name.startswith(_SERVER_ENVIRONMENT_PREFIX)}
+    if env_file is not None:
+        # ServerSettings reads POWERCONTEXT_SERVER_* through pydantic-settings' dotenv source.
+        # Provider SDKs and path helpers still consume their own variables from os.environ, so
+        # expose only those non-Server assignments for the bounded runtime context. Existing
+        # process values win over dotenv values and everything is restored on exit.
+        runtime_environment = {
+            name: value for name, value in loaded.items() if not name.startswith(_SERVER_ENVIRONMENT_PREFIX)
```
Teingi: [P2] Filter Server variable names case-insensitively

This prefix check is case-sensitive, while `ServerSettings` reads environment names case-insensitively. With `POWERCONTEXT_SERVER_HTTP_PORT=8123` in the process and `powercontext_server_http_port=8999` in the file, the lowercase assignment gets injected into `os.environ` and the CLI uses port 8999. Passing the file directly to Pydantic correctly preserves port 8123. This breaks the promised process-environment precedence. Please normalize names for this check so lowercase and mixed-case Server assignments also stay in the dotenv source.

src/powercontext/server/configuration.py
Zxf-xufeng: Fixed in b28d1ca3.
server_settings_context() no longer passes the environment file to pydantic-settings for the server run path. The file is parsed exactly once with PowerContext's strict parser and applied in a bounded environment context.
This preserves literal values such as ${...} and shell-quoted values containing apostrophes. The process environment still has higher precedence for server run.
Added regression coverage for both literal ${...} values and shlex.quote()-style values containing apostrophes.

src/powercontext/server/configuration.py
Zxf-xufeng: Fixed in b28d1ca3.
The process-environment precedence is now explicitly scoped to server run. service install, the native service launcher, and the controller continue to use the historical environment-file-authoritative behavior.
This keeps the service-install preflight consistent with the configuration used by the launcher. In particular, a loopback bind from the protected environment file is no longer overridden by an unrelated shell value during service install.
Added a regression test covering a non-loopback shell value and a loopback value in the protected environment file.

src/powercontext/server/configuration.py
Zxf-xufeng: Fixed in b28d1ca3.
Server environment variable names are now compared case-insensitively when applying dotenv values and checking existing process variables. This prevents a lowercase or mixed-case Server variable from the environment file from overriding an equivalent process variable.
Added a regression test covering POWERCONTEXT_SERVER_HTTP_PORT in the process environment and powercontext_server_http_port in the environment file.

src/powercontext/server/configuration.py
Teingi: The original literal-value cases now pass. I still reproduced a [P2] source-precedence issue in b28d1ca3 that also affects fresh configurations:

- Process: `POWERCONTEXT_SERVER_HTTP='{"port":8123}'`
- File: `POWERCONTEXT_SERVER_HTTP_PORT=8999`
- Result: `server run` uses port 8999, although the documented process-over-file precedence requires 8123.

Comparing variable names before putting both sources into `os.environ` loses the distinction between a JSON setting and its nested fields. This remains a correctness issue even if backward compatibility is out of scope. If preserving the old parser's syntax is unnecessary, using Pydantic's dotenv settings source for Server fields would preserve source precedence without this manual merge.

src/powercontext/server/configuration.py
Teingi: The original Server-variable case now passes. There is another [P2] issue because the new `casefold()` filter also applies to provider variables, whose names are case-sensitive on Linux.

With lowercase `openai_api_key` in the process and a valid uppercase `OPENAI_API_KEY` in the file, the uppercase assignment is skipped. The OpenAI SDK reads the uppercase name, so initialization fails with `Missing credentials`. I reproduced this with the actual SDK; the combination is less common, but it also fails on a fresh install and does not depend on backward compatibility.

Please scope case-insensitive matching to Server settings and follow the host environment's name-comparison rules for provider variables.

src/powercontext/server/configuration.py
Zxf-xufeng: Fixed in 4d3c14ec.

The strictly parsed environment file is now supplied as a separate Pydantic settings source instead of flattening Server assignments into os.environ. The normal process-environment source remains higher priority, so a JSON parent value from the process environment beats an exploded child value from the file (for example, HTTP={"port":8123} versus HTTP_PORT=8999).

Added a regression test for this nested-source precedence case.

Validation: 180 passed, 1 skipped in the configuration, CLI, and service test suites; ty and Ruff also pass.

src/powercontext/server/configuration.py
Zxf-xufeng: Fixed in 4d3c14ec.

Case-insensitive matching is now limited to POWERCONTEXT_SERVER_ assignments. Non-Server provider variables are applied using their exact names, so on POSIX systems a lowercase process variable such as openai_api_key no longer masks OPENAI_API_KEY from the environment file.

Added a regression test covering this provider-variable case.

Validation: 180 passed, 1 skipped in the configuration, CLI, and service test suites; ty and Ruff also pass.

Teingi (approved):
LGTM

PR #1535, opened 2026-09-09
feat(client): add explicit insecure HTTP opt-in across integrations

### LINKED ISSUE #1534: feat: add explicit insecure HTTP opt-in across Agent integrations
## Problem

When a user copies a remote Dashboard URL such as `http://192.0.2.10:8000` into an Agent integration, setup or the prompt hook rejects it because plaintext HTTP is restricted to loopback. This is difficult to diagnose during remote development and acceptance testing, especially when the host continues normally after a hook failure.

Transport behavior also differs between the standalone Python hooks, TypeScript plugins, SDK adapters, and host-managed MCP connections.

## Proposed behavior

- Keep HTTPS and loopback HTTP as the default supported transports.
- Add an explicit client-side `--allow-insecure-http` opt-in for remote plaintext HTTP.
- Interactive setup resolves the effective URL (including environment configuration) and asks a default-No question before installation if remote HTTP needs consent. Explain that credentials and content travel unencrypted.
- Noninteractive and JSON setup never prompt; require an explicit flag or environment opt-in.
- Save the selected endpoint and consent for that host so a new Agent session uses the same decision. Saved consent does not silently authorize a different endpoint.
- Support `POWERCONTEXT_CLIENT_ALLOW_INSECURE_HTTP`, with host-specific overrides including explicit false.
- Show the insecure transport as a warning in diagnostics while still reporting actual connection results.
- Cover Codex, Claude Code, WorkBuddy, DSH, OpenCode, Pi, Hermes, OpenClaw, shared Python adapters, and document the host-managed MCP boundary for generic Agent Plugins and MiniMax.
- Do not change Server bind/authentication policy or disable TLS certificate validation.

## Acceptance

Test public/private HTTP addresses, HTTPS, IPv4/IPv6 loopback, CLI/environment/persisted precedence, declined and noninteractive consent, endpoint changes, fresh-session settings, and native MCP/Hook endpoint consistency. Installation alone must not be reported as full model/Agent end-to-end verification.

frf12:

Closes #1534.

# Rationale for this change

A remote PowerContext Server can be reachable from an Agent machine while PowerContext-owned hooks and clients reject its non-loopback HTTP URL. Users currently discover this late, and environment-only workarounds are not durable across sessions. The default should remain secure, with an explicit, visible opt-in for controlled plaintext deployments.

- Add endpoint-bound client transport consent, strict boolean parsing, common/host environment configuration and non-secret per-host persistence.
- Add `--server-url` and `--allow-insecure-http/--no-allow-insecure-http` to all eight setup targets and `setup select`; resolve the effective URL before installation, prompt with default No only in a TTY, and never prompt in JSON/noninteractive mode.
- Save endpoint and consent in `~/.config/powercontext/clients.json` (or `POWERCONTEXT_CLIENT_CONFIG_FILE`), preserving unrelated hosts and rolling back paired native writes on failure.
- Keep installed native MCP and hook endpoints aligned for Codex, Claude Code and WorkBuddy; synchronize Hermes/OpenClaw native settings.
- Apply the policy to standalone Python hooks, DSH/Pi/OpenCode/OpenClaw clients, the Python Client, LangChain, LangGraph, Pydantic AI and Bub. Rebuild shipped DSH/OpenCode bundles.
- Prevent saved/native permission for endpoint A from authorizing endpoint B after an environment/settings override. Explicit false remains authoritative.
- Add read-only native transport diagnostics. Explicitly insecure HTTP is `degraded`; unsupported or ambiguous native configuration is not reported as safe.
- Document the transport boundary for MiniMax and generic Agent Plugin, whose native MCP transport belongs to the host; add bilingual remote-installation guidance and refresh integration pages.

Yes. HTTPS and IPv4/IPv6 loopback HTTP remain available by default; non-loopback HTTP needs explicit consent. This intentionally tightens DSH/OpenClaw paths that previously accepted remote plaintext HTTP without consent.

Example:

```bash
powercontext setup claude-code --server-url http://192.0.2.10:8000 --allow-insecure-http --json
```

The flag does not disable TLS certificate verification, configure credentials, change Server binding/authentication, or override host-owned MCP policy. `doctor` returns a nonzero status for `degraded` plaintext transport even when the endpoint is reachable.

Codex uses its installed `.mcp.json` as the common hook/MCP endpoint; rerun setup after an update replaces it. DSH's standard empty patch is supported, but custom runtime overlays require manual alignment/removal before setup: the installer fails before mutation rather than silently saving an endpoint that the native override would ignore. Unsupported JSON5/includes and ambiguous cached endpoints are reported as unknown.

# How was this change tested?

- Full Python/doctest run: `python -m pytest --doctest-modules`, deselecting the four baseline failures listed below: **2,205 passed, 64 skipped, 4 deselected**; three packaging cases failed during dependency download (`pypi.org` TLS EOF for hatchling/hatch-vcs), not test assertions about the implementation. Re-ran all three package suites with `UV_FROZEN=true UV_INDEX_URL=https://pypi.tuna.tsinghua.edu.cn/simple python -m pytest tests/langchain_middleware/test_packaging.py tests/langgraph_adapter/test_packaging.py tests/pydantic_ai_adapter/test_packaging.py -q`: **3 passed**.
- Targeted setup/native-policy suites: **180 passed**; final native/setup profile and rollback regressions: **51 passed**. Core/SDK tests and wheel checks passed; Bub evaluation-adapter tests: **13 passed**.
- TypeScript package suites: **DSH 220**, **Pi 77**, **OpenCode 52**, **OpenClaw 65**; includes DSH live-server cases and Pi CLI loading. Rebuilt the shipped bundles and ran applicable package typechecks.
- `prek run -a`, `ruff check`, `ruff format --check`, `ty check`, `ty check integrations/pydantic-ai/src`, and `git diff --check`: passed.
- Website `pnpm lint` and `pnpm build`: passed; **785 public pages and internal links** verified. Integration-manifest consistency checks passed.
- Separately reproduced all four deselected failures on a clean detached **c33df8d3** baseline with baseline source imports verified: three Receiver systemd service tests hit the Linux-only guard on macOS; the installed OpenCode 1.2.15 rejects the test's deliberate `invalid/model` before its capture callback. These unrelated cases are unchanged.

Acceptance boundaries: validated on macOS with Python 3.12; no production deployment or blanket fresh-session acceptance across every real Agent/native MCP implementation is claimed. TLS verification regression tests are mocked rather than a live certificate-handshake test.

# AI usage statement

OpenAI Codex assisted with implementation, tests, documentation, and parallel code review under human-provided requirements. Review findings led to additional endpoint-binding, environment-refusal, native-diagnostics and persistence-rollback regressions.

### DIFF
--- a/integrations/claude-code/plugins/powercontext/.claude-plugin/plugin.json
+++ b/integrations/claude-code/plugins/powercontext/.claude-plugin/plugin.json
@@ -28,6 +28,11 @@
       "title": "Capture user prompts",
       "description": "Store user prompts as ordinary Content Source evidence",
       "default": true
+    },
+    "allow_insecure_http": {
+      "type": "boolean",
+      "title": "Allow unencrypted HTTP",
+      "description": "Explicitly allow HTTP to a non-loopback PowerContext Server; TLS certificate verification remains enabled"
     }
   }
 }
--- a/integrations/langchain/src/powercontext_langchain/settings.py
+++ b/integrations/langchain/src/powercontext_langchain/settings.py
@@ -14,8 +14,12 @@
 
 """Process configuration for the PowerContext LangChain middleware."""
 
+from typing import ClassVar
+
 from pydantic import Field, SecretStr, field_validator
-from pydantic_settings import BaseSettings, SettingsConfigDict
+from pydantic_settings import SettingsConfigDict
+
+from powercontext.client.transport_policy import ClientTransportSettings
 
 try:
     from powercontext.http import ContextAssembly
@@ -24,7 +28,7 @@ except ImportError:
     from types import NoneType as ContextAssembly
 
 
-class PowerContextLangChainSettings(BaseSettings):
+class PowerContextLangChainSettings(ClientTransportSettings):
     """PowerContext settings read from ``POWERCONTEXT_LANGCHAIN_*`` variables."""
 
     model_config = SettingsConfigDict(
@@ -35,6 +39,7 @@ class PowerContextLangChainSettings(BaseSettings):
         hide_input_in_errors=True,
     )
 
+    transport_host: ClassVar[str] = "langchain"
     base_url: str = "http://127.0.0.1:8000"
     token: SecretStr | None = Field(default=None, repr=False)
     scope_id: str | None = None
--- a/src/powercontext/cli/native_transport.py
+++ b/src/powercontext/cli/native_transport.py
@@ -1,0 +1,211 @@
+# Copyright (c) 2026 OceanBase.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Read native host transport settings without guessing unsupported configuration."""
+
+from __future__ import annotations
+
+import json
+import os
+import re
+from pathlib import Path
+from typing import Any, cast
+
+from powercontext.client.transport_policy import (
+    load_client_settings,
+    normalize_client_url,
+    parse_client_boolean,
+    resolve_client_transport,
+)
+
+_DEFAULT_URL = "http://127.0.0.1:8000"
+_MISSING = object()
+_UNKNOWN = "Cannot determine PowerContext transport from unsupported or unreadable native host configuration"
+
+
+def _home(variable: str, directory: str) -> Path:
+    configured = os.environ.get(variable, "").strip()
+    return Path(configured).expanduser() if configured else Path.home() / directory
+
+
+def _object_at(value: object, *keys: str) -> dict[str, Any]:
+    for key in (*keys, None):
+        if not isinstance(value, dict) or "$include" in value:
+            raise ValueError(_UNKNOWN)
+        if key is None:
+            return cast(dict[str, Any], value)
+        value = value.get(key, {})
+    return {}  # pragma: no cover
+
+
+def _read(path: Path) -> dict[str, Any]:
+    try:
+        return _object_at(json.loads(path.read_text(encoding="utf-8")))
+    except FileNotFoundError:
+        return {}
+    except (OSError, UnicodeError, ValueError):
+        raise ValueError(_UNKNOWN) from None
+
+
+def _url(value: object) -> str:
+    if not isinstance(value, str) or not value.strip() or "${" in value:
+        raise ValueError(_UNKNOWN)
+    try:
+        return normalize_client_url(value).removesuffix("/mcp").rstrip("/")
+    except ValueError:
+        raise ValueError(_UNKNOWN) from None
+
+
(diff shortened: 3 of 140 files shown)


integrations/langchain/src/powercontext_langchain/settings.py
```diff
@@ -14,8 +14,12 @@
 
 """Process configuration for the PowerContext LangChain middleware."""
 
+from typing import ClassVar
+
 from pydantic import Field, SecretStr, field_validator
-from pydantic_settings import BaseSettings, SettingsConfigDict
+from pydantic_settings import SettingsConfigDict
+
+from powercontext.client.transport_policy import ClientTransportSettings
```
Teingi, line 22: **[P1] Update the core dependency before importing the new transport module**

The adapter dependency ranges still allow the published `powercontext==0.2.0`, which has no `powercontext.client.transport_policy`. With that wheel, the baseline LangChain, LangGraph, and Pydantic AI adapters all import successfully; all three adapters from this PR fail with `ModuleNotFoundError` on this new import. Installing the adapter can therefore succeed while importing it fails. Please require a core release containing this module, or provide compatibility support, and cover installation against the declared core dependency range.

integrations/claude-code/plugins/powercontext/.claude-plugin/plugin.json
```diff
@@ -28,6 +28,11 @@
       "title": "Capture user prompts",
       "description": "Store user prompts as ordinary Content Source evidence",
       "default": true
+    },
+    "allow_insecure_http": {
```
Teingi, line 32: **[P2] Bump the Claude plugin and marketplace versions for this change**

Both versions remain `0.1.0`. On Claude Code 2.1.197, I installed the baseline plugin, updated the marketplace source to this PR, then ran marketplace update, plugin update, and setup with `--allow-insecure-http`. Setup returned success and saved consent, but Claude reported the plugin was already current and kept the old cached code. That cache lacks the new option/helper, so its hook still rejects remote HTTP. Please bump both versions and verify upgrading an existing installation.

src/powercontext/cli/native_transport.py
```diff
@@ -0,0 +1,211 @@
...
+    common_url = _environment_url("POWERCONTEXT_CLIENT_SERVER_URL")
+    saved_url = load_client_settings(host).get("server_url")
+    environment_url = _environment_url(prefix + "_BASE_URL", prefix + "_SERVER_URL", prefix + "_ENDPOINT")
+    if host == "openclaw" and not (environment_url or common_url or native_url or saved_url):
+        raise ValueError("PowerContext endpoint is not configured in OpenClaw")  # noqa: TRY003
+    fallback = saved_url or _DEFAULT_URL
+    if host == "codex":
+        return _codex_url() or fallback
+    if host == "claude-code":
+        return _environment_url(prefix + "_SERVER_URL") or native_url or common_url or fallback
```
Teingi, line 162: **[P2] Check the Claude MCP endpoint separately from the hook override**

`POWERCONTEXT_CLAUDE_SERVER_URL` changes the hook endpoint, but native MCP still uses `${user_config.server_url}/mcp`. After a fresh setup with `http://192.0.2.10:8000 --allow-insecure-http`, setting that environment variable to `https://safe.example` makes `doctor claude-code --json` return `ok: true` and exit 0, although the installed MCP configuration still points to remote plaintext HTTP. Please check both effective endpoints and report their mismatch or the remaining insecure MCP transport instead of letting the HTTPS hook override hide it.

Teingi (approved):
LGTM

PR #1536, opened 2026-09-09
feat: add Topic Memory prompt management and Dashboard support

frf12:
## Summary

Adds Scope-level Prompt management for Topic Memory generation stages and read-only Topic Memory/Profile Dashboard views.

## Validation

- Topic Memory processing tests: 40 passed
- Dashboard self-test: Topic Memory/Profile pages, exact revision status, and paginated browsing
- Generated API check, compileall, ruff, and diff check: passed

The prompt demonstration and stage-validation follow-ups remain with the prompt-management implementation owner.


src/powercontext/builtin/runtime/composition.py
```diff
@@ -311,6 +311,11 @@ async def open_builtin_runtime(
             ("experience.generate", experience_generator, generated_experience),
             ("skill.generate", skill_generator, generated_skill),
             ("handoff.generate", handoff_pipeline, generated_handoff),
+            *(
+                (f"topic_memory.{stage}", None, object())
```
Teingi, line 315: [P2] Register demonstration generators for the Topic Prompt keys

These seven keys are advertised as `supported`, but `_register_prompt_demonstrators()` still registers only the existing keys. With the same configured model, all seven Topic demonstration endpoints return `422 prompt_customization_unavailable`, while `memory.extract` returns 200. Please register their demonstration generators as well so the newly exposed endpoints work.

src/powercontext/builtin/artifacts/prompt/definitions.py
```diff
@@ -73,11 +73,14 @@ def validate(self, content: PromptContent, /, *, during_inference: bool = False)
...
+                # Topic stage contracts use tuple fields for runtime immutability;
+                # JSON demonstrations naturally decode arrays into those tuples.
+                strict = not self.key.startswith("topic_memory.")
                 value = self.input_type.model_validate_json(
-                    json.dumps(demonstration.input), strict=True, extra="forbid"
+                    json.dumps(demonstration.input), strict=strict, extra="forbid"
                 )
                 output = self.output_type.model_validate_json(
-                    json.dumps(demonstration.expected_output), strict=True, extra="forbid"
+                    json.dumps(demonstration.expected_output), strict=strict, extra="forbid"
```
Teingi, line 83: [P2] Validate Topic demonstrations against the stage's reference rules

`validate_demonstration()` has no branches for the new Topic types. I could save a Probe example citing an absent evidence ID, a Planner example referencing an absent probe, and a Reduce example with `covered_indices: [99]` for one input; all returned 201. These examples are then included in the model instructions even though the runtime rejects the corresponding outputs. Please validate the stage-specific reference and coverage rules before accepting demonstrations.

src/powercontext/server/dashboard/templates/topics.html
```diff
@@ -0,0 +1,43 @@
...
+        <div class="d-flex justify-content-between gap-2"><strong>{{ item.title }}</strong><code>@{{ item.artifact.revision }}</code></div><p class="text-secondary mb-0 mt-1">{{ item.summary|truncate(140, true) }}</p>
+      </a>{% endfor %}</div>
+    {% else %}<div class="card-body"><h3>{{ t.topic_memory_empty }}</h3><p class="text-secondary mb-0">{{ t.topic_memory_empty_body }}</p></div>{% endif %}
+  </div></section>
+  <section class="col-lg-7" aria-labelledby="topic-detail-heading"><div class="card"><div class="card-body">
+    {% if errors.get('topic_memory_selected') %}{{ read_error(errors.topic_memory_selected, t.topic_memory) }}
+    {% elif data.topic_memory_selected %}{% set topic = data.topic_memory_selected %}
+      <div class="d-flex justify-content-between gap-3"><div><h2 class="card-title mb-1" id="topic-detail-heading">{{ topic.title }}</h2><code>topic-memory/{{ topic.artifact.artifact_id }}@{{ topic.artifact.revision }}</code></div><span class="badge bg-green-lt">{{ t.current_revision }}</span></div>
```
Teingi: [P2] Show the selected revision's actual publication status

The badge always says "Current revision", including for historical links. After publishing revision 2, opening the revision-1 URL still shows the old content with this green badge, although the repository reports `is_current=False`. Please use the actual current-head status when rendering the badge so an old bookmark does not present superseded content as current.

src/powercontext/server/dashboard/api.py
```diff
@@ -143,3 +144,28 @@ async def load(item: dict[str, Any]) -> dict[str, Any]:
...
+        items = await application.topic_memory.for_scope(scope).browse(limit=limit)
+        return {
+            "items": [
+                {
+                    **item.model_dump(mode="json"),
+                    "artifact": item.artifact_ref.model_dump(mode="json"),
+                }
+                for item in items
+            ],
+            "next_cursor": None,
```
Teingi: [P2] Preserve a cursor for browsing more than 50 topics

This reads only the first 50 heads and always reports `next_cursor=None`; the page also has no pagination control. With 51 published topics, I could browse only the newest 50, with no way to reach the oldest through the list. Please pass through the repository's existing keyset cursor and add a next-page link.

Issue #1533, opened 2026-09-09
feat: expose all supported artifact families in the Dashboard


frf12:
## Problem

The current Dashboard does not expose all artifact families supported by the Runtime. In the current UI, the sidebar shows:

- Home
- Handoff
- Memory
- Experience & Skills
- Usage

However, there is no dedicated Dashboard entry for Topic Memory, Profile, or Prompt configuration. This makes backend capabilities difficult to discover and prevents users from inspecting or managing artifacts that are already supported by the Server/API.

The attached user screenshot shows the current artifact navigation state.

## Expected behavior

The Dashboard should provide a discoverable view for every supported artifact family, or clearly explain when a family is API-only. At minimum, the UI should expose or link to:

- Memory
- Topic Memory
- Experience
- Skill
- Handoff
- Profile
- Prompt configuration
- Usage/diagnostics

Topic Memory should show active topics, revisions/publications, source lineage, and processing status. Prompt configuration should show auto/custom mode, effective instructions, and the configured scope.

## Acceptance criteria

- Dashboard navigation reflects the artifact families advertised by the Server capabilities/API.
- Topic Memory is inspectable without requiring direct API calls or SQLite access.
- Profile and Prompt configuration have either dedicated views or explicit, documented API-only status.
- Empty states and disabled capabilities are represented clearly instead of silently omitting the artifact.

PR #1537, opened 2026-09-09
feat: add Topic Memory dashboard

### LINKED ISSUE #1544: Add read-only Topic Memory Dashboard
## Scope

This issue narrows the Topic Memory portion of #1533 into a standalone deliverable.

- Add a discoverable read-only `/dashboard/topics` view.
- Support browsing, search, exact revision details, source references, and clear empty/no-match states.
- Keep the view read-only and aligned with the existing Topic Memory API.

Profile Dashboard and Prompt configuration are out of scope for this issue.

## Acceptance criteria

- Topic Memory can be inspected from the Dashboard without direct API or SQLite access.
- Active topics, revisions/publications, source lineage, and processing status are represented where available.
- Targeted tests and lint validation pass.

frf12:
Fixes #1544

Implements the read-only Topic Memory Dashboard requested by GitHub issue #1533.

- Adds `/dashboard/topics` with browse, search, exact revision detail, and source references.
- Uses `POST /v1/topic-memory/search` with the strict API contract (`scope_id`, `query`).
- Shows clear empty and no-match states.
- Does not add Profile or Prompt configuration pages.

Validation: targeted dashboard tests (2 passed), Ruff, and `git diff --check`.


src/powercontext/server/dashboard/templates/topics.html
```diff
@@ -0,0 +1,42 @@
...
+        <div class="d-flex justify-content-between gap-2"><strong>{{ item.title }}</strong><code>@{{ item.artifact.revision }}</code></div><p class="text-secondary mb-0 mt-1">{{ item.summary|truncate(140, true) }}</p>
+      </a>{% endfor %}</div>
+    {% else %}<div class="card-body"><h3>{{ t.topic_memory_empty }}</h3><p class="text-secondary mb-0">{{ t.topic_memory_empty_body }}</p></div>{% endif %}
+  </div></section>
+  <section class="col-lg-7" aria-labelledby="topic-detail-heading"><div class="card"><div class="card-body">
+    {% if errors.get('topic_memory_selected') %}{{ read_error(errors.topic_memory_selected, t.topic_memory) }}
+    {% elif data.topic_memory_selected %}{% set topic = data.topic_memory_selected %}
+      <div class="d-flex justify-content-between gap-3"><div><h2 class="card-title mb-1" id="topic-detail-heading">{{ topic.title }}</h2><code>topic-memory/{{ topic.artifact.artifact_id }}@{{ topic.artifact.revision }}</code></div><span class="badge bg-green-lt">{{ t.current_revision }}</span></div>
```
Teingi: **[P2] Check publication state before labeling a revision as current**

This badge is unconditional, but the URL selects an exact revision. I reproduced this with SQLite and Firefox: after publishing revision 2, reopening the revision 1 link still shows the old content with “Current revision”, even though the Runtime reports `is_current=False`. A saved link can therefore present outdated guidance as current. Please distinguish historical revisions using the actual publication state and provide a link to the current revision.

src/powercontext/server/dashboard/api.py
```diff
@@ -143,3 +144,28 @@ async def load(item: dict[str, Any]) -> dict[str, Any]:
...
+        items = await application.topic_memory.for_scope(scope).browse(limit=limit)
+        return {
+            "items": [
+                {
+                    **item.model_dump(mode="json"),
+                    "artifact": item.artifact_ref.model_dump(mode="json"),
+                }
+                for item in items
+            ],
+            "next_cursor": None,
```
Teingi: **[P2] Allow browsing beyond the first 50 topics**

The browse call defaults to 50 items, but `next_cursor` is always `None` and the page has no pagination controls. With 51 published topics in one Scope, I confirmed that the browser shows only the newest 50. The omitted topic is still searchable, but users cannot discover it by browsing the inventory. Please pass through the existing `after` cursor support and add navigation to subsequent pages.

src/powercontext/server/dashboard/templates/topics.html
frf12: Addressed in commit `8f86d5c6`: the dashboard now uses the exact Topic Memory publication state (`is_current`) and labels historical revisions explicitly. Historical pages also link to the current revision.

src/powercontext/server/dashboard/api.py
frf12: Addressed in commit `8f86d5c6`: browse now passes the existing keyset cursor through the dashboard, fetches one extra item to determine whether another page exists, and renders previous/next navigation while preserving scope and filters.

Teingi (approved):
LGTM

PR #1473, opened 2026-09-05
feat(prompts): add scope-owned prompt management

Teingi:
# Which issue or RFC does this PR close?

Closes #1465.

Implements the Scope-owned prompt management design in #1468.

# Rationale for this change

Agents and users need different memory extraction rules without bypassing PowerContext's Scope isolation. The same requirement applies to other built-in generation and reranking operations. This change makes operational prompts Scope-owned, versioned Artifacts while preserving server-owned schemas, evidence checks, model settings, and authority boundaries.

# What changes are included in this PR?

- Register the `prompt` Artifact family for `memory.extract`, `memory.rerank`, `experience.incubate`, `experience.generate`, `skill.generate`, and `handoff.generate`.
- Support Auto/custom modes and strict content containing `schema_version`, `mode`, `instructions`, and ordered typed `demonstrations` with `input` and `expected_output`. Validate operation-specific reference and output semantics as well as JSON structure.
- Reuse existing Artifact storage and conditional writes: no new database tables. Provide signed, Scope-bound revision-history pagination; restoring an old revision creates a new revision rather than moving the head backward.
- Resolve and freeze the effective prompt for an operation and its retries, including before coarse memory retrieval. Preserve injected-component interfaces and expose supported/disabled/injected capabilities.
- Record bounded prompt identity and digest metadata without adding prompt bodies to tracing. Preserve exact configuration lineage on generated Artifacts, and use signed generation receipts for Handoff prepare/finalize/commit attribution. Prompt configuration is not factual evidence.
- Add the `/prompts` Dashboard with Scope selection, instructions, positive/negative case groups, unsaved case generation, version history, restore, and ETag conflict handling.
- Update OpenAPI, generated Python models/operations, the HTTP client, and integration operation tables. The only two new HTTP operations are:

  - `GET /v1/scopes/{scope_id}/artifacts/{family}/{artifact_id}/revisions`
  - `POST /v1/scopes/{scope_id}/prompts/{prompt_key}/demonstrations`
- Include two fixes exercised by real-provider acceptance: use SQLite `lastrowid` instead of unsupported `INSERT ... RETURNING` on older SQLite, and validate generated Skill package constraints inside the existing bounded LLM retry loop.

# Are there any user-facing changes?

- Prompt customization is available through existing Artifact CRUD, the two operations above, and the Dashboard. Agent/user ownership continues to resolve through existing Scope bindings; there is no separate ownership or precedence model.
[...]

PR #1541, opened 2026-09-10
feat(profile): support scoped custom prompts

Zxf-xufeng:

Related to #1468, #1473, and #1536.

# Rationale for this change

Scope-owned Prompt management supports Memory, Experience, Skill, and Handoff operations, but Profile generation still uses fixed server instructions. This leaves Profile as the remaining generated artifact that cannot apply scoped guidance or record the exact Prompt revision used for generation.

- Register `profile.generate` as a Prompt key with the existing Profile input/output contract and unchanged Auto instructions.
- Bind the scoped Prompt selection around manual and automatic Profile flushes, including spawned family workers.
- Pass compiled Custom instructions and demonstrations into the built-in structured generator.
- Record the selected Custom Prompt artifact in both directly generated Profile lineage and review-required candidate lineage.
- Keep evidence trust, person attribution, sensitive-attribute safeguards, and output shape as server-owned invariants.
- Report injected `ProfileGenerator` components as unsupported for managed Prompt customization, matching the existing component capability model.
- Update the OpenAPI-generated client models, bilingual workflow documentation, and behavior/E2E tests.

Yes. Clients can create or replace a Scope-owned Prompt artifact with `prompt_key=profile.generate`. The Prompt affects future Profile processing windows only; saving it does not reprocess historical Sources. This is additive and requires no data migration.

# How was this change tested?

- `make check`
- `make contract-test` (47 passed)
- Focused Prompt/Profile/runtime/HTTP tests (79 passed)
- Full macOS-executable suite excluding Linux-only systemd receiver-service tests (2066 passed, 61 skipped)

# AI usage statement

Implemented and reviewed with OpenAI Codex (GPT-5), including code analysis, test generation, and validation.

### DIFF
--- a/src/powercontext/builtin/artifacts/profile/service.py
+++ b/src/powercontext/builtin/artifacts/profile/service.py
@@ -61,6 +61,7 @@ from powercontext.builtin.sources import SourceCursor
 from powercontext.errors import RevisionConflictError
 
 if TYPE_CHECKING:
+    from powercontext.builtin.artifacts.prompt.service import PromptService
     from powercontext.builtin.runtime.processing_execution import ScopeInvocation
 
 
@@ -77,6 +78,14 @@ class ProfileGenerator(Protocol):
     async def generate(self, value: ProfileGenerationInput, /) -> str | None: ...
 
 
+def _profile_prompt_refs():
+    # Import locally because Prompt Definitions import the Profile generation contract.
+    from powercontext.builtin.artifacts.prompt.service import current_prompt
+
+    selection = current_prompt("profile.generate")
+    return () if selection is None or selection.artifact is None else (selection.artifact,)
+
+
 class RelationalProfileService:
     def __init__(
         self,
@@ -86,6 +95,7 @@ class RelationalProfileService:
         candidates: CandidateRepository,
         *,
         generator: ProfileGenerator | None = None,
+        prompt_service: PromptService | None = None,
         id_factory: Callable[[str], str] | None = None,
         max_sources: int = 32,
     ):
@@ -98,6 +108,7 @@ class RelationalProfileService:
         self.policies = ProfilePolicyRepository()
         self.cursors = SourceCursorRepository()
         self.generator = generator
+        self._prompt_service = prompt_service
         self.operation_context: Callable[[], AbstractAsyncContextManager[Any]] = nullcontext
         self.scheduled_runner: Callable[[str, int], Awaitable[ProfileFlushResult]] | None = None
         self.max_sources = max_sources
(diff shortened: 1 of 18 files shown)


src/powercontext/builtin/artifacts/profile/service.py
```diff
@@ -264,6 +285,7 @@ async def _flush(  # noqa: C901
                 updated = await self.policies.update(connection, policy)
                 refs = tuple(item.ref for item in evidence)
                 parents = () if current is None else (current.as_ref(),)
+                artifacts = (*parents, *_profile_prompt_refs())
```
Teingi: [P2] Reserve evidence capacity for the selected Prompt

With a Custom `profile.generate` Prompt and `activation_mode="review_required"`, a full window exceeds the 32-reference candidate limit: 32 Sources + Prompt for the first Profile, or 31 Sources + previous Profile + Prompt for an update. I reproduced both on SQLite: each flush raises `ValidationError: Candidate evidence must not exceed 32 references`, and retrying invokes the generator again without advancing the cursor. Please reserve capacity for the Prompt when computing the Source window and cover both cases.

src/powercontext/builtin/artifacts/prompt/builtin.py
```diff
@@ -151,4 +153,21 @@ def builtin_prompt_definitions(
 """,
             default_instructions=HANDOFF_GENERATION_INSTRUCTIONS,
         ),
+        PromptDefinition(
+            key="profile.generate",
+            definition_version="powercontext.prompt.profile.generate.v1",
+            input_type=ProfileGenerationInput,
+            output_type=ProfileGenerationOutput,
```
Teingi, line 160: [P2] Validate Profile demonstration content against the runtime contract

`ProfileGenerationOutput` accepts blank strings, and `validate_demonstration()` has no Profile check. I verified that `{"content":"   "}` is accepted by both demonstration generation and Prompt creation, while the built-in Profile generator returning the same output causes flush to raise `InvalidInferenceOutputError` during Markdown validation. Please apply the Profile Markdown validation to non-null demonstration outputs and use `content=null` for no-op examples.

src/powercontext/builtin/artifacts/profile/service.py
Zxf-xufeng: Fixed in 595ecd76. Profile generation now reserves lineage capacity for both the current Profile revision and the selected Custom Prompt before loading the Source window. I added parameterized regression coverage for full review-required windows with and without an existing Profile, verifying that the combined evidence remains within the 32-reference limit and that the exact Prompt revision is retained in Candidate lineage.

src/powercontext/builtin/artifacts/prompt/builtin.py
Zxf-xufeng: Fixed in 595ecd76. Profile demonstration validation now applies the same Markdown validation used by the runtime to every non-null output. Blank content is rejected, while content=null remains the valid no-op representation. I also added coverage for Prompt validation, null no-op examples, and generated demonstrations retrying after blank Markdown.

Teingi (approved):
LGTM

PR #1548, opened 2026-09-10
feat(setup): persist PowerContext Agent authorization

### LINKED ISSUE #1538: feat(setup): persist PowerContext Agent authorization
## Feature description

Persist PowerContext Server authorization during Agent setup so users can start a supported coding agent normally after one setup run, without exporting a host-specific Authorization variable for every session.

This applies setup-managed persistence to Codex, Claude Code, DeepSeek Harness (DSH), OpenCode, Pi, and WorkBuddy. OpenClaw and Hermes already have native persistent secret inputs and should retain those flows.

Today an authenticated PowerContext Server works only when several Agent processes inherit variables such as `POWERCONTEXT_CODEX_AUTHORIZATION` or `POWERCONTEXT_DSH_AUTHORIZATION`. This is easy to forget, makes MCP and Hook behavior diverge, and previously led documentation to suggest sourcing the complete Server `.env`, which can unnecessarily expose Generation Provider and database credentials to an Agent process.

Proposed setup contract:

- Interactive `powercontext setup <host>` asks once for an optional token using hidden input. Non-interactive and `--json` setup read the one-shot `POWERCONTEXT_CLIENT_API_TOKEN` input.
- Do not accept a token as a CLI argument; do not print it in human, JSON, or diagnostic output.
- Preserve host-specific Authorization environment variables as runtime overrides.
- Bind every persisted credential to the normalized Server URL from the setup transport contract in #1535, and never forward a saved credential to a different URL.
- Store Codex, Claude Code, OpenCode, and Pi credentials in PowerContext-owned files below each host's user configuration root, with atomic writes and `0600` permissions on POSIX.
- For DSH, write the complete Bearer header to the Web profile patch's `powercontext-dsh.config.authorization`, preserving all unrelated patch fields. Detect a higher-priority conflicting home patch instead of silently writing an ineffective value.
- For WorkBuddy, write the credential to the owned MCP entry and pin the matching URL while persisted authentication is active, because its independent string templates cannot enforce URL affinity.
- Add `--clear-authorization`, idempotent credential states in setup JSON, and redacted local authorization diagnostics.
- Make MCP, automatic Hooks, and explicit PowerContext tools resolve the same credential source.
- Remove instructions to source an entire Server environment before launching an Agent.

`setup select` should cover all eight `agent_host` integrations from `integrations/capabilities.toml`; Hermes and OpenClaw report that authorization is managed by their native flow.

## Acceptance criteria

- A newly started Codex, Claude Code, DSH, OpenCode, Pi, or WorkBuddy session can use an authenticated PowerContext Server after setup with no authorization export.
- Environment Authorization overrides continue to work without overwriting saved credentials.
- URL mismatch, malformed storage, unsafe POSIX permissions, and conflicting DSH patch precedence fail closed for the credential and are visible through redacted diagnostics.
- Re-running setup without a token preserves an existing credential; re-running with a token rotates it; explicit clear removes only PowerContext authorization state.
- Setup output, logs, test failures, and diagnostics never contain the token or full Authorization header.
- OpenClaw and Hermes native persistent flows remain compatible.
- Focused Python/TypeScript suites and repository CI pass.

## Alternatives considered

- Continue requiring per-launch exports: too error-prone for the normal user journey.
- Source the Server `.env`: leaks unrelated Server authority into Agent processes.
- Add an OS keychain abstraction or loopback credential proxy: stronger isolation but disproportionate complexity for this local, user-owned setup flow; it can be revisited separately.
- Reuse Codex/Claude provider credentials: incorrect security boundary; those credentials authenticate different services.

## Additional context

This is independent from the Claude SDK OAuth response-shape bug in #1530. It depends on the setup-time endpoint and insecure-HTTP consent contract in #1535 and should reuse that implementation rather than duplicate transport state.

## Are you willing to submit a PR?

Yes.

frf12:
Closes #1538.

## Summary

- Persist a setup-time PowerContext authorization in a host-owned, URL-bound credential file.
- Load credentials for Codex, Claude Code, DSH, OpenCode, Pi, and WorkBuddy without requiring an export before every session.
- Keep credentials out of plugin configuration and diagnostics; reject unsafe permissions, malformed records, and URL mismatches.
- Add setup authorization state reporting and user-facing quick-start documentation.

## Validation

- make check
- Focused Python suites: 111 passed
- OpenCode: 25 passed
- Pi: 50 passed
- DSH: 177 passed

This PR supersedes the closed draft PR #1539 and contains implementation, tests, and user-facing documentation only.

## AI usage

OpenAI Codex was used to inspect the repository, implement the change, and run validation. The human selected the host-owned credential persistence policy.

### DIFF
--- a/integrations/codex/plugins/powercontext/settings.py
+++ b/integrations/codex/plugins/powercontext/settings.py
@@ -17,6 +17,9 @@
 from __future__ import annotations
 
 import ipaddress
+import json
+import os
+import stat
 from pathlib import Path
 from typing import Any, Literal
 from urllib.parse import urlsplit, urlunsplit
@@ -85,7 +88,9 @@ class _McpEndpointSettingsSource(PydanticBaseSettingsSource):
 
     @override
     def __call__(self) -> dict[str, Any]:
-        return {"server_url": _server_url_from_mcp_configuration()}
+        server_url = _server_url_from_mcp_configuration()
+        authorization = _stored_authorization(server_url)
+        return {"server_url": server_url, **({"authorization": authorization} if authorization else {})}
 
 
 class CodexPluginSettings(BaseSettings):
@@ -179,6 +184,35 @@ class CodexPluginSettings(BaseSettings):
 def _server_url_from_mcp_configuration() -> str:
     configuration = _McpConfiguration.model_validate_json(_MCP_CONFIGURATION_PATH.read_text())
     return _http_base_url(configuration.mcp_servers["powercontext"].url, allow_insecure_http=True)
+
+
+def _stored_authorization(server_url: str) -> str | None:
+    path = Path(os.environ.get("CODEX_HOME", Path.home() / ".codex")).expanduser() / "powercontext" / "credentials.json"
+    try:
+        if path.is_symlink() or not path.is_file() or (os.name != "nt" and stat.S_IMODE(path.stat().st_mode) & 0o077):
+            return None
+        payload = json.loads(path.read_text(encoding="utf-8"))
+        if not isinstance(payload, dict) or payload.get("version") != 1:
+            return None
+        stored_url, authorization = payload.get("server_url"), payload.get("authorization")
+        if not isinstance(stored_url, str) or not isinstance(authorization, str):
+            return None
+        if _http_base_url(f"{stored_url.rstrip('/')}/mcp", allow_insecure_http=True) != _http_base_url(
+            server_url, allow_insecure_http=True
+        ):
+            return None
+        scheme, separator, credential = authorization.partition(" ")
+        if (
+            scheme.casefold() != "bearer"
+            or not separator
+            or not credential
+            or any(character.isspace() for character in credential)
+        ):
+            return None
+        else:
+            return authorization
+    except (OSError, TypeError, ValueError, json.JSONDecodeError):
+        return None
 
 
 def _http_base_url(mcp_url: str, *, allow_insecure_http: bool = False) -> str:
--- a/src/powercontext/cli/authorization.py
+++ b/src/powercontext/cli/authorization.py
@@ -28,8 +28,25 @@ from powercontext.client.settings import normalize_server_url
 
 AuthorizationStatus = Literal["configured", "not_configured", "url_mismatch", "invalid", "unsafe_permissions"]
 SetupAuthorizationStatus = Literal[
-    "configured", "preserved", "not_configured", "url_mismatch", "invalid", "unsafe_permissions"
+    "configured", "preserved", "cleared", "not_configured", "url_mismatch", "invalid", "unsafe_permissions"
 ]
+
+_AUTHORIZATION_ENVIRONMENTS = {
+    "codex": "POWERCONTEXT_CODEX_AUTHORIZATION",
+    "claude-code": "POWERCONTEXT_CLAUDE_AUTHORIZATION",
+    "opencode": "POWERCONTEXT_OPENCODE_AUTHORIZATION",
+    "pi": "POWERCONTEXT_PI_AUTHORIZATION",
+    "workbuddy": "POWERCONTEXT_WORKBUDDY_AUTHORIZATION",
+    "dsh": "POWERCONTEXT_DSH_AUTHORIZATION",
+}
+_SERVER_URL_ENVIRONMENTS = {
+    "codex": "POWERCONTEXT_CODEX_SERVER_URL",
+    "claude-code": "POWERCONTEXT_CLAUDE_SERVER_URL",
+    "opencode": "POWERCONTEXT_OPENCODE_BASE_URL",
+    "pi": "POWERCONTEXT_PI_BASE_URL",
+    "workbuddy": "POWERCONTEXT_WORKBUDDY_SERVER_URL",
+    "dsh": "POWERCONTEXT_DSH_BASE_URL",
+}
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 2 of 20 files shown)


src/powercontext/cli/system.py
```diff
@@ -941,11 +943,19 @@ def install_codex_plugin(*, source: str, ref: str) -> CodexSetupResult:
     marketplace_name = _required_string(marketplace, "marketplaceName")
 
     plugin = _run_codex_json("plugin", "add", f"{PLUGIN_NAME}@{marketplace_name}")
+    from powercontext.cli.authorization import configure_stored_authorization
+
+    authorization_state = configure_stored_authorization(
+        "codex",
+        server_url=DEFAULT_CLAUDE_CODE_SERVER_URL,
+        value=os.environ.get("POWERCONTEXT_CLIENT_API_TOKEN"),
```
Teingi: [P1] Wire saved credentials into native MCP

Saving this file only updates the hook credential source. Codex's `.mcp.json` still reads Authorization from the environment; Claude Code and WorkBuddy have the same gap. With a matching saved credential and no auth environment variable, the native Codex CLI sent `/mcp` without Authorization. Adding the environment override completed `initialize` and `tools/list`. Please connect native MCP to the same URL-bound credential source so setup actually removes the need for per-session exports.

integrations/codex/plugins/powercontext/settings.py
```diff
@@ -159,10 +167,36 @@ def normalize_scope_id(cls, value: str | None) -> str | None:
...
     return _http_base_url(configuration.mcp_servers["powercontext"].url)
 
 
+def _stored_authorization(server_url: str) -> str | None:
+    path = Path(os.environ.get("CODEX_HOME", Path.home() / ".codex")).expanduser() / "powercontext" / "credentials.json"
+    try:
+        if path.is_symlink() or not path.is_file() or stat.S_IMODE(path.stat().st_mode) & 0o077:
+            return None
+        payload = json.loads(path.read_text(encoding="utf-8"))
+        if payload.get("version") != 1 or _http_base_url(payload["server_url"]) != _http_base_url(server_url):
```
Teingi: [P1] Validate saved base URLs without requiring /mcp

`_http_base_url()` requires its input to end in `/mcp`, but both the saved URL and the `server_url` argument are already base URLs. A normal setup therefore raises ValueError here and silently discards the saved credential. I reproduced setup returning `configured` while `CodexPluginSettings.authorization` remained None. Please use base-URL normalization for this comparison and cover the setup-to-load path.

src/powercontext/cli/workbuddy.py
```diff
@@ -130,12 +131,19 @@ def install_workbuddy_plugin(*, source: str, ref: str) -> WorkBuddySetupResult:
...
+
     return WorkBuddySetupResult(
         plugin=WORKBUDDY_PLUGIN_NAME,
         plugin_path=str(plugin_dir),
         workbuddy_home=str(home),
         hooks_dir=str(hooks_dir),
         data_dir=str(data_dir),
+        authorization_state=configure_stored_authorization(
+            "workbuddy",
+            server_url="http://127.0.0.1:8000",
```
Teingi: [P2] Bind the credential to the configured server

This hardcodes port 8000 even when the host uses another endpoint. Running the WorkBuddy installer with `POWERCONTEXT_WORKBUDDY_SERVER_URL=http://127.0.0.1:18765` still returned `configured` and saved port 8000; the hook then rejected the credential because the URLs differed. DSH, OpenCode, and Pi also hardcode this value. Please resolve the effective host endpoint before persisting the credential.

integrations/workbuddy/plugins/powercontext/hooks/workbuddy_settings.py
```diff
@@ -151,6 +157,19 @@ def _authorization_header(value: str | None) -> str | None:
     return normalized
 
 
+def _stored_authorization(*, server_url: str, root: Path) -> str | None:
+    path = root / "powercontext" / "credentials.json"
+    try:
+        if path.is_symlink() or not path.is_file() or stat.S_IMODE(path.stat().st_mode) & 0o077:
```
Teingi: [P2] Restrict the POSIX permission check to POSIX

All six plugin loaders apply `mode & 0o077` unconditionally, although the shared reader explicitly excludes Windows. Windows emulates these mode bits without separate owner/group/other permissions, so this rejects normally saved credentials there. Please use platform-appropriate permission handling. This finding is based on static analysis, not a Windows run; see the [Python](https://docs.python.org/3/library/os.html#os.chmod) and [Node](https://nodejs.org/api/fs.html#fschmodpath-mode-callback) documentation.

src/powercontext/cli/authorization.py
```diff
@@ -0,0 +1,171 @@
...
+    if os.name != "nt" and stat.S_IMODE(path.stat().st_mode) & 0o077:
+        return AuthorizationResolution("unsafe_permissions", None)
+    try:
+        payload = json.loads(path.read_text(encoding="utf-8"))
+        if not isinstance(payload, dict) or payload.get("version") != 1:
+            raise ValueError("invalid persisted authorization record")  # noqa: TRY003, TRY301
+        stored_url = normalize_server_url(payload["server_url"])
+        authorization = normalize_authorization(payload["authorization"])
+        effective_url = normalize_server_url(server_url)
+    except (OSError, TypeError, ValueError, json.JSONDecodeError):
```
Teingi, line 147: [P2] Handle malformed credential records consistently

A file containing `{"version":1}` raises an uncaught KeyError when the required fields are read, so rerunning setup fails instead of reporting `invalid`. Separately, a file containing `[]` raises AttributeError in all three Python plugin loaders at `payload.get()`. Both cases were reproduced. Please validate the object and field types and return the documented invalid/no-credential result for malformed records.

src/powercontext/cli/authorization.py
```diff
@@ -0,0 +1,171 @@
...
+        return AuthorizationResolution("invalid", None)
+    if stored_url != effective_url:
+        return AuthorizationResolution("url_mismatch", None)
+    return AuthorizationResolution("configured", authorization)
+
+
+def clear_stored_authorization(path: Path) -> Literal["cleared", "not_configured"]:
```
Teingi, line 154: [P2] Expose credential clearing through setup

This helper has no production callers, and the setup commands do not expose `--clear-authorization`. Running `setup workbuddy --clear-authorization --json` returns `No such option`; rerunning without a token preserves the saved credential. Please wire the explicit clear option into the supported setup commands, as required by #1538, while preserving unrelated host configuration.

Teingi (approved):
LGTM

PR #1549, opened 2026-09-10
RFC: unify artifact family reads and registration

Zxf-xufeng:

This RFC scopes the change to exposing Topic Memory through the standard Artifact read interfaces required by the memory showcase page.

Implementation draft: #1550.

# Rationale for this change

Topic Memory is already persisted as an Artifact, but the standard `list_artifacts` route rejects `topic-memory`. The showcase therefore cannot use the standard Artifact read contract for its list page.

- Define standard read support for `topic-memory` in `list_artifacts`, `get_artifact`, `list_artifact_revisions`, and `get_artifact_revision`.
- Keep read-family validation separate from writable and tag-related Artifact family validation.
- Define Topic Memory list semantics: published items only, publication-time descending order, signed cursor pagination, and available display metadata.
- Reuse the existing standard Artifact response for detail and revision reads; keep Topic Memory-specific search, get, and flush semantics on their existing endpoints.
- Keep the current Artifact repository composition and family-specific management wiring unchanged.

This RFC does not introduce a general family registry, automatic registration framework, or a new generic capability model.

None in this PR: it contains design documentation only.

# How was this change tested?

- `git diff --check` passed for the documentation changes.
- No runtime behavior is changed by this PR.

# AI usage statement

Authored with OpenAI Codex assistance.

### DIFF
--- a/docs/en/rfcs/0000_artifact_family_unification.md
+++ b/docs/en/rfcs/0000_artifact_family_unification.md
@@ -1,0 +1,114 @@
+---
+title: Unified Artifact Family Reads and Registration
+---
+
+# Unified Artifact Family Reads and Registration
+
+- Proposal Name: `artifact_family_unification`
+- Start Date: 2026-09-10
+- RFC PR: TBD
+- Status: Proposed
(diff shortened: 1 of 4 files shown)


Teingi (approved):
LGTM

PR #1558, opened 2026-09-10
feat(dify): add memory tools and an automatic-memory agent strategy

thunguo:

NONE.
However, we first need to wait for Dify's PR https://github.com/langgenius/dify/pull/42152 to be merged.

# Rationale for this change

Dify applications need a way to recall PowerContext memory and retain useful execution evidence across runs. This change adds two independently installable plugins: memory tools for workflows and an Agent strategy that automatically recalls context and captures visible execution events.

- Add eight typed HTTP tools for context preparation, event capture, memory search and maintenance, and explicit extraction checkpoints.
- Add a Function Calling strategy for the legacy Workflow Agent, with private recall/capture callbacks and bounded context management.
- Route memory through provisioned bindings derived from deployment namespace, app, and user or business identity. Missing bindings do not fall back to a default Scope.
- Bound and redact captured events, use stable source identifiers, and return sanitized errors. Capture persists evidence; searchable memory depends on the configured extraction pipeline or an explicit checkpoint.
- Add SDK declarations, package metadata, privacy documentation, source packaging, localized integration documentation, and local CE acceptance scripts.

Users can configure a PowerContext tool credential and invoke memory tools in workflows. Legacy Workflow Agents can select the PowerContext Function Calling strategy for automatic memory callbacks.

Native Agent V2 automatic memory requires the companion Dify runtime/API/Web changes. Installing these plugins alone does not add that capability to an unmodified Dify release. Existing PowerContext HTTP contracts are unchanged.

The connection needs a reachable PC Server and provisioned Scope bindings. Binding identity selects data; deployments must enforce authorization separately. The plugin currently pins its PC client to an immutable source commit.

# How was this change tested?

- `make dify-test`: Ruff and ty passed; 11 plugin tests and 1 PC HTTP/SQLite integration test passed. Strategy tests use simulated model responses, and the extraction test uses a deterministic pipeline.
- Local CE deployment with PostgreSQL, Redis, a real plugin daemon, both SDK plugin processes, and PC Server.
- `live_workflow.py`: verified direct memory writes, recall, business Scope isolation, missing-binding behavior, and persisted Source redaction through actual workflow runs.
- `live_users.py`: verified cross-request recall for Alice and empty recall for Bob through the local published Workflow Service API.
- `live_native_config.py`: verified V2 configuration save/reload and credential references without exposing the PC token, using the companion Dify changes.

Real-model automatic Agent recall/capture, model-based extraction, long-context compaction, and pause/resume acceptance remain outstanding. Official `.difypkg` installation and Marketplace publication have not been performed. See `integrations/dify/local/VALIDATION.md` for the deployment evidence and its limits.


integrations/dify/powercontext_agent/strategies/function_calling.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def _execute_tool(
+        self, state: RunState, call: AssistantPromptMessage.ToolCall
+    ) -> Generator[AgentInvokeMessage, None, None]:
+        state.memory.observe(
+            "tool_call", {"tool": call.function.name, "arguments": call.function.arguments, "tool_call_id": call.id}
+        )
+        try:
+            result_messages = self._invoke_tool(state.tools, call)
+            result = "\n".join(_tool_text(item) for item in result_messages)
```
Teingi, line 184: [P1] Preserve structured tool results until capture redaction

Joining the messages here bypasses credential-field redaction when a tool returns text followed by a JSON message containing `api_key` (or multiple JSON messages). The resulting string is no longer a valid JSON document, so the bridge's sanitizer treats it as free text and preserves the secret. I reproduced this through the strategy and the real PC HTTP/SQLite app: capture returned `accepted`, and reading the Source back exposed the synthetic key. The single-JSON-message control was correctly redacted. Please pass structured messages to capture, or redact each message before joining, and cover mixed text/JSON and multiple-JSON results.

integrations/dify/powercontext/operation.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def _invoke(self, tool_parameters: dict[str, Any]) -> Generator[ToolInvokeMessage, None, None]:
+        try:
+            connection = Connection.model_validate(self.runtime.credentials)
+            identity_value = tool_parameters.get("memory_context")
+            if not identity_value:
+                identity_value = {"app_id": self.session.app_id, "subject_id": self.runtime.user_id}
+            identity = MemoryIdentity.model_validate(_object(identity_value))
+            request = _object(tool_parameters.get("request", {}))
+            result = asyncio.run(self._execute(connection, identity, request))
```
Teingi, line 45: [P2] Avoid asyncio.run in concurrent gevent workers

The pinned Dify SDK applies gevent monkey-patching, so its executor workers can share the native thread's running asyncio loop. With that SDK and a loopback HTTP server, two sequential tool invocations both succeeded, but two concurrent invocations caused one to raise `RuntimeError: asyncio.run() cannot be called from a running event loop`. It returned `error/invalid_response` without sending an HTTP request. This affects all operations using this adapter, so overlapping workflows can lose recall, writes or capture; automatic callbacks also suppress subsequent attempts for that run. Please use a gevent-compatible execution boundary and add a concurrent-invocation regression test.

integrations/dify/powercontext_agent/strategies/window.py
```diff
@@ -0,0 +1,113 @@
...
+            model_config=model,
+            stream=False,
+            prompt_messages=[
+                SystemPromptMessage(content=SUMMARY_INSTRUCTION),
+                UserPromptMessage(content=chunk),
+            ],
+        )
+        strategy.increase_usage(usage, result.usage)
+        summaries.append(result.message.get_text_content())
+    return "\n".join(summaries)
```
Teingi, line 113: [P2] Bound the combined summary, not just each chunk

Each chunk gets its own output allowance, but all summaries are concatenated without a total budget or a reduction pass. With 25 historical exchanges (~42 KB) and a 12,000-byte input budget, simulated summaries of only 990 ASCII characters each caused 19 model calls and over 18 KB of summary text. `fit_window` then raised `Compacted history still exceeds the model input budget`, aborting the agent even though the current task and retained tail fit. Please allocate a total summary budget from the remaining context space, or recursively reduce the summaries, and test multiple realistic-length chunk summaries.

Teingi (commented):
Three reproducible issues below. The existing 11 plugin tests, HTTP/SQLite test, Ruff and type checks passed. Additional probes used the pinned SDK, a loopback HTTP server, and the real PC HTTP app backed by SQLite. Model responses in the strategy and compaction probes were simulated; full CE/plugin-daemon/model acceptance was not run.

PR #1562, opened 2026-09-10
fix(pi): keep failure diagnostics off stderr by default

### LINKED ISSUE #1528: bug(pi): console.warn in diagnostics emitter corrupts TUI input bar
## Description

The Pi integration's diagnostic emitter writes JSON to stderr via `console.warn()`, which corrupts the TUI input bar rendering. The stderr output interleaves with the TUI's stdout-based cursor-positioned rendering, causing the diagnostic JSON to appear inline with the `❯` prompt and breaking the border.

```
── π ... ─────────────────────────────────────
{"component":"powercontext.pi","event":"flush_memory","outcome":"server_unavailable",...}
─────────────────────────────────────────
```

## Root cause

In `integrations/pi/plugins/powercontext/extensions/powercontext.ts`:

```typescript
const emitDiagnostic = createDiagnosticEmitter((line) => console.warn(line))
```

`console.warn()` writes to `process.stderr`, which goes directly to the terminal. Pi's TUI uses ANSI cursor positioning on stdout for incremental rendering. Any stderr output during rendering corrupts the display.

## The DSH integration does this correctly

In `integrations/dsh/plugins/powercontext/src/index.ts`:

```typescript
const emitDiagnostic = createDiagnosticEmitter((line) => ctx.logger.warn(line))
```

The DSH host provides a proper logger. The Pi `ExtensionContext` does not expose a `logger` property, so the Pi integration fell back to `console.warn()`.

## Proposed fix

Options (from least to most effort):

1. **Write diagnostics to a file** — e.g. `$PI_POWERCONTEXT_DIAGNOSTIC_LOG` or a fixed path like `~/.powercontext/pi-diagnostics.log`. Silent by default, opt-in via env var.
2. **Gate behind `PI_DEBUG`** — only emit to stderr when `PI_DEBUG=1` is set, so normal usage is clean.
3. **Provide a logging API in Pi's ExtensionContext** — add `ctx.log(level, message)` or `ctx.logger` that the host can route safely (to file, to a status widget, or suppressed in TUI mode). This would also benefit other extensions.

Option 1 or 2 is a quick fix for the powercontext side. Option 3 is the proper long-term fix on Pi's side.

## Environment

- powercontext pi plugin: current master
- Pi version: latest
- Ghostty
- Reproducible whenever PowerContext server returns 503 (triggers `flush_memory` → `server_unavailable` diagnostic)

L4XB:

Closes #1528.

# Rationale for this change

The Pi extension wrote its diagnostic JSON lines through `console.warn`, so every `server_unavailable` notice (for example after a 503 from `flush_memory`) landed inside Pi's TUI: Pi renders on stdout with cursor positioning, and stderr output interleaves with the `❯` input bar and breaks the border. Pi's `ExtensionContext` exposes no logger, unlike the DSH host the sibling plugin routes its diagnostics through, so the extension needs its own sink.

- `src/config.ts`: new `diagnostics` setting from `POWERCONTEXT_PI_DIAGNOSTICS` — `off` (default), `stderr`, or a file path (`~/` expanded like `POWERCONTEXT_CLIENT_CONFIG_FILE`).
- `src/diagnostics.ts`: `diagnosticWriter(sink)` returns the line sink for the emitter: silent, `console.warn`, or append-as-JSON-lines to the file; write failures are swallowed so diagnostics can never break the extension.
- `extensions/powercontext.ts`: the emitter uses `diagnosticWriter(config.diagnostics)` instead of `console.warn`.
- README: documents the setting and why the default is silent.
- Tests: `config.spec.ts` (parsing, default, `~/` expansion), `diagnostics.spec.ts` (all three sinks, write failure), `extension.spec.ts` (a new case asserts stderr stays clean by default; the four existing stderr assertions now opt in with `POWERCONTEXT_PI_DIAGNOSTICS=stderr`), plus the two hand-built config fixtures.

Yes: diagnostics are no longer printed by default. `POWERCONTEXT_PI_DIAGNOSTICS=stderr` restores the previous behaviour; `/pc` still shows the current status. No API or persisted-format changes.

# How was this change tested?

```
cd integrations/pi/plugins/powercontext
pnpm test        # 82 passed
pnpm typecheck   # clean
```

Not exercised against a live Pi TUI; the change only replaces the stderr sink.

# AI usage statement

Claude Code (Claude Fable 5.1) was used to analyse the issue, write the change and the tests; the author directed the work and takes responsibility for it.

### DIFF
--- a/integrations/pi/plugins/powercontext/README.md
+++ b/integrations/pi/plugins/powercontext/README.md
@@ -20,8 +20,9 @@ explicit override, and automatic prompt capture.
 
 Failure diagnostics (for example `server_unavailable` after a 503) are silent by default: Pi's TUI renders on
 stdout with cursor positioning, so anything written to stderr corrupts the input bar. Set
-`POWERCONTEXT_PI_DIAGNOSTICS=stderr` to print them, or point it at a file (`~/` is expanded) to append them as
-JSON lines. `/pc` shows the current status either way.
+`POWERCONTEXT_PI_DIAGNOSTICS=stderr` to print them, or point it at an absolute path (`~/` is expanded; the
+parent directory is created) to append them as JSON lines. The keywords are case-insensitive; any other value
+keeps diagnostics off. `/pc` shows the current status either way.
 
 Remote HTTP is rejected by default. Explicitly allow it with `POWERCONTEXT_PI_ALLOW_INSECURE_HTTP=true`,
 or use `POWERCONTEXT_CLIENT_ALLOW_INSECURE_HTTP=true` as the common fallback. A host flag of `false` overrides
--- a/integrations/pi/plugins/powercontext/README.md
+++ b/integrations/pi/plugins/powercontext/README.md
@@ -18,6 +18,11 @@ The package resolves an explicit Scope, a durable workspace binding, or the Serv
 `POWERCONTEXT_PI_BASE_URL`, `POWERCONTEXT_PI_SCOPE_ID`, and `POWERCONTEXT_PI_CAPTURE_PROMPTS` to adjust the connection,
 explicit override, and automatic prompt capture.
 
+Failure diagnostics (for example `server_unavailable` after a 503) are silent by default: Pi's TUI renders on
+stdout with cursor positioning, so anything written to stderr corrupts the input bar. Set
+`POWERCONTEXT_PI_DIAGNOSTICS=stderr` to print them, or point it at a file (`~/` is expanded) to append them as
+JSON lines. `/pc` shows the current status either way.
+
 Remote HTTP is rejected by default. Explicitly allow it with `POWERCONTEXT_PI_ALLOW_INSECURE_HTTP=true`,
 or use `POWERCONTEXT_CLIENT_ALLOW_INSECURE_HTTP=true` as the common fallback. A host flag of `false` overrides
 the common flag. Flags accept only `true/false`, `1/0`, `yes/no`, or `on/off`; HTTPS certificate validation and
--- a/integrations/pi/plugins/powercontext/src/config.ts
+++ b/integrations/pi/plugins/powercontext/src/config.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { homedir } from 'node:os'
+import { join } from 'node:path'
 import { resolveTransport } from './transport.ts'
 
 export interface ResolvedConfig {
@@ -28,6 +30,8 @@ export interface ResolvedConfig {
   maxBytes: number
   flushOnCapture: boolean
   flushMaxCalls: number
+  /** Where failure diagnostics go: `off`, `stderr`, or a file path to append JSON lines to. */
+  diagnostics: string
 }
 
 const DEFAULTS: ResolvedConfig = {
@@ -41,6 +45,7 @@ const DEFAULTS: ResolvedConfig = {
   maxBytes: 8000,
   flushOnCapture: false,
   flushMaxCalls: 4,
+  diagnostics: 'off',
 }
 
 function envString(env: NodeJS.ProcessEnv, name: string): string | undefined {
@@ -105,5 +110,13 @@ export function resolveConfig(env: NodeJS.ProcessEnv = process.env): ResolvedCon
     maxBytes: envInteger(env, 'POWERCONTEXT_PI_MAX_BYTES', DEFAULTS.maxBytes, 512, 32_768),
     flushOnCapture: envBoolean(env, 'POWERCONTEXT_PI_FLUSH_ON_CAPTURE') ?? DEFAULTS.flushOnCapture,
     flushMaxCalls: envInteger(env, 'POWERCONTEXT_PI_FLUSH_MAX_CALLS', DEFAULTS.flushMaxCalls, 1, 16),
+    diagnostics: diagnosticsSink(envString(env, 'POWERCONTEXT_PI_DIAGNOSTICS'), env),
   }
+}
+
+function diagnosticsSink(raw: string | undefined, env: NodeJS.ProcessEnv): string {
+  if (raw === undefined) return DEFAULTS.diagnostics
+  if (raw === 'off' || raw === 'stderr') return raw
+  if (raw.startsWith('~/')) return join(envString(env, 'HOME') ?? homedir(), raw.slice(2))
+  return raw
 }
--- a/integrations/pi/plugins/powercontext/src/config.ts
+++ b/integrations/pi/plugins/powercontext/src/config.ts
@@ -15,7 +15,7 @@
  */
 
 import { homedir } from 'node:os'
-import { join } from 'node:path'
+import { isAbsolute, join } from 'node:path'
 import { resolveTransport } from './transport.ts'
 
 export interface ResolvedConfig {
@@ -116,7 +116,12 @@ export function resolveConfig(env: NodeJS.ProcessEnv = process.env): ResolvedCon
 
 function diagnosticsSink(raw: string | undefined, env: NodeJS.ProcessEnv): string {
   if (raw === undefined) return DEFAULTS.diagnostics
-  if (raw === 'off' || raw === 'stderr') return raw
-  if (raw.startsWith('~/')) return join(envString(env, 'HOME') ?? homedir(), raw.slice(2))
-  return raw
+  const token = raw.trim().toLowerCase()
+  if (token === 'off' || token === 'stderr') return token
+  const path = raw.trim()
+  if (path.startsWith('~/')) return join(envString(env, 'HOME') ?? homedir(), path.slice(2))
+  // Only an unambiguous file path becomes a file sink. Anything else (a typo such as
+  // `STDER`, a bare relative name) stays silent instead of creating a stray file in cwd.
+  if (isAbsolute(path)) return path
+  return DEFAULTS.diagnostics
 }
--- a/integrations/pi/plugins/powercontext/src/diagnostics.ts
+++ b/integrations/pi/plugins/powercontext/src/diagnostics.ts
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import { appendFileSync } from 'node:fs'
+import { appendFileSync, mkdirSync } from 'node:fs'
+import { dirname } from 'node:path'
 import { InvalidResponseError, ServerResponseError, TransportError } from './errors.ts'
 
 export interface DiagnosticEvent {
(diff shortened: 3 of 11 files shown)


integrations/pi/plugins/powercontext/src/config.ts
```diff
@@ -105,5 +110,13 @@ export function resolveConfig(env: NodeJS.ProcessEnv = process.env): ResolvedCon
     maxBytes: envInteger(env, 'POWERCONTEXT_PI_MAX_BYTES', DEFAULTS.maxBytes, 512, 32_768),
     flushOnCapture: envBoolean(env, 'POWERCONTEXT_PI_FLUSH_ON_CAPTURE') ?? DEFAULTS.flushOnCapture,
     flushMaxCalls: envInteger(env, 'POWERCONTEXT_PI_FLUSH_MAX_CALLS', DEFAULTS.flushMaxCalls, 1, 16),
+    diagnostics: diagnosticsSink(envString(env, 'POWERCONTEXT_PI_DIAGNOSTICS'), env),
   }
 }
+
+function diagnosticsSink(raw: string | undefined, env: NodeJS.ProcessEnv): string {
+  if (raw === undefined) return DEFAULTS.diagnostics
+  if (raw === 'off' || raw === 'stderr') return raw
```
AlexStocks: [P2] Case variants fall through to the file branch and silently create a junk file in cwd

`off` / `stderr` are matched with exact string equality, while every other flag in the same README paragraph (`POWERCONTEXT_PI_CAPTURE_PROMPTS` and friends) is normalised through `envBoolean` with `toLowerCase()`, and the README explicitly lists `true/false`, `1/0`, `yes/no`, `on/off` as accepted spellings.

Verified on Head `7307e8e0` with a probe in an isolated mirror:

- `resolveConfig({ POWERCONTEXT_PI_DIAGNOSTICS: 'STDERR' }).diagnostics === 'STDERR'`
- `diagnosticWriter` therefore falls into the file branch and creates a file literally named `STDERR` in the Pi process cwd, appending JSON lines to it, while stderr stays silent.

Impact: a user who naturally writes `STDERR` or `Off` gets neither stderr output nor any error message, and additionally leaves a meaningless file in the session working directory — which is exactly the symptom this PR is meant to remove.

Suggestion: normalise the token before comparison (`raw.toLowerCase()`), or treat only absolute paths and `~/` prefixed paths as a file sink and explicitly reject or ignore any other unrecognised value. Tightening `ResolvedConfig.diagnostics` to a union such as `'off' | 'stderr' | { file: string }` would also catch this class of misrouting at compile time.

integrations/pi/plugins/powercontext/src/diagnostics.ts
```diff
@@ -73,6 +74,27 @@ export function failureEvent(event: string, error: unknown): DiagnosticEvent | u
...
+export function diagnosticWriter(
+  sink: string,
+  append: (path: string, line: string) => void = (path, line) => appendFileSync(path, line),
+  warn: (line: string) => void = (line) => console.warn(line),
+): (line: string) => void {
+  if (sink === 'off') return () => {}
+  if (sink === 'stderr') return warn
+  return (line) => {
+    try {
+      append(sink, `${line}\n`)
```
AlexStocks, line 109: [P2] The file sink fails completely silently when the parent directory is missing

Swallowing every write error in the catch below is an intentional design choice and I am not arguing against it; the problem is that the file branch neither creates the parent directory nor leaves any observable trace when a write fails.

Verified on Head: with the sink pointing at `<tmp>/missing-dir/diag.jsonl`, calling the writer throws nothing, creates no directory, creates no file, and writes nothing to stderr.

Impact: issue #1528 itself suggests `~/.powercontext/pi-diagnostics.log`, and `~/.powercontext` does not exist on most machines. Anyone following that advice ends up with zero lines written and zero feedback, and will conclude that the server simply had no failures — which defeats the purpose of having a diagnostics sink.

Suggestion: run `mkdirSync(dirname(sink), { recursive: true })` once when the file sink is constructed (its failure can still be swallowed), or surface the first write failure through the `/pc` status. A README note that the parent directory must exist would also help.

integrations/pi/plugins/powercontext/README.md
```diff
@@ -18,6 +18,11 @@ The package resolves an explicit Scope, a durable workspace binding, or the Serv
 `POWERCONTEXT_PI_BASE_URL`, `POWERCONTEXT_PI_SCOPE_ID`, and `POWERCONTEXT_PI_CAPTURE_PROMPTS` to adjust the connection,
 explicit override, and automatic prompt capture.
 
+Failure diagnostics (for example `server_unavailable` after a 503) are silent by default: Pi's TUI renders on
```
AlexStocks, line 21: [P2] The website docs environment-variable tables are not updated

`docs/en/docs/integrations/pi.md` and `docs/zh/docs/integrations/pi.md` each carry an Environment variables / 环境变量 table that lists every other `POWERCONTEXT_PI_*` variable together with its default. The new `POWERCONTEXT_PI_DIAGNOSTICS` (default `off`) and the user-facing behaviour change "diagnostics are silent by default" appear in neither; only the plugin's own README was updated. CONTRIBUTING.md asks for the docs to be updated when functionality is added.

Suggestion: add the row to both tables, and add a sentence after them noting that diagnostics are no longer written to stderr unless explicitly enabled.

integrations/pi/plugins/powercontext/tests/config.spec.ts
```diff
@@ -38,8 +38,16 @@ describe('Pi configuration', () => {
...
       flushMaxCalls: 4,
+      diagnostics: 'off',
     })
   })
+
+  it('keeps diagnostics silent unless a sink is configured', () => {
+    expect(resolveConfig({}).diagnostics).toBe('off')
+    expect(resolveConfig({ POWERCONTEXT_PI_DIAGNOSTICS: 'stderr' }).diagnostics).toBe('stderr')
+    expect(resolveConfig({ POWERCONTEXT_PI_DIAGNOSTICS: '/tmp/pc.log' }).diagnostics).toBe('/tmp/pc.log')
+    expect(resolveConfig({ POWERCONTEXT_PI_DIAGNOSTICS: '~/pc.log', HOME: '/home/demo' }).diagnostics).toBe('/home/demo/pc.log')
```
AlexStocks: [P2] This assertion assumes POSIX path joining and fails on Windows

`diagnosticsSink` uses `node:path.join`, whose separator is platform dependent. On Windows the expression evaluates to `\home\demo\pc.log`, which is not equal to the `/home/demo/pc.log` asserted on this line, so `pnpm test` goes red.

Everywhere else in this suite path expectations are built with `join(...)` rather than hard-coded POSIX literals (see `tests/transport-consent.spec.ts`); there is no precedent for this style. CI only runs ubuntu-latest, so it will not catch it.

Suggestion: `expect(resolveConfig({ POWERCONTEXT_PI_DIAGNOSTICS: '~/pc.log', HOME: '/home/demo' }).diagnostics).toBe(join('/home/demo', 'pc.log'))`.

integrations/pi/plugins/powercontext/tests/diagnostics.spec.ts
```diff
@@ -82,3 +82,34 @@ describe('host-visible diagnostic classification', () => {
...
+    diagnosticWriter('stderr', append, warn)('{"event":"flush_memory"}')
+    expect(warn).toHaveBeenCalledWith('{"event":"flush_memory"}')
+    expect(append).not.toHaveBeenCalled()
+  })
+
+  it('appends JSON lines to a configured file and survives write failures', () => {
+    const append = vi.fn()
+    const warn = vi.fn()
+    diagnosticWriter('/var/log/pc.jsonl', append, warn)('{"event":"flush_memory"}')
+    expect(append).toHaveBeenCalledWith('/var/log/pc.jsonl', '{"event":"flush_memory"}\n')
```
AlexStocks, line 112: [P2] The file sink is only tested through injected mocks; the real `appendFileSync` path is uncovered

All three cases inject `append` / `warn` into `diagnosticWriter`, so the branch coverage is fine; but the default parameter `appendFileSync` is never executed, and the `\n` delimiter contract is only asserted against a mock.

The implementation itself is fine — I drove the whole chain through a real filesystem in an isolated mirror (env var -> extension entry -> file content is a single parseable JSON line plus one `\n`). What is missing is regression protection: if someone later changes the delimiter, or adds parent-directory creation for the file sink, none of these tests would notice.

Suggestion: add a case that points the sink at a real path under `mkdtempSync`, writes twice, reads the file back, splits on `\n` and asserts the content. That would also cover the actual on-disk result of `~/` expansion.

integrations/pi/plugins/powercontext/src/config.ts
L4XB: Done: the token is trimmed and lower-cased before the keyword comparison, and only an absolute path or a `~/` path is treated as a file sink; anything else (`STDER`, `pc.log`, `./logs/pc.log`) resolves to `off`, so no stray file can appear in the session cwd. I kept `diagnostics: string` rather than the union to avoid touching the extension entry and fixtures for this PR; tests cover `STDERR`/`Off`/` stderr ` and the non-path values.

integrations/pi/plugins/powercontext/src/diagnostics.ts
L4XB: Done: the file sink now runs `mkdirSync(dirname(sink), { recursive: true })` once when it is constructed (failure swallowed, like the writes), so `~/.powercontext/pi-diagnostics.jsonl` works on a fresh machine. The README says the parent directory is created.

integrations/pi/plugins/powercontext/README.md
L4XB: Done: `POWERCONTEXT_PI_DIAGNOSTICS` (default `off`) is in both environment tables (`docs/en` and `docs/zh`), each followed by a sentence that diagnostics are silent by default and why.

integrations/pi/plugins/powercontext/tests/config.spec.ts
L4XB: Done: the expectation is `join('/home/demo', 'pc.log')` now.

integrations/pi/plugins/powercontext/tests/diagnostics.spec.ts
L4XB: Done: added a case that points the sink at `<mkdtemp>/nested/pi-diagnostics.jsonl` with the real `appendFileSync`, writes two events, reads the file back and asserts two parseable JSON lines plus the trailing newline (it also covers the parent-directory creation). Copilot's note about the stderr writer is addressed in the same commit: `warn(line)` is wrapped like the file sink, with a test for a throwing writer.

Teingi (approved):
LGTM

PR #1563, opened 2026-09-10
feat(prompts): add Topic Memory prompt management

### LINKED ISSUE #1532: feat: support custom prompts for Topic Memory generation stages
## Problem

Scope-level Prompt customization currently exposes Prompt keys for Memory, Experience, Skill, and Handoff, but not Topic Memory. The Topic Memory pipeline has six generation stages whose instructions remain code-defined and cannot be overridden through the Prompt API:

- probe
- global
- planner
- evolve
- temporary
- reconcile

This makes it impossible to apply a scope-level language policy consistently. For example, a Chinese Source can produce English Topic titles/summaries depending on the generation model, while other configurable artifacts can receive a custom instruction to follow the Source language.

## Expected behavior

Expose Topic Memory prompt definitions through the same scope-level custom Prompt mechanism, or provide an equivalent supported configuration. The effective instruction should be able to include a rule such as:

> The output language should follow the dominant language of the source content. If the source is primarily Chinese, write the natural-language output in Chinese; preserve code, identifiers, paths, commands, and proper nouns as supplied.

The six stages should retain their existing evidence, schema, citation, and safety constraints when customized.

## Acceptance criteria

- Topic Memory prompt keys are discoverable through the Prompt configuration API.
- A scope can create/replace custom instructions without replacing the built-in safety/schema rules unintentionally.
- The effective prompt for every Topic Memory stage is observable and versioned.
- A Chinese Source produces Chinese Topic title, summary, and detail in a new processing run.
- Existing auto-mode behavior remains unchanged when no custom Prompt is configured.

frf12:

Closes #1532.

# Rationale for this change

Topic Memory generation needs Scope-owned prompt configuration so each Scope can select the built-in prompt or provide a validated custom prompt without changing server-owned schemas or crossing Scope boundaries.

- Add Auto/custom management for the seven Topic Memory prompt keys.
- Resolve the effective prompt for Topic Memory processing and bind prompt metadata to generated artifacts.
- Update the OpenAPI schema and generated API models for prompt configuration.
- Extend prompt validation and HTTP API integration coverage.

The HTTP API exposes Scope-level Topic Memory prompt configuration and reports the effective built-in or custom prompt. No Dashboard or Profile changes are included.

# How was this change tested?

- `uv run --no-sync pytest tests/builtin/artifacts/prompt/test_prompt_validation.py -q`: 29 passed.
- `uv run --no-sync pytest tests/e2e/test_prompt_management_api.py -q`: 6 passed.
- `git diff --check`: passed.

# AI usage statement

OpenAI Codex assisted with implementation, tests and API verification.


src/powercontext/builtin/artifacts/prompt/builtin.py
```diff
@@ -171,3 +198,49 @@ def builtin_prompt_definitions(
...
+        ),
+    )
+    return tuple(
+        PromptDefinition(
+            key=cast(PromptKey, f"topic_memory.{name}"),
+            definition_version=f"powercontext.prompt.topic_memory.{name}.v1",
+            input_type=input_type,
+            output_type=output_type,
+            builtin_version=f"powercontext.topic_memory.{name}.v1",
+            invariant_instructions=_COMMON_INVARIANTS
```
Teingi: [P2] Preserve the stage-specific rules in custom mode

`PromptDefinition.resolve()` compiles custom prompts from `invariant_instructions` and the saved guidance, without `default_instructions`. Setting only a language preference therefore removes the Planner's complete-partition/candidate-grouping rules and the Reducer's exact evidence-union and full-coverage rules. These relationships are not expressed by the JSON schemas, but the runtime still rejects outputs that violate them. Please retain these requirements in each stage's invariant instructions so a language-only customization does not remove the operation contract.

src/powercontext/builtin/runtime/composition.py
```diff
@@ -332,6 +332,11 @@ async def open_builtin_runtime(
             ("experience.generate", experience_generator, generated_experience),
             ("skill.generate", skill_generator, generated_skill),
             ("handoff.generate", handoff_pipeline, generated_handoff),
+            *(
+                (f"topic_memory.{stage}", None, object())
```
Teingi, line 336: [P2] Register demonstration generators for the new prompt keys

These entries advertise all seven Topic Memory keys as `supported`, but `_register_prompt_demonstrators()` still registers only the existing generation keys. With a configured generation model, I reproduced `422 prompt_customization_unavailable` from every new `/prompts/{key}/demonstrations` endpoint, despite the configuration endpoint reporting `supported`. Please register the seven keys with the demonstration generator as well; adding them to the capabilities and OpenAPI enums alone leaves the new endpoints unusable.

src/powercontext/builtin/artifacts/prompt/definitions.py
```diff
@@ -73,11 +73,14 @@ def validate(self, content: PromptContent, /, *, during_inference: bool = False)
...
+                # Topic stage contracts use tuple fields for runtime immutability;
+                # JSON demonstrations naturally decode arrays into those tuples.
+                strict = not self.key.startswith("topic_memory.")
                 value = self.input_type.model_validate_json(
-                    json.dumps(demonstration.input), strict=True, extra="forbid"
+                    json.dumps(demonstration.input), strict=strict, extra="forbid"
                 )
                 output = self.output_type.model_validate_json(
-                    json.dumps(demonstration.expected_output), strict=True, extra="forbid"
+                    json.dumps(demonstration.expected_output), strict=strict, extra="forbid"
```
Teingi, line 83: [P2] Validate relationships within Topic Memory demonstrations

`validate_demonstration()` has no Topic Memory branches, so validating the input and output separately accepts impossible examples. Through the SQLite-backed HTTP API, all seven keys accepted invalid demonstrations with `201` and returned them in the effective configuration. Examples include citing evidence/probe IDs absent from the input and returning `covered_indices=[99]` for a single-input reduction. These examples are then compiled into subsequent prompts even though the processing pipeline rejects equivalent outputs. Please validate evidence references, planner assignments, and reduction/reconciliation relationships before saving or resolving these demonstrations.

src/powercontext/builtin/runtime/topic_memory_processing.py
```diff
@@ -1556,16 +1573,20 @@ def stage(
...
             raw = PydanticAIStructuredGenerator(
                 model=model,
-                instructions=instructions,
+                instructions=selected_instructions,
                 input_type=input_type,
                 output_type=output_type,
                 limits=limits,
                 model_settings=settings,
                 name=name,
+                prompt_key=prompt_key,
```
Teingi, line 1610: [P2] Persist the exact custom Prompt references in Topic lineage

Passing `prompt_key` supplies inference trace metadata, but the selected Prompt references never reach `TopicMemoryDraft.artifacts`. In a real spawned-worker test with SQLite and a loopback model stub, two Scopes used their respective custom Probe/Global instructions, yet both published Topics had `lineage.artifacts=[]`, including after reopening the database. Once a Prompt head changes, the Topic cannot identify the revision used to generate it. Please carry the exact custom Prompt references from the stages actually used into the published Artifact lineage, preserving the distinction between configuration inputs and Source evidence.

frf12:
Thanks @Teingi — all four findings are accepted and fixed in 3defaa049b22a2757d73c541de13de1a127c59ce:

1. **Custom mode losing stage rules** — each Topic Memory stage's schema-inexpressible operation contract (planner complete-partition/candidate-grouping, reducer exact evidence-union/full `covered_indices` coverage, reconcile identity preservation, probe/temporary evidence grounding, evolve historical-target rule) now lives in the stage's `invariant_instructions`, so a language-only customization keeps the operation contract.
2. **Demonstration generators** — `_register_prompt_demonstrators()` now registers all seven `topic_memory.*` keys; the previously reproduced `422 prompt_customization_unavailable` is gone (covered by an e2e test that generates `topic_memory.probe` demonstrations through the HTTP API).
3. **Demonstration relationship validation** — `validate_demonstration()` gained per-stage Topic Memory branches: evidence/probe references must exist in the input, planner outputs must partition every probe with exact candidate grouping, reduction must cover every zero-based position with the exact evidence union and the matching output kind, and reconciliation must reuse input `proposal_id`s without merging historical identities. The reviewer's `covered_indices=[99]` example is now a regression case.
4. **Prompt references in Topic lineage** — `TopicMemoryProcessor` now receives the exact custom Prompt refs selected at compose time, records which stages actually ran, and appends those refs to the published `TopicMemoryDraft.artifacts` (after the revised-topic ref), with a real-pipeline e2e asserting the lineage (and that unused stages' refs do not appear).

src/powercontext/builtin/artifacts/prompt/builtin.py
frf12: Accepted — fixed in 3defaa04.

Each Topic Memory stage's schema-inexpressible operation contract now lives in its `invariant_instructions` (planner complete-partition/candidate-grouping, reducer exact evidence-union/full `covered_indices` coverage, reconcile historical-identity preservation, probe/temporary evidence grounding, evolve historical-target rule). A language-only customization now keeps these requirements; Auto mode is unchanged. Regression: `test_custom_prompts_retain_stage_operation_contracts`.

src/powercontext/builtin/runtime/composition.py
frf12: Accepted — fixed in 3defaa04.

`_register_prompt_demonstrators()` now registers all seven `topic_memory.*` keys alongside the existing ones. The e2e test generates `topic_memory.probe` demonstrations through the HTTP API with a configured generation model; the reproduced `422 prompt_customization_unavailable` is gone.

src/powercontext/builtin/artifacts/prompt/definitions.py
frf12: Accepted — fixed in 3defaa04 (plus a ty narrowing fix in d2a2efb3).

`validate_demonstration()` gained per-stage Topic Memory branches: cited evidence/probe/proposal IDs must exist in the input, planner outputs must partition every probe with exact candidate grouping, reduction must cover every zero-based input position with the exact evidence union and the matching output kind, and reconciliation must reuse input `proposal_id`s without merging historical identities. Your `covered_indices=[99]` example is now a regression case, along with unknown-reference, out-of-range, wrong-kind, and identity-merge cases.

src/powercontext/builtin/runtime/topic_memory_processing.py
frf12: Accepted — fixed in 3defaa04.

`TopicMemoryProcessor` now receives the exact custom Prompt refs selected at compose time (`selection.artifact` per stage), records which stages actually ran in this window (via the existing reserve wrapper), and appends those refs to the published `TopicMemoryDraft.artifacts` — after the revised-topic ref, so configuration inputs stay distinguishable from Source evidence. A real SQLite pipeline e2e asserts the published lineage contains the used Probe ref and does not contain a stage that never ran. The upstream revision must exist (lineage FK), which the test honors by creating the prompt artifact first.

src/powercontext/builtin/artifacts/prompt/definitions.py
Teingi: Rechecked on `d2a2efb3`. The original invalid-reference and `covered_indices=[99]` cases are now rejected, but two P2 validator/runtime mismatches remain:

1. **Planner — `validation.py:200–208`: shared candidates can still be split across create items.** If two probes both list candidate `c1`, an output containing two separate items with no `candidate_id` passes demonstration validation and can be saved through the HTTP API (`201`). The same grouping fails in the SQLite processing pipeline with `invalid_plan`, leaving the cursor unchanged. The `candidate_id is None` shortcut skips this case, while the runtime checks grouping for every shared candidate. Please enforce that rule regardless of whether an item selects a target, and keep the Planner invariants aligned with it.

2. **Reconcile — `validation.py:260`: a valid new binding to supplied history is rejected.** If input proposal `p1` has no target and `historical` supplies `c1`, returning `p1` with `candidate_id=c1` is a valid reconciliation. The SQLite pipeline successfully publishes revision 2 and advances the cursor, but the same demonstration returns `422 prompt_definition_incompatible`. Requiring output targets to be a subset of `input_targets` excludes this supported update path. Please allow targets supplied in `value.historical`, while preserving existing proposal/target assignments and rejecting duplicate targets.

Both cases were verified through the Prompt HTTP API and the full SQLite processing pipeline with controlled stage outputs. The seven demonstration endpoints now return `200`, and the exact used Prompt references survive spawned-worker execution and database reopening.

frf12: Both confirmed and fixed in be4cf0b0 — thank you for the precise repro cases.

1. **Planner grouping**: the validator now mirrors `_plan`'s `item_by_probe` check — for every candidate listed by any probe, all probes sharing it must land in the same work item, regardless of whether any item selects it as its target. Selected targets must also be unique across items (the runtime's `targets` uniqueness check). Your split-create case is now rejected (`test_planner_demonstrations_group_shared_candidates` covers it and the single-item pass), and the Planner invariant text now states the shared-candidate rule explicitly.

2. **Reconcile binding**: `allowed_targets` is now `input_targets | supplied historical slots`, matching `_validate_reconciliation`'s `allowed_targets` from the component's history. An unbound proposal binding to a supplied historical candidate is accepted (`test_reconcile_demonstrations_may_bind_unbound_proposals_to_supplied_history`), while changing an existing assignment, duplicate targets, unknown proposal refs, and evidence outside the input union stay rejected. The Reconcile invariant text now mentions the binding path.

Verified locally: prompt suites 73 + 79 passed (incl. HTTP e2e), CI-exact `uv run --locked ty check` on the pypi index passes, ruff check/format clean.

Teingi (approved):
LGTM

PR #1566, opened 2026-09-11
docs: correct context management workflows

Zxf-xufeng:
## Related issue

None (documentation calibration).

## Summary

This PR aligns the context-management workflows with the current implementation and adds verification guidance for the areas most likely to cause incorrect data, permissions, or Scope isolation.

### Documentation changes

- Make the Codex workflow create and bind a dedicated Scope, document the prompt-capture boundary of the hook, and remove the unsupported Workstream selector step.
- Harden context-text export with `pipefail`, temporary files, response validation, and atomic replacement so failed requests cannot overwrite the previous context file.
- Correct the OpenDAL defaults and examples, including the server address, asynchronous Python entry point, bearer-token handling, and file-size limit.
- Clarify managed Skill package contents, external Skill target scan/resolve behavior, host-local paths, and remote receiver requirements.
- Add end-to-end vector-search acceptance checks, including the returned mode, `entry_id`, and `matched_by` fields, plus guidance for first-time embedding enablement.
- Document the no-model Experience path and the typed proposal option.
- Add Prompt CRUD, capability checks, ETag/If-Match handling, and correct Prompt permissions.

## User-facing impact

Documentation only. No runtime code or API behavior changes are included. The examples now require checking actual responses, permissions, Scope isolation, and failure handling.

## Validation

- `git diff --check`
- `.venv/bin/python /tmp/powercontext_workflow_audit.py static` (76 fenced Python/JSON blocks checked)
- Relevant tests: **68 passed, 1 deselected** (after removing inherited `POWERCONTEXT_*` environment variables)
- `make docs-test` was attempted, but the website build was blocked by pnpm dependency downloads in the current network/non-TTY environment.
- Prompt API smoke test confirmed that an unconfigured customization provider returns `prompt_customization_unavailable`; the prerequisite is now documented.

## AI usage

Codex (GPT-5) assisted with the documentation review, implementation cross-checks, example validation, and edits. The final content was reviewed by the submitter.


docs/zh/docs/workflows/prepare-context-text.md
```diff
@@ -29,15 +29,25 @@ Scope；读取其引用的其他 Scope 也需要对应权限。
...
 ```bash
+set -o pipefail
+tmp_context="$(mktemp "${TMPDIR:-/tmp}/powercontext-context.XXXXXX")"
+trap 'rm -f "$tmp_context"' EXIT
 curl --fail-with-body -sS http://127.0.0.1:8000/v1/context/prepare \
   -H 'Content-Type: application/json' --data-binary @prepare.json \
-  | jq -j '.content // empty' > context.md
+  | jq -er 'if .status == "empty" then "" elif .status == "ready" and (.content | type) == "string" then .content else error("unexpected prepare response") end' \
+  > "$tmp_context"
+mv "$tmp_context" context.md
```
Teingi, line 43: [P2] Only replace the context file after a successful pipeline

`pipefail` changes the pipeline exit status but does not stop Bash from running the following `mv`. I ran this block against an endpoint returning HTTP 500: the existing `context.md` was replaced with an empty file, and the script exited with status 0. Please make the replacement conditional on the request and response validation both succeeding, and propagate failures without replacing the previous file.

docs/zh/docs/workflows/architecture.md
```diff
@@ -0,0 +1,97 @@
...
+Scope；其中的箭头表示内容形成的路径。
+
+### 内容如何形成
+
+![一个 Scope 内的两类制品形成路径：从 Source 处理，或显式创建和更新；需要审核时经过 Candidate](/docs-diagrams/context-formation-zh.svg)
+
+Source 保存证据。处理证据与提交制品是不同的动作：采集完成不表示制品已经生成，生成完成也不一定表示内容已经获批。
+制品还可以通过支持该操作的类型接口显式创建或更新，不必先经过自动提取。
+
+- Experience、Skill 的提案通过 Candidate 审核后提交；显式操作也必须遵循各自的审核规则。
```
Teingi: [P2] Distinguish direct writes from Candidate review

This wording suggests that explicit Experience and Skill writes also require review. On this head, I verified that the generic Artifact Create endpoint returns 201 for both families with no pending Candidates: the Experience is immediately recalled into PreparedContext, and the Skill package is immediately available to read. Please distinguish the reviewed proposal/generation flow from direct Create/Replace commits here and in the English page, so callers do not rely on a review gate that these endpoints do not enforce.

Zxf-xufeng: Addressed in commit 943068a5.
The architecture documentation now distinguishes generated proposals from direct Artifact writes. Generated Experience and Skill proposals go through Candidate review before commit, while direct Create/Replace requests can commit a validated payload as a formal Revision immediately and do not create a Candidate automatically. The lifecycle descriptions and examples were updated in both Chinese and English to reflect this behavior.

Teingi (approved):
LGTM

PR #1569, opened 2026-09-11
feat(openclaw): add pc commands and work coordination workflows

XnLemon:

No issue or RFC linked. This PR addresses the identified OpenClaw integration gap.

Reference #1338

# Rationale for this change

OpenClaw can provide memory context, but it previously lacked an explicit PowerContext command surface and work-coordination workflows. This change makes PowerContext discoverable from the WebUI and enables durable Work Contract and Handoff operations.

- Add the `/pc` OpenClaw command:
  - `/pc`
  - `/pc help`
  - `/pc status`
  - `/pc handoff`
  - `/pc contract`
  - `/pc outcome`
- Add Work Contract creation.
- Add Handoff preparation, commit, continuation, and acknowledgement.
- Add Task Outcome recording.
- Reuse OpenClaw private-session checks and PowerContext scope resolution.
- Register new tools in the plugin manifest and OpenClaw allowlist configuration.
- Add tests and update OpenClaw integration documentation.
- No OpenAPI contract or persisted format changes.

Yes.

Users can access PowerContext through `/pc` in OpenClaw WebUI and use explicit Work Contract, Handoff, and Task Outcome workflows.

The new workflows are limited to authorized private sessions. Handoff and Work Contract content remains untrusted input/history. No breaking changes to existing Memory tools or public API contracts are introduced.

# How was this change tested?

- `pnpm --dir integrations/openclaw/plugins/memory-powercontext test`
- `pnpm --dir integrations/openclaw/plugins/memory-powercontext run typecheck`
- `pnpm --dir integrations/openclaw/plugins/memory-powercontext build`
- `uv run pytest tests/test_openclaw_configure_script.py tests/test_openclaw_cli.py`
- `make check`
- Manual validation with a local OpenClaw Gateway:
  - Plugin loaded successfully.
  - `/pc` command returned PowerContext help.
  - PowerContext Server started successfully after processing schema initialization.
  - `/health/ready` returned `{"status":"ready"}`.

# AI usage statement

Built with OpenAI Codex using GPT-5.6.

### DIFF
--- a/integrations/openclaw/plugins/memory-powercontext/src/work.ts
+++ b/integrations/openclaw/plugins/memory-powercontext/src/work.ts
@@ -1,0 +1,323 @@
+/*
+ * Copyright (c) 2026 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { createHash } from "node:crypto";
+import {
+  asToolParamsRecord,
+  jsonResult,
+  readStringParam,
+} from "openclaw/plugin-sdk/memory-core-host-runtime-core";
+import { Type } from "typebox";
+import type { OpenClawPluginToolContext } from "openclaw/plugin-sdk/plugin-entry";
+import type { PowerContextClient } from "./http.js";
+import { PowerContextRequestError } from "./http.js";
+import { resolveToolScope, type ToolDependencies } from "./tools.js";
+
+export const POWERCONTEXT_WORK_CONTRACT_TOOL = "powercontext_work_contract_create";
+export const POWERCONTEXT_HANDOFF_CURRENT_WORK_TOOL = "powercontext_handoff_current_work";
+export const POWERCONTEXT_HANDOFF_COMMIT_TOOL = "powercontext_handoff_commit";
+export const POWERCONTEXT_HANDOFF_CONTINUE_TOOL = "powercontext_handoff_continue";
+export const POWERCONTEXT_HANDOFF_ACKNOWLEDGE_TOOL = "powercontext_handoff_acknowledge";
+export const POWERCONTEXT_TASK_OUTCOME_TOOL = "powercontext_task_outcome";
+
+type JsonObject = Record<string, unknown>;
+
+const jsonObject = Type.Record(Type.String(), Type.Unknown());
+
+function readJsonObject(raw: Record<string, unknown>, name: string): JsonObject {
+  const value = raw[name];
+  if (!value || typeof value !== "object" || Array.isArray(value)) {
+    throw new Error(`${name} must be a JSON object`);
+  }
+  return value as JsonObject;
+}
+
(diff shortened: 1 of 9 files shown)


integrations/openclaw/plugins/memory-powercontext/src/work.ts
```diff
@@ -0,0 +1,323 @@
...
+export const POWERCONTEXT_WORK_CONTRACT_TOOL = "powercontext_work_contract_create";
+export const POWERCONTEXT_HANDOFF_CURRENT_WORK_TOOL = "powercontext_handoff_current_work";
+export const POWERCONTEXT_HANDOFF_COMMIT_TOOL = "powercontext_handoff_commit";
+export const POWERCONTEXT_HANDOFF_CONTINUE_TOOL = "powercontext_handoff_continue";
+export const POWERCONTEXT_HANDOFF_ACKNOWLEDGE_TOOL = "powercontext_handoff_acknowledge";
+export const POWERCONTEXT_TASK_OUTCOME_TOOL = "powercontext_task_outcome";
+
+type JsonObject = Record<string, unknown>;
+
+const jsonObject = Type.Record(Type.String(), Type.Unknown());
```
Teingi: [P2] Expose the required workflow payload fields in the tool schemas

`contract`, `handoff`, and `outcome` are exposed as arbitrary JSON objects, but the server requires specific schema/trust values, required fields, and nested claim structures. Neither the tool descriptions nor the `/pc` guidance explains these requirements, leaving the agent to guess the payload format.

I reproduced this on both OpenClaw `2026.8.1-beta.2` and `2026.9.1-beta.1` using the real plugin loader and a SQLite-backed PowerContext server. The three request examples in `work.test.ts` all return `invalid_request`; the agent receives only "The request violates the API contract." without the field-level validation details. For example, the prepare-current test sends `powercontext.prepared-handoff.v1`, while that endpoint expects `powercontext.current-work-handoff.v1`. Manually supplying valid payloads completes the full workflow on both versions.

Please expose the required fields/enums in the tool schemas, or accept simpler typed arguments and construct the API payloads in the plugin. A server-backed acceptance test would also catch the invalid examples that the current HTTP mocks accept.

integrations/openclaw/plugins/memory-powercontext/src/work.ts
```diff
@@ -0,0 +1,473 @@
...
+        const raw = asToolParamsRecord(params);
+        const selection = readStringParam(raw, "selection", { required: true });
+        if (selection !== "prepared" && selection !== "exact" && selection !== "latest") {
+          throw new Error("selection must be prepared, exact, or latest");
+        }
+        const body: JsonObject = {
+          scope_id: await resolveToolScope(ctx, deps, signal),
+          selection,
+        };
+        if (raw.prepared !== undefined) body.prepared = readJsonObject(raw, "prepared");
```
Teingi: [P2] Handle the null values allowed by the Handoff tool schemas

The schemas now allow `null` for `prepared`, `revision`, and `receiver_checks`, but the execution paths only exclude `undefined` before calling `readJsonObject()`. For example, `{"selection":"latest","prepared":null,"revision":null}` passes the tool schema, then returns `unavailable` with `prepared must be a JSON object`. The acknowledgement tool has the same issue, including a valid declined receipt with `receiver_checks: null`.

I reproduced five such cases on both OpenClaw `2026.8.1-beta.2` and `2026.9.1-beta.1` with the current plugin build and a real SQLite-backed server. Every case passed schema validation and failed through the plugin, while the equivalent direct API request returned HTTP 200. Omitting the unused fields allows the workflow to complete.

Please preserve `null` or treat it as an omitted optional field in both tools, and cover these cases with a regression test. Otherwise valid continuation and acknowledgement calls fail and incorrectly tell the agent to check the endpoint and credentials.

Teingi (approved):
LGTM

Issue #1571, opened 2026-09-12
bug: flaky macOS native service lifecycle test


PsiACE:
## Describe the bug

`test_native_personal_service_lifecycle` intermittently fails during service installation in the `macos-launchagent` CI job. Rerunning the same commit passed without code changes.

Observed on PR #1570, commit `b2e59269`.

## Evidence

- Failed attempt: https://github.com/oceanbase/powercontext/actions/runs/34663181365/job/103469752578
  Result: 1 failed, 3 passed, 1 skipped.
- Successful rerun: https://github.com/oceanbase/powercontext/actions/runs/34663181365/job/103471384127
  Result: 4 passed, 1 skipped.

## Steps to reproduce

Run the `Native personal service` workflow on a disposable macOS runner. The failing test is:

`tests/native/test_personal_service_lifecycle.py::test_native_personal_service_lifecycle`

The failure is intermittent; a deterministic reproduction is not yet available.

## Expected behavior

The registered LaunchAgent starts successfully and becomes reachable within the startup window.

## Actual behavior

`controller.install()` raises:

> the personal service was registered but did not become live: cannot reach http://127.0.0.1:49212

The controller uses a 30-second startup window. The reported `server.stderr.log` tail was empty.

The subsequent launchd diagnostic step reported that the service could not be found. The test performs cleanup in `finally`, so this diagnostic does not establish its state at the moment of failure.

## Environment

- Python 3.12.10
- macOS 26.6.2, ARM64
- Runner image: `macos-26-arm64`, version `20260907.0351.1`
- Both attempts used the same commit and runner image version.

## Investigation

Capture launchd state and process diagnostics before test cleanup, then determine whether the intermittent failure comes from startup timing, runner conditions, or service behavior. The root cause is not yet confirmed.

Modo-me:
Hello, I'm interested in this issue. Could you assign it to me?

sunlight798:
I notice that this issue has not been resolved, and I am very interested in it.

PR #1575, opened 2026-09-12
test(native): report the service's state at the moment it fails to start

L4XB:
## What

`test_native_personal_service_lifecycle` fails intermittently on the `macos-launchagent` job with "the personal service was registered but did not become live", and the run says nothing else: the only evidence printed is the tail of `server.stderr.log`, which was empty on the reported failure. The `finally` block then stops, disables and removes the registration, so the launchd diagnostic step that follows reports the service as not found whatever the failure was.

This adds the evidence the issue asks for first, at the moment of failure, before the cleanup. It does not fix the flake, and does not claim to.

Refs #1571.

## The report

Collected by `_failure_diagnostics(adapter, tmp_path)` and attached to the install failure, the `installed.ok` assertion and the post-restart status assertion:

| probe | answers |
|---|---|
| `loaded_registration`, `manager_state`, `inspect` | what the adapter itself believes, from its three independent views |
| artifact path | whether the plist / unit / task file exists, and its size |
| port | whether `127.0.0.1:<port>` accepts a connection, plus `lsof -nP -iTCP:<port>` for who holds it |
| entry point | `pgrep -fl powercontext_service_bootstrap`, or the `pythonw.exe` filter on Windows |
| platform manager | `launchctl print gui/<uid>/<label>` and `launchctl list <label>`; `systemctl --user status` plus 50 journal lines; `schtasks /Query /V` |
| logs | tails of `server.stdout.log` **and** `server.stderr.log`, distinguishing "not created" from "empty" |

Two properties worth calling out:

- **Nothing in the report can replace the failure.** Every probe swallows its own errors and records that it failed, so a diagnostic bug cannot turn "service did not start" into a `FileNotFoundError` from `pgrep`.
- **Nothing is dumped wholesale.** Each probe is scoped to this service's own label, port and entry point. The first version used `ps -Ao pid,etime,command`, and its output on my machine carried unrelated command lines, tokens and paths; a CI log is public, so the process probe matches the service's entry point instead.

The previously stderr-only helper is gone, replaced by `_log_tail`, which is what makes the absent-versus-silent distinction the reported failure lacked.

## Evidence

`test_failure_diagnostics_reports_an_absent_service` drives the report against a service that was never installed, which is the state it has to survive, and asserts every section including the platform manager's own answer. It installs nothing: all probes are read-only.

```
POWERCONTEXT_RUN_NATIVE_SERVICE_TESTS=1 pytest tests/native/test_personal_service_lifecycle.py -k failure_diagnostics
# 1 passed
```

Sample output on macOS 26 for a label that was never registered, 1.2 KB whole:

```
loaded_registration: ManagerRegistration(state=<ManagerOwnershipState.NOT_LOADED: 'not_loaded'>, definition=None, detail=None)
manager_state: inactive
inspect: NativeRegistration(state=<RegistrationState.NOT_INSTALLED: 'not_installed'>, ...)
artifact /Users/…/Library/LaunchAgents/com.oceanbase.powercontext.native.<uuid>.preview.plist: exists=False
port 55782: reachable=False
$ lsof -nP -iTCP:55782 -> exit 1
(no output)
$ pgrep -fl powercontext_service_bootstrap -> exit 1
(no output)
$ launchctl print gui/501/com.oceanbase.powercontext.native.<uuid>.preview -> exit 113
Could not find service "…" in domain for user gui: 501
server.stdout.log: not created
server.stderr.log: not created
```

Negative controls: dropping the log section or the platform section each fails the new test. `ruff check` and `ruff format --check` clean with the pinned 0.15.7.

The lifecycle tests themselves need a disposable runner (`POWERCONTEXT_RUN_NATIVE_SERVICE_TESTS=1` installs a real LaunchAgent), so I did not run them here; only the report is exercised locally.

## What this does not do

It does not change the 30-second startup window, the adapter, or the service. If the next failure is a slow start, the report will show a registration that exists with a port that is not yet accepting; if the agent never launched, it will show launchd's own refusal and no process. That is the distinction the issue needs before anything is tuned.

### DIFF
--- a/tests/native/test_personal_service_lifecycle.py
+++ b/tests/native/test_personal_service_lifecycle.py
@@ -122,11 +122,15 @@ def test_failure_diagnostics_reports_an_absent_service(tmp_path: Path) -> None:
     assert "server.stdout.log: not created" in report
     assert "server.stderr.log: not created" in report
     # The platform's own service manager was asked, and its answer is quoted.
-    expected_manager = {
+    # Annotated: without it the key type is inferred as the union of the three
+    # concrete classes, which `type(adapter)` (a `type[NativeServiceAdapter]`)
+    # cannot index.
+    expected_managers: dict[type[NativeServiceAdapter], str] = {
         LaunchdUserAdapter: "launchctl",
         SystemdUserAdapter: "systemctl",
         WindowsTaskSchedulerAdapter: "schtasks",
-    }[type(adapter)]
+    }
+    expected_manager = expected_managers[type(adapter)]
     assert expected_manager in report
 
 
--- a/tests/native/test_personal_service_lifecycle.py
+++ b/tests/native/test_personal_service_lifecycle.py
@@ -68,9 +68,9 @@ def test_native_personal_service_lifecycle(tmp_path: Path) -> None:
         try:
             installed = controller.install(env_file=environment)
         except ServiceError as error:
-            pytest.fail(f"{error}\n{_server_error_tail(tmp_path)}")
+            pytest.fail(f"{error}\n\n{_failure_diagnostics(adapter, tmp_path)}")
 
-        assert installed.ok
+        assert installed.ok, f"{installed}\n\n{_failure_diagnostics(adapter, tmp_path)}"
         assert installed.manager_ownership is ManagerOwnershipState.OWNED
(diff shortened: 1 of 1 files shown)


Teingi (approved):
LGTM

PR #1576, opened 2026-09-13
fix(website): handle Windows paths in link validation

### LINKED ISSUE #1500: bug(windows): documentation link checks report existing routes as missing
### Describe the bug

The website's link-checking gate reports valid routes as missing on Windows. The route index contains filesystem route groups and unexpanded dynamic segments, such as `/(default)/api` and `/(localized)/[lang]/(documentation)/docs/[[...slug]]`, instead of the public routes that Markdown links use.

The project invokes the Next preset from locked `next-validate-link` 1.6.7 in [`website/scripts/validate-links.ts`](https://github.com/oceanbase/powercontext/blob/80d618bd61432d1bb4eae85433fa22f2bbd50bce/website/scripts/validate-links.ts#L45). That preset receives forward-slash paths from its glob but splits route directories using `path.sep`. On Windows, the separator is a backslash, so an entire multi-segment route is treated as one segment. Route-group removal and dynamic-route population consequently do not run on the intended segments.

### Steps to reproduce

On Windows at master `80d618bd61432d1bb4eae85433fa22f2bbd50bce`, install the locked development dependencies and run:

```powershell
uv sync --frozen
pnpm --dir website install --frozen-lockfile
pnpm --dir website check:links
```

An independent reproduction isolates the route scanner from PowerContext's documents, Python API generation, and line endings. Save this in the `website` directory as `probe-routes.mts` and run it with `pnpm exec tsx probe-routes.mts`:

```typescript
import { mkdir, mkdtemp, writeFile } from 'node:fs/promises';
import os from 'node:os';
import path from 'node:path';
import { scanURLs } from 'next-validate-link';

const root = await mkdtemp(path.join(os.tmpdir(), 'pc-route-scan-'));
const page = path.join(root, 'app', '(group)', 'api', 'page.tsx');
await mkdir(path.dirname(page), { recursive: true });
await writeFile(page, 'export default function Page() { return null; }\n');

const discovered = await scanURLs({ preset: 'next', cwd: root });
console.log([...discovered.urls.keys()]);

const nativeInput = await scanURLs({
  preset: 'next', cwd: root, pages: ['(group)\\api\\page.tsx'],
});
console.log(nativeInput.urls.has('/api'));
```

### Expected behavior

The discovered `(group)/api/page.tsx` route should yield `/api`. The full website route index should omit route groups and populate localized/dynamic routes from the provided configuration on Windows as it does on Linux.

Normalize discovered path separators before splitting into route segments, through a verified dependency fix or a project adapter. Keep validation of genuinely missing routes and anchors enabled. Add Windows coverage for a route group, localized dynamic route, and a real invalid link.

### Actual behavior

The isolated Windows reproduction prints `['/(group)/api']`; supplying the equivalent native-separated input makes `/api` appear. This establishes a scanner path-separator problem without involving the documents themselves.

For the real site, the index contains only 11 concrete entries, including `/(default)/api`, and does not contain `/api` or `/en/docs/tutorials/agent-quickstart`. Full link validation reports **458 errors in 122 files** and exits 1, blocking `pnpm build` before Next runs. This count describes the observed output, not 458 independently confirmed broken documentation links.

This is separate from the homepage's CRLF frontmatter parsing problem: the minimal route fixture has no Markdown frontmatter, and the link gate runs earlier in the build pipeline.

### Environment

- Windows x64, Node.js 24.14.1, pnpm 11.13.1.
- Locked `next-validate-link` 1.6.7, Fumadocs dependencies, and Next.js 16.3.3.
- No corresponding open PowerContext issue or implementation PR was found. Open Windows-related items in the dependency repository were also checked without finding a matching fix.
- AI assistance: OpenAI Codex assisted with source inspection, the isolated scanner reproduction, and this report.

Doar999:

Closes #1500.

## Rationale for this change

On Windows, `next-validate-link@1.6.7` receives forward-slash paths from its glob but splits route segments using `path.sep`, which is a backslash. As a result, valid routes such as `/api` are indexed as `/(default)/api`, and localized dynamic routes are not populated correctly. This causes documentation link validation to reject existing pages.

- Add a pnpm patch so Next route segmentation accepts both `/` and `\`.
- Register the patch and update the lockfile without changing dependency versions.
- Add regression tests for both `app` and `src/app`, covering route groups, English/Chinese dynamic routes, and anchors.
- Verify that missing pages and invalid anchors are still rejected.
- Run the regression tests before documentation link validation.
- Add a Windows CI job; existing website CI retains Ubuntu coverage.

Windows contributors can run documentation link validation without the route-scanning false positives described in #1500.

This change affects website development tooling only.

Validated on Windows against upstream `62e4c821`:

- Both regression tests failed before the patch and passed afterward.
- `pnpm --dir website install --frozen-lockfile` passed.
- `pnpm --dir website test:links` passed.
- `pnpm --dir website lint` passed.
- `pnpm --dir website check:links` passed, validating 201 public pages and 272 repository files.
- A focused TypeScript check for the regression tests passed.
- `uv lock --check` passed.
- `uv run prek run -a` passed except for the existing repository-wide `ty-check` failure: seven Windows-specific diagnostics in unchanged `artifact_processing.py` and `test_artifact_processing.py`, involving `PipeConnection`/`Connection` and `os.WNOHANG`.

The full Next.js production build and Python test suites were not rerun for this tooling change. 

## AI usage statement

OpenAI Codex (GPT-6) assisted with reproduction, implementation, regression tests, and validation.


Teingi (approved):
LGTM

PR #1581, opened 2026-09-13
feat(website): add GitHub stars to shared navigation

zhanghuidinah:

No linked issue or RFC.

# Rationale for this change

Visitors whose browsers cannot reach GitHub should still see the repository's Star count. Include the count in the static website and refresh it asynchronously through Shields.io.

- Add a compact GitHub icon and count to shared English and Chinese navigation.
- Fetch the count through Fumadocs during the existing content-generation step and pass it into the rendered component. Store the value only in ignored build output. Existing deployment workflows provide `GITHUB_TOKEN` to the build step with their existing repository read permission.
- Read Shields.io display text once per repository/page session, shared across navigation instances. Preserve the build count on timeout, network failure, or an invalid badge response.
- Retain the bilingual invitation and decorative hand guide on wide mouse-driven screens. Below 1024px, show the static icon and count to avoid clipping inside compact navigation. Respect keyboard focus, Escape, and reduced-motion preferences.
- Keep two server-rendered behavior tests for first-paint counts, including zero, and the usable-link fallback without invented numbers. Simplify the website README to development and publishing essentials.

The initial HTML includes the build count when available. Browsers request Shields.io without a token and do not call the GitHub API or poll. CDN values may lag behind the build count; compact text such as `1.2k` is displayed without inventing an exact count.

The browser request times out after five seconds and retains the build count on failure. A failed build-time count request does not prevent website generation; if neither source supplies a count, the GitHub link remains usable without a number.

# How was this change tested?

- `pnpm test`: both behavior tests pass for English and Chinese.
- `pnpm types:check`: pass.
- `pnpm lint`: no errors; three existing unused-variable warnings in generated `.source/` files.
- `pnpm build`: static build and verification of 800 public pages and their internal links pass.
- `uv run --locked prek run -a`: pass.
- Local browser acceptance checks: count present with JavaScript disabled, CDN updates, preservation of the build count when requests fail, shared requests across navigation, and mobile display without a clipped prompt.

# AI usage statement

Codex assisted with implementation, review, and validation.


PsiACE (approved):
Totally LGTM.

I changed the Star count to render at build time, with an async update from Shields. If that request fails, the page keeps the build count. I also removed the scheduled snapshot workflow and data-branch logic.

On narrow screens, I disabled the prompt and animation that were getting clipped. I removed tests tied to implementation details, kept the first-paint behavior tests, and shortened the website README and translated it into English.

PR #1557, opened 2026-09-10
docs(rfc): record recurring failures and gate their repair

AlexStocks:
[...]

PR #1586, opened 2026-09-13
feat: record recurring failures and gate their repair (#1557)

### LINKED ISSUE #1554: feat: localize recurring failures to memory components and gate their repair (Recuris)
## Feature description

Record recurring task failures as structured negative knowledge that carries three things `Experience` does not have today:

1. a **machine-matchable trigger signature**, so that a *recurrence* of the same failure can be recognized at recall time instead of being re-learned as a fresh lesson;
2. a **repair-routing attribution**, naming which layer of the harness the fix must touch, so the response is "patch the broken layer" rather than "write another lesson";
3. an **avoided-recurrence ledger**, so an entry that keeps being recalled but never actually helps becomes visible for Review instead of persisting silently forever.

PowerContext covers the positive case well: [`ExperienceContent`](https://github.com/oceanbase/powercontext/blob/master/src/powercontext/builtin/artifacts/experience/models.py) captures reusable judgment as `situation` / `action` / `outcome` / `lesson`. What is missing is the negative counterpart with a stable identity: there is currently no way to express "this failure has happened before, here is how to recognize it, here is which layer is broken, and here is the evidence that we stopped making it".

### Research basis

Zhaochen Yu, Yingcheng Wu, Zhenfei Yin, Kaiyuan Chen, Zhe Zhao, Mengdi Wang, Shuicheng Yan, and Ling Yang. **Recursive Experiential-Working Memory Evolution for Long-Horizon Agent Harnesses** (Recuris), arXiv:2608.24876v1, August 25, 2026.

- [Abstract and publication metadata](https://arxiv.org/abs/2608.24876)
- [Full paper](https://arxiv.org/html/2608.24876v1)
- [PDF](https://arxiv.org/pdf/2608.24876)
- [Code](https://github.com/Gen-Verse/Recuris) (Apache-2.0)

Recuris models the harness memory-control layer as a four-component tuple `M_k = (E_k, W_k, rho_k, C_k)` and defines that tuple as **the patch space**: failures are localized onto components, not merely summarized.

| Component | Definition in the paper (Sections 2.1.1, 2.3.1) | Failure localized to it means |
| --- | --- | --- |
| `E` Experiential Memory | stores reusable skills in the agent-skill format | missing or flawed skill |
| `W` Working-Memory Specification | defines the state schema and update proposal for its task-specific instance | omitted goal, unsuitable state field, or incorrect state-update proposal |
| `rho` Invocation Policy | decides at which execution events skills are retrieved and which entries enter the context | missed, late, or irrelevant skill invocation |
| `C` Checker Set | tests whether observations support proposed state changes | a supported transition was rejected, or an unsupported transition was accepted |

Two properties are worth carrying over directly:

- **Attribution is a repair decision, not a causal claim.** The paper states: *"This is a repair decision rather than a claim of causal identification... Recuris attributes each diagnosed failure to the component on which a localized intervention is most likely to help."* That framing is compatible with PowerContext's evidence discipline, and is the opposite of writing a confident but unfalsifiable lesson.
- **Admission is validation-gated and bounded** (Section 2.3.3). A candidate patch is accepted only when it repairs the target failure *and* satisfies a preset regression criterion evaluated on the failed task plus a held-out development set of anchor tasks the current memory already solves. Test tasks, trajectories, and scores never enter localization, patch generation, or admission. The base model, Meta-Agent, localization/patching procedures, and the gate itself stay fixed.

**Reported results, not reproduced PowerContext results.** The authors report +17.8 points on tau-bench for GPT-5.6 Sol and +15.6 for Claude Opus 5 (to 87.9%), +16.6/+13.5 on SkillFlow for Qwen3.6-27B/35B, +32.2 points on the longest tasks, and up to 80% fewer common long-horizon failures across four benchmarks and ten models. These are author-reported single-paper numbers.

**Known limits of the paper, which bound what we should copy.** Recuris has **no per-failure recurrence counter**. It measures reuse and avoidance through aggregate proxies only: *Reach* (share of held-out tasks where the memory was invoked at least once), held-out success gains, and the dev-set regression rate (e.g. gated updates broke 4 of 42 dev tasks the base already solved). It also targets skill/memory patches admitted against a held-out dev set, which is a stronger validation budget than a per-turn production recall loop can assume.

### Reference implementation

[claw-mem](https://github.com/opensourceclaw/claw-mem) v7.6.0 (Apache-2.0, small community project) operationalizes the `E`-adjacent diagnosis into a per-item "Error Pattern Card". Its card format is a useful concrete starting point:

- [`src/types.ts`](https://github.com/opensourceclaw/claw-mem/blob/main/src/types.ts) — `RootCauseCategory = "skill-defect" | "state-defect" | "invocation-timing" | "transition-judgment"`, whose doc comment states the semantics are mapped from Recuris `E/W/rho/C` onto *"the layer the fix must touch"*; and `ErrorSignature { trigger, symptom }` (when to recall the card, and how the error looks).
- Card fields: stable semantic id `cardId` (`epc:<trigger>`, so re-storing the same id is an edit rather than a new entry), `errorSignature`, `rootCauseCategory`, `resolution` (minimum 20 characters), optional `verification`, server-owned `effectiveness`, `provenance`. A near-duplicate trigger (normalized bigram overlap, threshold 0.8) only *suggests* editing the existing card; it does not hard-reject.
- Effectiveness ledger: per-card `hitCount` / `avoidedCount` / `lastHitAt`; **5 consecutive non-avoided hits move the card to `inactive`**, and the first avoided hit revives it. Every rejection and warning is appended to an append-only audit file.

Two honest caveats about using it as a source: its README claims 100% on LoCoMo, ConvoMem, and LongMemEval simultaneously and a "subagent memory merge" that does not exist in its source, so its published numbers should not be treated as a baseline. The card format and constants above are verifiable in code, and those are the only parts cited here.

### Existing foundations and precise gaps

Code references are pinned to source `master` at [`c2016666`](https://github.com/oceanbase/powercontext/commit/c2016666037ad063e3914985dda3b5cafd5ed976), inspected on September 10, 2026.

| Existing capability | Gap this proposal addresses |
| --- | --- |
| [`ExperienceContent`](https://github.com/oceanbase/powercontext/blob/c2016666037ad063e3914985dda3b5cafd5ed976/src/powercontext/builtin/artifacts/experience/models.py) holds `situation` / `action` / `outcome` / `lesson`, all free text. | Free text is not a match key. Two encounters with the same failure cannot be reliably recognized as the same failure, so recurrence cannot be counted and a "lesson" cannot be falsified. |
| [Candidate Review](https://github.com/oceanbase/powercontext/blob/c2016666037ad063e3914985dda3b5cafd5ed976/src/powercontext/builtin/review/service.py) validates evidence, content, and revision consistency before committing. | Review validates *before* publication. Nothing observes *after* publication whether the recorded knowledge did anything. |
| [#1508](https://github.com/oceanbase/powercontext/issues/1508) connects recurring Task Outcomes to an existing Experience and evaluates a proposed Skill revision under a paired comparison. | #1508 consolidates failures into Experience and gates *Skill revisions*. It does not type the failure against a repair space, and does not track per-entry avoided recurrence in production. |
| [#1509](https://github.com/oceanbase/powercontext/issues/1509) / [#1510](https://github.com/oceanbase/powercontext/pull/1510) (Dream) provide bounded, evidence-backed Memory-to-Experience-to-Skill refinement behind Review. | Dream decides *which artifact to propose*. It assumes the input concept already exists; there is no negative-knowledge type with a trigger identity to consolidate in the first place. |
| [#1422](https://github.com/oceanbase/powercontext/issues/1422) defines the governed evaluation loop with `validation_status` independent of Candidate review. | The loop has no named outcome category for "a previously recorded failure did not recur". That is the metric this feature needs to be evaluated by. |
| Task Outcomes ([`builtin/work/models.py`](https://github.com/oceanbase/powercontext/blob/c2016666037ad063e3914985dda3b5cafd5ed976/src/powercontext/builtin/work/models.py)) and keyed Sources carry exact provenance. | Provenance is already stronger than the reference implementation's. The gap is only the negative-knowledge concept and its ledger, not evidence plumbing. |

A search of `master` for `recurrence`, `effectiveness`, and `avoided` returns no matching concept in source, so this is not partially implemented under another name.

### Proposed design direction

**1. A trigger signature as the identity, not free text.** Store a structured `error_signature` (`trigger`: when the entry should be recalled; `symptom`: the observable shape of the error) alongside the entry. This is the only field that makes recurrence detectable, and therefore the only field that makes an avoided-recurrence ledger possible. Following the reference implementation, the signature should be the entry's stable match key so that re-recording the same failure edits the entry rather than duplicating it.

**2. Repair-routing attribution, mapped onto PowerContext's actual control layer.** Do not copy the four enum names. Map the paper's patch space onto the surfaces PowerContext would actually patch:

| Recuris component | PowerContext surface the fix must touch |
| --- | --- |
| `E` Experiential Memory | Memory entry text, `ExperienceContent`, or a Skill package |
| `W` Working-Memory Specification | Handoff `objective` / `state[]` / `next_action`, or the Task Outcome fields being recorded |
| `rho` Invocation Policy | Scope recall configuration, how the `prepare` query is constructed, `assembly.sections` selection |
| `C` Checker Set | Handoff `disposition` / acceptance criteria, or a verification instruction attached to Experience or a Handoff |

The point of the enum is to route the repair, and to make "the card is right but recall never fires it" a distinct, expressible diagnosis (`rho`) rather than an invisible bug. Attribution should stay a human Review decision; it should not be inferred automatically and then treated as fact.

**3. An avoided-recurrence ledger, with Review-gated degradation instead of automatic deactivation.** When an entry is selected into prepared context, record the selection. When a later Task Outcome carries the same trigger signature, record whether the failure recurred. That gives `selected` / `avoided` / `recurred` per entry.

This is the one place where the reference implementation must **not** be copied literally: claw-mem deactivates a card automatically after 5 non-avoided hits. PowerContext's immutable revisions and the boundary recorded in [RFC 0051](https://github.com/oceanbase/powercontext/blob/c2016666037ad063e3914985dda3b5cafd5ed976/docs/en/rfcs/0051_experience_skill_artifact_families.md) ("no automatic decay, importance, or retirement in the first release") make automatic deactivation unacceptable. The equivalent behavior should be: a low-yield entry is marked as needing review and surfaced through the existing Candidate Review path, where a human decides whether to revise or retire it through explicit revision. Revival on first avoidance can be retained, since it only moves an entry back into normal recall.

**4. A write-time structural validation gate with an append-only record.** Minimum required fields, a minimum-resolution-length rule, and a **confidence floor** below which the entry is dropped rather than stored ("missing is better than wrong"). Near-duplicate triggers should *suggest* editing the existing entry rather than silently creating a near-twin. Every rejection and warning should be persisted append-only, which fits PowerContext's Source/Observation model directly.

**5. One new metric for the evaluation loop.** An "avoided recurrence" outcome category, consumed by the protocol in #1422. Without it, this feature cannot be evaluated, and the ledger has no purpose. Note the paper's own honest limitation here: it does not have this counter either, so PowerContext would be going past the published baseline rather than reproducing it.

**Explicit non-goals for a first scope.** No automatic attribution inference. No automatic retirement or decay. No model calls added to `prepare` (the constraint already recorded in [RFC 1489](https://github.com/oceanbase/powercontext/blob/c2016666037ad063e3914985dda3b5cafd5ed976/docs/en/rfcs/1489_prepared_context_text_assembly.md) applies). No cross-Scope learning.

### Open questions

1. **Where should this live?** Three options, with different costs:
   - extend `ExperienceContent` with optional `error_signature` / `attribution` fields — cheapest, but mixes positive and negative knowledge in one content shape and still leaves the ledger homeless;
   - keep it in Memory under a reserved `kind` — also cheap, but the match key and counters are not `MemoryEntryVersion` fields, so the schema grows anyway, and Memory's admission rules are written around statements that change future judgment rather than around recognizable failures;
   - a new Artifact family — cleanest for a stable match key plus a ledger, and consistent with how Experience and Skill were separated, but it carries API, review, Dashboard, and documentation cost.

   A fourth option may be the sharpest: **split the two concepts the reference implementation conflates.** A rejected approach or an API pitfall is decision knowledge and has no need for a ledger; only a *recurring* failure needs a match key and an avoided-recurrence counter. The first might belong in Memory or an Experience applicability note, the second in its own type.

2. **When is the ledger written?** Writing on every `prepare` adds a mutation to a read path. Writing only from Task Outcome linkage is later and sparser but keeps `prepare` read-only.

3. **Does attribution belong in the artifact or in a Review annotation?** Putting it in the artifact makes it durable and queryable; putting it in Review keeps the artifact immutable and the judgment auditable.

4. **How does this interact with #1508?** Should an entry with an `E`-class attribution feed Skill revision proposals, and should a `rho`-class attribution feed recall-policy evaluation instead? If yes, the attribution enum becomes the routing key between two different downstream loops.

## Alternatives considered

- **Rely on #1508 consolidation alone.** It genuinely covers recurring failures becoming Experience, and it is a prerequisite-level capability rather than a duplicate. But it has no match key, so recurrence stays uncountable; no repair typing, so "the recall policy is broken" can only be written as prose; and no per-entry yield signal after publication.
- **Adopt the reference implementation's card format as-is.** Rejected: its automatic deactivation after 5 non-avoided hits contradicts RFC 0051 and the immutable-revision model.
- **Rules-based extraction of the same three classes from a free-text history.** This is a related pattern in the same ecosystem (protected-zone extraction of confirmed facts, rejected approaches, and API pitfalls, with a confidence floor and mandatory provenance). It is attractive for provenance discipline but it is extraction from a transcript, not an identity and ledger for a known failure, so it solves a different problem. The one rule worth borrowing is the confidence floor.
- **Do nothing until the evaluation loop in #1422 lands.** Reasonable sequencing argument, and one reason this issue is filed before an RFC. But the metric this feature needs has to be requested from #1422, otherwise it will not exist when the loop is built.

## Relationship to existing work

This is not a duplicate of #1508, #1509/#1510, or #1422. The delta is narrow and specific: a **negative-knowledge identity** (trigger signature) plus a **repair-routing attribution** plus a **per-entry avoided-recurrence ledger**, all gated by the existing Review path. If maintainers judge that #1508's Experience consolidation should absorb the identity, or that #1422 should own the ledger as a generic per-artifact outcome signal, that is a better outcome than a new family, and I would rather reframe this issue than add a parallel concept.

If the direction is accepted, I am ready to open a pull request with a bilingual RFC under `docs/en/rfcs/` and `docs/zh/rfcs/`, plus the corresponding website documentation.

## Contribution

- [x] I am willing to contribute the bilingual RFC, documentation, and implementation.

AlexStocks:
Closes #1554

Implements RFC #1557 (recurring failure repair) with an append-only recurrence ledger and evidence-bound runtime recording.

## What changes are included

- Add `ExperienceContent.failure` with stable failure signatures, repair surfaces, and verification criteria.
- Add pure recurrence matching and verdict logic, the single runtime write-path orchestrator, and append-only persistence tables/repository.
- Record `selected`, `recurred`, `avoided`, and `unknown` only from positive evidence; read/prepare paths do not write recurrence events.
- Add `ScopeStats.recurrence` aggregation and API/OpenAPI generated models, mapping, and schema updates.
- Add focused unit, persistence, runtime, statistics, CLI, contract, and cross-component E2E tests.
- Add the Chinese PRD/design diagrams and synchronize troubleshooting guidance in English and Chinese.

## Compatibility

`ScopeStats.recurrence` is required in the public API. This is a breaking response-schema change; consumers must accept the new recurrence block. The recurrence tables are append-only and do not add a new artifact family or MCP tool.

## Validation

- Focused recurrence and affected tests: `109 passed`.
- API contract and JS operation tests: `48 passed`.
- Integration manifest: `26 passed`.
- `ruff check tests/ src/`: passed.
- `ruff format --check tests/ src/`: passed.
- Generated schema and enabled server OpenAPI both match `openapi/powercontext.yaml` exactly.
- Full Windows unit-test comparison: implementation tree `2252 passed, 58 skipped, 47 failed`; clean baseline `2177 passed, 58 skipped, 46 failed`. The extra implementation-tree failure was the stale CLI stats fixture and is fixed in this PR; the remaining failures are existing Windows/platform/environment failures in the same groups as baseline.
- `make check` integration and prek hooks pass; `ty check` remains non-zero with 7 pre-existing diagnostics in multiprocessing stubs and Windows-only `os.WNOHANG` tests.

## AI usage statement

Implemented and validated with OpenAI Codex (GPT-5-based coding agent). The agent generated and edited code, tests, OpenAPI generated artifacts, and documentation; all changes were reviewed against the repository contract and verified with the commands listed above.


src/powercontext/builtin/runtime/relational.py
```diff
@@ -1717,6 +1734,16 @@ async def flush(  # noqa: C901
                     )
                     candidate_ids.append(candidate.candidate_id)
                     candidates.append(candidate)
+                for proposal in await self._services.recurrence_ledger().record_window(connection, eligible_rows):
+                    candidate = await review.propose_experience(
+                        proposal.proposal,
+                        sources=proposal.sources,
+                        artifacts=(),
```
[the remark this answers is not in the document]
Teingi: [P1] Confirmed on `4ee2d2ae` through the public runtime incubation API with SQLite. The third recurrence raises `InvalidCandidateError: invalid Candidate artifacts: must include the exact target experience Artifact`. The transaction rolls back, leaving the recurrence count at 2 and the pending Inbox empty; retrying fails again. Pass the exact target in `artifacts` and add a test that actually submits the recurrence proposal to Review. The current E2E test stops at the returned proposal.

src/powercontext/builtin/artifacts/experience/models.py
```diff
@@ -16,23 +16,72 @@
...
+
+
+class ExperienceContent(_ExperienceValue):
     """A reusable judgment grounded in exact task evidence."""
 
     situation: ExperienceText
     action: ExperienceText
     outcome: ExperienceText
     lesson: ExperienceText
+    failure: FailureRecord | None = None
```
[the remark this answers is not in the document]
Teingi: [P2] Confirmed on `4ee2d2ae` with SQLite: I proposed an Experience containing a failure block whose only cited Source was a `succeeded` Task Outcome with no failed checks, then approved it through the public runtime API. Both calls succeeded and published the failure record. Please validate that the cited content actually supports a failure before admitting a failure-bearing candidate; checking that the Source exists is insufficient.

src/powercontext/builtin/persistence/recurrence.py
```diff
@@ -0,0 +1,271 @@
...
+            return 0
+        linked_keys = {(family, artifact_id, revision, key) for family, artifact_id, revision, key in linked}
+        rows = (
+            await connection.execute(
+                select(
+                    RECURRENCE_OBSERVATION_TABLE.c.family,
+                    RECURRENCE_OBSERVATION_TABLE.c.artifact_id,
+                    RECURRENCE_OBSERVATION_TABLE.c.revision,
+                    RECURRENCE_OBSERVATION_TABLE.c.signature_key,
+                ).where(RECURRENCE_OBSERVATION_TABLE.c.scope_id == scope)
```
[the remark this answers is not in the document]
Teingi: [P2] Confirmed on `4ee2d2ae` with SQLite. A Handoff cites an Experience revision, and a separate Task Outcome without a receipt records one `recurred` event for it. There is no `selected` event or Handoff linkage, but `unlinked_handoff_citations` returns 0 instead of 1. Only events with a valid Handoff association should clear this coverage gap; an unlinked recurrence does not establish that association.

src/powercontext/builtin/artifacts/experience/recurrence.py
```diff
@@ -0,0 +1,667 @@
...
+    verified with non-empty exact evidence and a ``failed``, ``timed_out``, or
+    ``unavailable`` status. ``skipped``, ``cancelled``, and ``unknown`` never
+    qualify.
+    """
+
+    if kind == "observation":
+        if index >= len(outcome.observations) or outcome.status not in _FAILURE_OUTCOME_STATUSES:
+            return False
+        claim = outcome.observations[index]
+        return claim.basis == "verified" and bool(claim.evidence)
```
Teingi, line 138: [P2] Require signature-specific failure evidence

A verified observation can describe a normal risk condition even when the task fails for an unrelated reason. With a cue describing the contract being edited, a matching verified observation, a passing bound check, and an unrelated failed database check, the SQLite probe records `recurred` from the observation and suppresses `avoided`. The parent status does not establish that this particular failure occurred. Require evidence linking the signature to an actual failure before writing the recurrence.

src/powercontext/builtin/runtime/recurrence.py
```diff
@@ -0,0 +1,507 @@
...
+                matches.append(existing)
+                continue
+            item = _item(outcome, failure_ref)
+            if item is None:
+                continue
+            candidates = freeze_candidate_set(
+                mode=mode,
+                refs=tuple(ref for ref, _ in contents),
+            )
+            eligible = eligible_candidates(failure_item_text(item), contents)
```
Teingi, line 321: [P2] Use a consistent check binding for passed and failed verdicts

`avoided_refs()` binds a check through `verification.check_subject`, but this path matches its name against `recall_cue`. Using the fixture's distinct cue and check subject, a verified passing check records `avoided`; the corresponding failed outcome for the same check produces `unmatched` and leaves only `selected`. A failure of the declared verification therefore never increments recurrence or its review streak. Use a consistent evidence binding for both outcomes and cover this passed-to-failed case.

src/powercontext/builtin/artifacts/experience/recurrence.py
```diff
@@ -0,0 +1,667 @@
...
+    Pydantic model and therefore cannot key a mapping.
+    """
+
+    try:
+        key = normalize_match_text(failure_text)
+    except ValueError:
+        return ()
+    return tuple(
+        ref
+        for ref, content in sorted(candidates, key=lambda pair: _ref_identity(pair[0]))
```
Teingi: [P2] Deduplicate eligible revisions before deciding ambiguity

A Handoff can cite the same immutable Experience revision more than once. This comprehension preserves those duplicates even though `freeze_candidate_set()` removes them. In SQLite, changing one citation to two identical citations changes `matched` to `ambiguous` and drops the `recurred` event, while the persisted candidate set still contains exactly one revision. Deduplicate by the complete ArtifactRef identity before counting eligible candidates, and before applying the citation limit.

Teingi (commented):
Reviewed `4ee2d2ae`. All 146 focused tests passed, but separate SQLite probes reproduced these three issues and the three existing findings I replied to. The probes cover the public runtime and ledger behavior; they do not include a real LLM or OceanBase run.

src/powercontext/builtin/persistence/recurrence.py
```diff
@@ -0,0 +1,271 @@
...
+            "result": match.result,
+            "target_family": None if match.artifact_ref is None else match.artifact_ref.family,
+            "target_artifact_id": None if match.artifact_ref is None else match.artifact_ref.artifact_id,
+            "target_revision": None if match.artifact_ref is None else match.artifact_ref.revision,
+            "signature_key": match.signature_key,
+            "payload": dump_model(match, kind=_MATCH_KIND, name=key),
+        }
+        try:
+            await connection.execute(insert(RECURRENCE_MATCH_TABLE).values(**values))
+        except IntegrityError:
```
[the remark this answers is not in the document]
AlexStocks: Fixed in 6adc1e14: match and observation writes now run inside a nested savepoint; after an integrity conflict the persisted decision is re-read by its immutable identity, and constraint errors that cannot be confirmed as replays are re-raised. Added regression tests for a differing replay payload and for a non-replay `IntegrityError`; the affected tests pass.

src/powercontext/builtin/runtime/recurrence.py
```diff
@@ -0,0 +1,507 @@
...
+        return None
+    try:
+        return TaskOutcome.model_validate_json(value.content)
+    except ValidationError:
+        return None
+
+
+def _item(outcome: TaskOutcome, ref: TaskOutcomeItemRef, /) -> Any | None:
+    items: tuple[Any, ...] = outcome.checks if ref.item_kind == "check" else outcome.observations
+    return items[ref.item_index] if ref.item_index < len(items) else None
```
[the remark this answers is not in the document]
AlexStocks: Fixed in 6adc1e14: resolving a Task Outcome locator now recomputes the canonical digest of the referenced item. An out-of-range index or a digest mismatch is rejected, so a different item can no longer be used as evidence.

src/powercontext/builtin/runtime/relational.py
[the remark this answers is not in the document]
AlexStocks: Fixed in 6adc1e14: the recurrence proposal now passes the exact target `ArtifactRef` in `artifacts`, and the cross-layer E2E test covers the third recurrence reaching Review. The recurrence ledger and E2E tests pass.

src/powercontext/builtin/artifacts/experience/models.py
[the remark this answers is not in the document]
AlexStocks: Fixed in 6adc1e14: the admission gate now requires a failure proposal to cite a resolvable task-outcome Source and to contain at least one failure item that satisfies the failed/blocked + verified + non-empty evidence rule and whose normalized text equals the normalized `recall_cue`. Unrelated successful or failing evidence is rejected; regression tests were added.

src/powercontext/builtin/persistence/recurrence.py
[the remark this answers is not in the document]
AlexStocks: Fixed in 6adc1e14: `unlinked_handoff_citations` now derives linkage from `selected` events only, so an isolated `recurred` row no longer clears the provenance gap.

src/powercontext/builtin/artifacts/experience/recurrence.py
AlexStocks: Fixed in 6adc1e14: signature-specific admission now restricts `failure_refs` to qualified failure items whose normalized text strictly equals the proposal's normalized `recall_cue`, so an unrelated failure cannot be treated as evidence for that signature. Added a regression test with an unrelated failed check.

src/powercontext/builtin/runtime/recurrence.py
AlexStocks: This one should not be changed as suggested. RFC/P0-11 states that the failure-matching input is the failure item itself; for a `TaskCheck` that is `TaskCheck.name`, compared against `FailureSignature.recall_cue`. `FailureVerification.check_subject` binds only the check that is expected to run and pass on the `avoided` path (P0-28). `recall_cue` and `check_subject` are therefore allowed to differ, and switching the failing check to `check_subject` would yield `unmatched` per the contract, which is the expected behavior.

src/powercontext/builtin/artifacts/experience/recurrence.py
AlexStocks: Fixed in 6adc1e14: `eligible_candidates` now deduplicates by full immutable ArtifactRef identity, and does so before counting candidates; runtime handoff citations are also deduplicated before truncation.

src/powercontext/builtin/artifacts/experience/recurrence.py
Teingi: [P2] Still reproducible on `5f51b7f` through the public SQLite runtime. A verified observation says `contract file was edited`, the bound `make contract-test` passes, and an unrelated database health check fails. The failure Experience is admitted, and a linked follow-up produces `selected=1, recurred=1, avoided=0`.

The observation already equals `recall_cue`, so the added text comparison does not address this case: it proves that the risk condition occurred, not that this particular failure occurred. Please require signature-specific failure evidence before classifying an observation as `recurred`. The broad observation rule also needs updating; otherwise it continues to suppress the valid `avoided` result.

src/powercontext/builtin/persistence/tables.py
```diff
@@ -1091,11 +1091,102 @@ def _entry_text_type():
...
+    SHARED_METADATA,
+    Column("scope_id", identity_string(MAX_SCOPE_ID_LENGTH), primary_key=True),
+    Column("match_key", identity_string(71), primary_key=True),
+    Column("task_outcome_source_type", identity_string(MAX_SOURCE_TYPE_LENGTH), nullable=False),
+    Column("task_outcome_source_id", identity_string(MAX_SOURCE_ID_LENGTH), nullable=False),
+    Column("task_outcome_position", BigInteger, nullable=False),
+    Column("failure_item_kind", String(16), nullable=False),
+    Column("failure_item_index", Integer, nullable=False),
+    Column("failure_item_digest", identity_string(71), nullable=False),
+    Column("candidate_set_mode", String(16), nullable=False),
```
Teingi: [P1] Widen candidate_set_mode to fit handoff_citations

`handoff_citations` is 17 characters, but this column is `VARCHAR(16)`. On `5f51b7f`, I reproduced this with the actual table definitions and `RecurrenceRepository.append_match()` against a fresh local MySQL 8.0.45 instance in strict mode: `scope_heads` inserts successfully, while `handoff_citations` raises `DataError 1406: Data too long for column 'candidate_set_mode' at row 1`.

Any linked-Handoff failure window that reaches this insert aborts the incubation transaction, so its cursor cannot advance and retries fail again. Please widen the column and cover both modes with a real MySQL-compatible database write.

src/powercontext/builtin/review/service.py
```diff
@@ -550,6 +572,27 @@ def _validate_proposal_family(family: str, proposal: object) -> None:
...
+        outcome = TaskOutcome.model_validate_json(value.content)
+    except ValidationError:
+        return False
+    try:
+        cue_key = normalize_match_text(cue)
+    except ValueError:
+        return False
+    for ref in failure_refs(outcome, row.ref):
+        items = outcome.checks if ref.item_kind == "check" else outcome.observations
+        if normalize_match_text(failure_item_text(items[ref.item_index])) == cue_key:
```
Teingi: [P2] Resolve nested failure evidence before admitting the candidate

This validates the parent Task Outcome and matching text, but never resolves the citations inside the matching item. On `5f51b7f`, I used public Source capture to store a schema-valid Task Outcome whose verified failed check cites `SourceRef(source_type='content', source_id='never-existed')`. Public propose and approve both succeeded, and incubation then recorded `recurred=1`.

A non-empty evidence tuple is insufficient when the referenced evidence does not exist. Please resolve the relevant nested Source/Artifact citations in the current scope during admission and ledger recording, including outcomes ingested through generic Source capture.

Teingi (commented):
Rechecked `5f51b7f`. I recommend addressing the P1 and three P2 findings before merging. The signature-specific false positive is detailed in [my reply to the existing discussion](https://github.com/oceanbase/powercontext/pull/1586#discussion_r4015030209); the other three findings are inline below.

The 210 focused tests passed on this exact head, and I reran all four counterexamples using public SQLite runtime/repository probes and a disposable local MySQL 8.0.45 instance. The third recurrence now reaches Review successfully. These probes do not include a real OceanBase or external LLM run.

src/powercontext/builtin/runtime/recurrence.py
```diff
@@ -0,0 +1,543 @@
...
+                    ARTIFACT_HEADS_TABLE.c.artifact_id,
+                    ARTIFACT_HEADS_TABLE.c.revision,
+                )
+                .where(
+                    ARTIFACT_HEADS_TABLE.c.scope_id == self.scope_id,
+                    ARTIFACT_HEADS_TABLE.c.family == Experience.family,
+                    ARTIFACT_HEADS_TABLE.c.lifecycle_state == "active",
+                )
+                .order_by(ARTIFACT_HEADS_TABLE.c.artifact_id, ARTIFACT_HEADS_TABLE.c.revision)
+                .limit(MAX_RECURRENCE_CANDIDATES)
```
Teingi: [P2] Filter failure candidates before applying the 64-entry bound

This limit is applied to all active Experience heads before checking for failure metadata. In real SQLite repository/ledger probes on `5f51b7f`, 63 ordinary Experiences followed by the only matching failure Experience produce `matched` and one `recurred` event. Adding one ordinary Experience changes the same failure to `unmatched` with no event because the matching head is now outside the first 64 IDs.

Ordinary Experiences should not consume the failure-candidate budget. Please filter relevant failure candidates before bounding the snapshot, and represent an incomplete scan explicitly instead of freezing it as a definitive `unmatched` decision.

src/powercontext/builtin/persistence/tables.py
AlexStocks: 已将 pc_recurrence_match.candidate_set_mode 扩大到 VARCHAR(32)，并新增模式长度回归测试，覆盖 handoff_citations 与 scope_heads。

src/powercontext/builtin/review/service.py
AlexStocks: 已在 Experience candidate admission 中解析匹配失败项的嵌套 Source、Artifact 和 Memory 引用；不存在的引用不再构成有效 failure evidence。已添加 generic Source capture 的缺失 Source 回归测试。

src/powercontext/builtin/runtime/recurrence.py
AlexStocks: 已移除 ledger 匹配候选的未文档化 64 条截断：Handoff 引用与 scope head 现在都会冻结完整、有序的候选快照，仍保持 matched/unmatched/ambiguous 的既有确定性合约。新增 SQLite 回归覆盖 64 条普通 Experience 后的唯一 failure Experience，验证会得到 matched 并写入 recurred。统计路径独立的有界 Handoff 扫描未改动。

AlexStocks:
针对 `5f51b7f` 的复查结论，当前 Head `8202c77f` 已完成以下修复：

- `candidate_set_mode` 已从 `VARCHAR(16)` 扩展为 `VARCHAR(32)`，覆盖 `handoff_citations` 与 `scope_heads`；
- Experience candidate admission 现在会解析匹配 failure item 内嵌的 Source、Artifact、Memory 引用，缺失引用不再构成 failure evidence；
- ledger 匹配候选不再对 Handoff citations 或 scope heads 作未文档化的 64 条截断；新增 SQLite 回归覆盖 64 条普通 Experience 后的唯一匹配 Experience，确认会得到 `matched` 并写入 `recurred`。

关于 `signature-specific false positive`：该反例在当前 RFC/PRD 的既有规则下仍是可表达的边界，而不是已被实现遗漏的检查。P0-11 规定 `recurred` 以失败 item 本身（`WorkClaim.text` 或 `TaskCheck.name`）与 `recall_cue` 的精确归一化匹配判定；P0-28 则仅把 `check_subject` 用于 `avoided` 路径中通过的绑定检查。将该 observation 场景改为不产生 recurrence，会改变这两个既有绑定的语义，需要先形成 RFC/PRD 层面的 contract 决策，而不应在本 PR 中静默改写为另一套判定规则。

当前 Head 的 quality、Python 3.11-3.14、SQLite/OceanBase acceptance 及其余可见检查均已通过。该结论仅覆盖当前 CI 和已处理评论，不替代所需的人工 review 决策。

AlexStocks:
@Teingi current head `df8e9392` has no merge conflict, no current unresolved review thread, and all visible checks are green. The prior review findings have been addressed or replied to. Please continue review when you have time.

src/powercontext/builtin/persistence/recurrence.py
```diff
@@ -0,0 +1,303 @@
...
+            "result": match.result,
+            "target_family": None if match.artifact_ref is None else match.artifact_ref.family,
+            "target_artifact_id": None if match.artifact_ref is None else match.artifact_ref.artifact_id,
+            "target_revision": None if match.artifact_ref is None else match.artifact_ref.revision,
+            "signature_key": match.signature_key,
+            "payload": dump_model(match, kind=_MATCH_KIND, name=key),
+        }
+        try:
+            async with connection.begin_nested():
+                await connection.execute(insert(RECURRENCE_MATCH_TABLE).values(**values))
```
Teingi, line 81: [P1] Keep recurrence savepoints inside the outer SQLite transaction

With the current SQLite profile, if the window has only read before this call, there is no actual BEGIN before SAVEPOINT. Releasing the savepoint commits the ledger write, so a later failure cannot roll it back. The same applies to `append_observation()`.

On this head, both in-memory and file SQLite retain one match and one event after an outer rollback; an explicit-BEGIN control retains neither. Through the public runtime, I used valid evidence and injected one transient error before the third recurrence's Review insert. After the error, `recurred=3` was already persisted. Retrying advanced the cursor but returned `candidate_count=0` with an empty Inbox: `_record_recurrences()` treats the retained event as a replay and skips the proposal.

Please ensure these savepoints participate in a real outer transaction, and cover rollback/retry of the ledger, candidate, and cursor together.

src/powercontext/builtin/runtime/recurrence.py
```diff
@@ -0,0 +1,541 @@
...
+
+        matches: list[RecurrenceMatch] = []
+        for failure_ref in failure_refs(outcome, row.ref):
+            existing = await self.recurrence.find_match(connection, self.scope_id, row.ref, failure_ref)
+            if existing is not None:
+                matches.append(existing)
+                continue
+            item = _item(outcome, failure_ref)
+            if item is None:
+                continue
```
Teingi, line 332: [P2] Resolve nested evidence before recording recurrence verdicts

The admission fix validates nested citations when proposing a failure Experience, but this ledger path still checks only the Outcome item and its digest. It never resolves the item's evidence. `_record_avoided()` has the same gap.

On this head, I first approved a failure Experience using valid evidence, then used public Source capture to ingest failed checks citing `SourceRef(source_type='content', source_id='never-existed')`. Two incubation windows still produced `recurred=1` and `recurred=2`. In another scope, a condition and passing check with that nonexistent citation produced `avoided=1` and reset an existing streak from 1 to 0.

Please resolve the relevant nested Source, Artifact, and Memory references in the current scope before freezing matches or recording terminal verdicts. Unresolvable evidence must remain insufficient; rejecting a candidate only when the third recurrence reaches Review leaves the earlier ledger entries incorrect.

Teingi (commented):
Reviewed `df8e9392`. I recommend fixing the P1 and P2 below before merging.

All 218 focused tests passed. Separate SQLite probes reproduced both findings, including a valid-evidence retry case with one injected transient Review failure. I also confirmed the column-width fix on MySQL 8.0.45 in strict mode and the candidate-snapshot fix beyond 64 entries. Normal third-recurrence Review admission succeeds. This validation does not include a real OceanBase or external LLM run.

src/powercontext/builtin/persistence/recurrence.py
AlexStocks: Fixed in `774a0de7`.

`append_match()` and `append_observation()` now ensure SQLite has a real outer `BEGIN` before opening their nested savepoints. This keeps the ledger writes inside the incubation transaction even when the window only performed reads before the append. Added rollback regressions for both match and observation writes after read-only work; both now disappear after the outer rollback.

Local validation:

- `33 passed` for `tests/builtin/persistence/test_recurrence_repository.py`, `tests/builtin/runtime/test_recurrence_ledger.py`, and `tests/e2e/test_recurring_failure_repair.py`.
- ruff check, ruff format --check, git diff --check, and scoped `ty check` passed on the changed files.
- Full `ty check` still reports pre-existing unrelated diagnostics in `artifact_processing.py`, `opencode.py`, and `test_artifact_processing.py`; none point at this change.

GitHub checks are running on `774a0de7`.

src/powercontext/builtin/runtime/recurrence.py
AlexStocks: Fixed in `774a0de7`.

The recurrence ledger now validates the nested evidence on the exact TaskOutcome items before recording terminal verdicts. `_freeze_matches()` validates the failure item evidence before freezing a match that can become `recurred`; `_record_avoided()` validates both the condition and passed-check evidence before writing `avoided`. Unresolvable Source, Artifact, or Memory citations now leave that item insufficient instead of writing an incorrect ledger verdict.

Added regressions for both sides: a failed check citing a missing Source no longer records `recurred`, and a passed condition/check pair citing a missing Source no longer records `avoided`.

Local validation:

- `33 passed` for `tests/builtin/persistence/test_recurrence_repository.py`, `tests/builtin/runtime/test_recurrence_ledger.py`, and `tests/e2e/test_recurring_failure_repair.py`.
- ruff check, ruff format --check, git diff --check, and scoped `ty check` passed on the changed files.
- Full `ty check` still reports pre-existing unrelated diagnostics in `artifact_processing.py`, `opencode.py`, and `test_artifact_processing.py`; none point at this change.

GitHub checks are running on `774a0de7`.

Teingi (approved):
LGTM

PR #1587, opened 2026-09-13
ci: harden workflow reproducibility and evidence gates

AlexStocks:
## Summary

This maintenance PR brings the following CI hardening patterns from `powercontext-go` to the Python repository:

- Pin every third-party GitHub Action in the tracked workflows and composite environment action to an immutable commit SHA, while retaining the human-readable version comments.
- Add least-privilege read permissions, concurrency control, bounded job timeouts, and disable persisted checkout credentials on read-only checkouts where the workflow does not need them.
- Make the E2E replay-evidence secret scan fail closed: missing or empty evidence and scanner failures fail the acceptance job; raw evidence is uploaded only after a successful scan.
- Upload only a bounded diagnostic summary when evidence validation fails, and run the scanner without network access.

The existing Python workflow topology, SQLite/OceanBase acceptance matrix, release jobs, and product behavior are unchanged. Go-specific build and release gates are intentionally out of scope.

## Verification

- `actionlint` on all tracked workflow files: passed
- YAML/action policy contract check: passed; 55 external action references are pinned to 40-character SHAs
- `git diff --check`: passed
- `make contract-test`: passed (48 tests)
- `make harness-sync && make harness-check`: passed (34 tests)
- `make check`: YAML, formatting, and Ruff checks passed; the existing baseline `ty` stage remains blocked by 7 unrelated diagnostics in `src/powercontext/builtin/runtime/artifact_processing.py` and `tests/builtin/runtime/test_artifact_processing.py`
- Docker Compose acceptance was not run locally because Docker is unavailable in the validation environment

No issue is referenced because this is a bounded CI maintenance change rather than an implementation of an existing issue.

PR #1589, opened 2026-09-14
ci: improve evidence scan diagnostics and action pinning

AlexStocks:
Related to #1587.

# Rationale for this change

The CI hardening in #1587 made replay evidence fail-closed, but the scanner step discarded all output and treated TruffleHog findings and scanner infrastructure failures identically. The repository also had no automated guard or update configuration for its pinned GitHub Actions.

- Keep TruffleHog JSON and stderr in runner temporary storage, derive bounded sanitized diagnostics, and upload only `summary.txt` when the scenario or scan fails.
- Classify exit code 183 as `findings`; classify other non-zero results as `infra_error` after one retry.
- Upgrade the pinned TruffleHog image to the verified v3.97.4 digest.
- Add a repository checker that rejects mutable third-party Action references and run it from `make check`.
- Add weekly Dependabot updates for GitHub Actions.
- Add contract and fake-Docker tests for pinning, scan classification, retry behavior, and redaction.

No product or API changes. CI failures now include actionable sanitized diagnostics while raw scanner output remains excluded from artifacts.

# How was this change tested?

- `uv run python -m pytest tests/test_workflow_actions.py tests/test_e2e_workflow_contract.py tests/test_scan_e2e_evidence.py tests/test_api_contract.py tests/test_js_operations.py -q` (57 passed)
- `uv run ruff check` and `uv run ruff format --check` on changed Python files
- Actions pin checker: 67 references checked
- Bash syntax check, YAML parsing, and `git diff --check`
- `make contract-test` (48 passed)
- `make check` reaches the existing `ty` gate but remains blocked by pre-existing type errors in `src/powercontext/builtin/runtime/artifact_processing.py` and `tests/builtin/runtime/test_artifact_processing.py`.

# AI usage statement

Implemented with Codex using repository-local tests and review evidence.


scripts/scan_e2e_evidence.sh
```diff
@@ -0,0 +1,135 @@
...
+    echo "scan_status=infra_error" >> "${summary}"
+    echo "scanner_exit_code=${scanner_status}" >> "${summary}"
+    {
+        echo "scanner_stderr_tail_begin"
+        tail -n 50 "${scanner_stderr}" \
+            | tr -cd '\11\12\15\40-\176' \
+            | sed -E \
+                -e 's/(Authorization:[[:space:]]*Bearer[[:space:]]+)[^[:space:]]+/\1[REDACTED]/Ig' \
+                -e 's/((token|secret|password|api[_-]?key)[=:][[:space:]]*)[^[:space:]]+/\1[REDACTED]/Ig' \
+                -e 's/[A-Za-z0-9+\/_=-]{32,}/[REDACTED]/g'
```
Teingi: [P1] Keep arbitrary scanner text out of uploaded diagnostics

These substitutions miss JSON password fields, Basic auth, and credentials in URLs. With a fake Docker command returning 125, each of those synthetic credentials survived unchanged in `summary.txt`. I also reproduced a path containing `{"password":"SyntheticP4ss!"}` reaching the summary through the findings branch with the exact pinned v3.97.4 image. Since the workflow uploads this file on failure, withholding the raw files does not prevent a credential leak. Could we emit only explicitly allowed diagnostic fields and apply the same safe handling to finding paths, rather than copying arbitrary stderr through these regexes?

scripts/check_workflow_actions.py
```diff
@@ -0,0 +1,75 @@
...
+
+"""Require immutable commit pins for third-party GitHub Actions."""
+
+from __future__ import annotations
+
+import re
+import sys
+from pathlib import Path
+
+ACTION_REFERENCE = re.compile(r"^\s*(?:-\s*)?uses:\s*([^\s#]+)", re.MULTILINE)
```
Teingi: [P2] Parse YAML before checking action references

Both `- {uses: actions/checkout@v7}` and `- "uses": actions/checkout@v7` are valid YAML, but the checker exits 0 with `0 action references checked` for either, so mutable references can pass the new gate. Conversely, quoting a valid SHA-pinned reference or a local action makes it fail. Could we parse the workflow/composite-action YAML and inspect the actual `uses` values? Regression cases for these equivalent YAML spellings would cover both the bypass and the false rejection.

scripts/scan_e2e_evidence.sh
AlexStocks: Fixed in 5aef9002. The uploaded summary no longer copies scanner stderr or arbitrary finding text. Infrastructure failures now expose only the scanner exit code, retry count, fixed error category, bounded tail line count, and stderr SHA-256; raw JSON/stderr remain runner-local. Finding DetectorName and repository-relative paths pass an explicit allowlist and unsafe values become [REDACTED]. Regression coverage includes JSON password fields, Basic auth, URL credentials, and credential-bearing finding paths. Focused tests: 13 passed including the E2E workflow contract.

scripts/check_workflow_actions.py
AlexStocks: Fixed in 5aef9002. The pin checker now parses each YAML file with PyYAML, recursively inspects actual uses keys, rejects malformed YAML, accepts quoted SHA/local action references, and catches flow-style and quoted-key mutable references. Repository inventory still reports 67 checked references. Regression coverage includes both bypass spellings, quoted valid references, nested composite actions, and malformed YAML.

scripts/scan_e2e_evidence.sh
AlexStocks: Additional self-review hardening is in 5c250c6c: DATABASE/SCENARIO_OUTCOME/image headers are now allowlisted, and an invalid scanner image is rejected before Docker is invoked. Non-ASCII detector/path values are fixed-redacted. The clean immutable-digest path also has a regression test.

scripts/check_workflow_actions.py
AlexStocks: Additional self-review hardening is in 5c250c6c: duplicate YAML keys and unhashable mapping keys now fail closed without echoing source text, aliases are cycle-safe, traversal is iterative, and parse/violation diagnostics redact untrusted YAML values. The repository inventory remains 67 checked references.

AlexStocks: The follow-up quality failure was traced to a ty warning from implicit yaml.resolver access; fc606e55 explicitly imports BaseResolver. Local make check now reports only the seven pre-existing artifact_processing.py/type-check diagnostics, with no warning from this checker.

scripts/scan_e2e_evidence.sh
Teingi: [P1] Credential-bearing paths still reach the uploaded summary

Rechecked fc606e55 with the pinned TruffleHog v3.97.4 image. A synthetic GitHub token used as the directory name in `ghp_…/replay.json` survives unchanged in `summary.txt`: its letters, digits, and underscores pass the character allowlist without matching any blocked keyword. The workflow then uploads that summary when the scan fails. The original JSON-password and stderr cases are fixed, but this path still leaks the full credential. Could we omit or hash arbitrary paths, or restrict them to predefined safe paths?

scripts/scan_e2e_evidence.sh
```diff
@@ -0,0 +1,241 @@
...
+    if command -v jq >/dev/null 2>&1; then
+        while IFS=$'\t' read -r detector path; do
+            detector="${detector%$'\r'}"
+            path="${path%$'\r'}"
+            printf 'finding=%s path=%s\n' \
+                "$(sanitize_detector "${detector}")" \
+                "$(sanitize_path "${path}")" >> "${summary}"
+        done < <(
+            jq -r '
+                limit(20;
```
Teingi, line 192: [P2] Apply the findings limit to the whole JSONL stream

`jq` runs this filter separately for each input object, so `limit(20; ...)` resets for every JSONL record. A fake scanner emitting 45 findings produced all 45 `finding=` lines in `summary.txt`. With many findings, both the uploaded summary and the per-finding shell work grow without the intended bound. Could we limit the complete input stream, for example with `jq -nr 'limit(20; inputs | ...)'`, and cover a scan containing more than 20 records?

Teingi (approved):
LGTM

PR #1592, opened 2026-09-14
fix(opencode): normalize Windows file URI paths

### LINKED ISSUE #1497: bug(windows): OpenCode doctor misidentifies valid file URI plugins
### Describe the bug

On Windows, `powercontext doctor opencode` can report that a PowerContext plugin is not configured even when OpenCode's configuration contains a valid `file:///C:/...` reference to that plugin. It also loses the distinction between a configured-but-inactive plugin and a missing configuration.

In [`_configured_plugin()`](https://github.com/oceanbase/powercontext/blob/80d618bd61432d1bb4eae85433fa22f2bbd50bce/src/powercontext/cli/opencode.py#L324), a file URI is decoded with `unquote(parsed.path)` and passed directly to `Path`. On Windows, the resulting `/C:/...` is not the native `C:/...` filesystem path, so the existing `package.json` cannot be found. The bridge installed under OpenCode's config directory can mask this problem through the alternative `_owned_plugin()` check; configurations using the file URI directly remain affected.

### Steps to reproduce

Use a Windows checkout of `80d618bd61432d1bb4eae85433fa22f2bbd50bce` with development dependencies installed:

```powershell
uv run pytest -q tests/test_opencode_cli.py::test_doctor_opencode_rejects_configured_but_inactive_plugin
```

The existing test invokes the public Doctor CLI with a valid plugin directory, an OpenCode config containing `plugin.as_uri()`, and an inactive activation probe. No installed OpenCode instance or model credentials are required for this deterministic regression.

The path-decoding cause can also be isolated from the host process:

```python
import json
from pathlib import Path
from powercontext.cli.opencode import _configured_plugin

plugin = Path("integrations/opencode/plugins/powercontext").resolve()
print(_configured_plugin(json.dumps({"plugin": [str(plugin)]})))
print(_configured_plugin(json.dumps({"plugin": [plugin.as_uri()]})))
```

### Expected behavior

Both references identify the same installed plugin. Doctor should independently report configuration presence and runtime activation; the inactive fixture should report `configured but did not activate`.

Convert file URIs using Windows-aware path handling, including drive letters and escaped characters. Add appropriate UNC coverage and retain the current distinction between filesystem references and package specifications. The successful native-path and installed-bridge cases must remain valid.

### Actual behavior

The isolated comparison prints `True` followed by `False`. The public CLI regression fails with:

```text
AssertionError: assert 'did not activate' in 'PowerContext OpenCode plugin is not configured'
```

This is an incorrect diagnostic and can send users toward reinstalling or editing configuration when the failure is actually activation. It is independent of the DSH Doctor changes in #1493 / #1494.

### Environment

- Windows x64, CPython 3.12.13.
- Reproduced with unmodified sources in an isolated checkout.
- Open issues and the changed files of open PRs were checked; no existing open fix for this URI conversion was found.
- AI assistance: OpenAI Codex assisted with source inspection, reproduction, and this report.

sunlight798:

Closes #1497 

# Rationale for this change

On Windows, OpenCode may report configured plugins using a `file:///C:/...` URI. The existing diagnostic code decoded the URI path directly and passed `/C:/...` to `Path`, which is not a valid native Windows drive path.

As a result, `powercontext doctor opencode` could incorrectly report that the PowerContext plugin was not configured, even when the configuration was valid. This also hid the distinction between a missing configuration and a configured plugin that failed to activate.

- Convert file URI paths using the platform-aware `urllib.request.url2pathname()` function.
- Preserve URI authorities so Windows UNC paths such as `file://server/share/...` are handled correctly.
- Continue treating native filesystem paths and package specifications separately from file URIs.
- Extend the configured-but-inactive regression test to cover percent-encoded paths containing spaces.

On Windows, `powercontext doctor opencode` now correctly recognizes PowerContext plugins configured through valid file URIs.

# How was this change tested?

- `uv run pytest -q tests/test_opencode_cli.py` — 17 passed.
- `make check` — passed.
- `make unit-test` — 2240 passed, 45 skipped, and 3 unrelated failures in Linux-only systemd receiver-service tests.

# AI usage statement

OpenAI Codex was used to inspect the affected implementation, analyze the cross-platform path behavior


Teingi (approved):
LGTM

Issue #1501, opened 2026-09-08
bug(windows): unit tests rely on POSIX semantics and overlong temporary paths


knqiufan:
### Describe the bug

Several unit tests for portable behavior fail on Windows because their fixtures or assertions depend on POSIX semantics or excessively long pytest temporary paths. These failures prevent a useful Windows unit-test baseline and can obscure genuine regressions.

The failures below are test portability defects. They should not be reported as evidence that Windows authorization, subprocess cleanup, or default host installation is broken.

### Steps to reproduce

On Windows at master `80d618bd61432d1bb4eae85433fa22f2bbd50bce`, with development dependencies installed, run these existing nodes through `uv run pytest -q`:

| Test node | Incorrect platform assumption |
| --- | --- |
| `tests/builtin/artifacts/skill/test_package.py::test_directory_package_round_trips_exact_files_and_executable_mode` | The fixture uses platform-translating `write_text()`, but asserts archived LF bytes; it also expects POSIX executable bits. |
| `tests/builtin/runtime/test_external_skills.py::test_explicit_external_skill_import_captures_exact_snapshot_and_enters_review` | `chmod(0755)` on a Windows fixture is assumed to provide POSIX executable metadata. |
| `tests/test_cli.py::test_remote_enroll_can_install_automatic_service_in_one_command` | Windows `st_mode` is compared directly to POSIX `0600`. |
| `tests/test_openclaw_cli.py::test_build_openclaw_plugin_runs_pnpm_install_non_interactively` | The fake `pnpm` executable is a shebang script without a Windows launcher. |
| `tests/test_opencode_cli.py::test_run_opencode_probe_executes_request_waits_for_nonce_and_stops_process` | `_assert_process_stopped()` treats `os.kill(pid, 0)` as a POSIX liveness check and catches only `ProcessLookupError`. |
| `tests/test_opencode_cli.py::test_run_opencode_probe_handles_process_exit_and_stops_process` | Same subprocess assertion. |
| `tests/test_opencode_cli.py::test_run_opencode_probe_times_out_and_stops_process` | Same subprocess assertion. |
| `tests/test_service_bootstrap.py::test_unsafe_retry_state_exits_nonzero_and_removes_token[group-readable]` | A launchd/POSIX mode check runs on Windows, where the implementation deliberately does not apply POSIX permission bits. |

The first seven nodes and the parameterized retry-state test reproduced **8 failures and 1 platform skip** in an unmodified checkout.

[...]

PR #1595, opened 2026-09-14
test(windows): make skill snapshot tests portable

Lianminx:

Refs #1501. This addresses only the two Skill snapshot tests; the issue's other Windows portability cases remain open.

# Rationale for this change

On Windows, the directory fixture writes CRLF through `write_text()` but expects LF bytes, and the external-skill import test assumes that `chmod(0755)` supplies POSIX executable metadata. Both existing tests fail on unmodified master `3738c586`.

- Write the LF-sensitive fixture as explicit bytes and scope filesystem executable-bit assertions to POSIX platforms.
- Check the external snapshot's expected platform mode and compare every archived file against its source bytes.
- Exercise explicit ZIP modes `0644` and `0755` on every platform, including exact CRLF preservation, materialization, and canonical archive recapture. POSIX also checks the materialized mode.

No. Only two test files change; production behavior, package validation, and stored formats are unchanged. Neither existing round-trip test is skipped on Windows.

# How was this change tested?

On Windows, CPython 3.12.9:

- Before the change: the two issue-listed test nodes failed with the expected CRLF/LF and `0644`/`0755` mismatches.
- `uv run --locked pytest -q tests/builtin/artifacts/skill/test_package.py tests/builtin/runtime/test_external_skills.py`: **19 passed, 1 skipped**. The skip is the existing Unix symlink-swap test.
- `uv run --locked ty check tests/builtin/artifacts/skill/test_package.py tests/builtin/runtime/test_external_skills.py`: passed.
- `uv lock --check` and `git diff --check`: passed.
- `uv run --locked prek run -a`: all hooks passed except repository-wide `ty check`, which reported seven Windows-specific diagnostics in unchanged `artifact_processing.py` and `test_artifact_processing.py` (`PipeConnection` versus `Connection`, and `os.WNOHANG`). Both files match the upstream baseline.

On Ubuntu 20.04 under WSL, CPython 3.12.13:

- The same two-file pytest command: **20 passed** (including the POSIX executable-mode and symlink checks).

The PR's [CI quality job](https://github.com/oceanbase/powercontext/actions/runs/34848953431/job/103992040731), including repository-wide checks on Linux, also passed.

# AI usage statement

OpenAI Codex (GPT-6) assisted with diagnosis, implementation, diff review, and focused validation.


Teingi (approved):
LGTM

Issue #1556, opened 2026-09-10
feat(context): add a recall sufficiency gate with bounded expansion to prepare_context


AlexStocks:
## Feature description

Add an internal **recall sufficiency gate with bounded expansion** to the Runtime `prepare_context` operation.

Today `prepare_context` performs exactly one search over the participating artifact families (Memory, and since RFC 1489 also Experience / Profile snapshot / Topic Memory), then fits the hits into the caller's `max_bytes` budget and renders the final string. When that single search under-retrieves, the operation still returns normally — it simply delivers fewer items, or a normal empty result.

This issue proposes that the Runtime, *before* rendering, cheaply assess whether the candidate set is sufficient for the query and, when it is not, run at most `N` controlled expansion rounds — widen the candidate pool, relax the time window, or admit one more family — and then re-select within the **same** caller-provided total budget.

The public contract stays exactly as it is: `PreparedContext(schema, status, content, content_bytes)`.

## Problem and proposed solution

**Problem.** RFC 0028 deliberately collapses internal reasons (no memory, no match, error) into one normal empty result for the caller, and that decision should stand. Collapsing the *reported reason* is a different matter from having no *recovery path at all*. Concretely, a query whose relevant evidence sits just outside the first recall round — a differently worded entry, an entry outside the default time window, or evidence that lives in Experience rather than Memory — currently receives a silently thinner context, and every integration observes the same thinness.

RFC 1489 gave callers control over which families participate, their order, and a per-family item limit. It did not address what the Runtime should do when the participating families together return too little to be useful.

**Proposed solution.** A two-stage pipeline inside `prepare_context`:

```text
Stage A  recall
           └─ RetrievalSufficiencyGate.assess(candidates, query)
                 sufficient            -> proceed to Stage B
                 insufficient, r < N   -> expand (bounded) and re-recall
                 insufficient, r == N  -> proceed to Stage B with what we have
Stage B  select + section assembly + budget water-filling + render
Stage C  report omission / effort (in-process trace only)
```

Properties I believe this must have to fit the existing design:

1. **Gate signals must be model-free by default.** RFC 1489 states plainly that assembly adds no model calls. Candidate cheap signals: candidate count, score distribution and top-1/mean separation, lexical overlap between query and top hits, family coverage, and evidence time span. If a deployment has RFC 0080 rerank enabled, an expansion round that re-enters search may add one generation call — that cost should be explicit and configurable, not silent.
[...]

PR #1596, opened 2026-09-14
feat(runtime): add a bounded recall sufficiency gate to prepare_context

AlexStocks:

Implements the in-process recall sufficiency gate described by [RFC 1560 `recall_sufficiency_gate`](https://github.com/oceanbase/powercontext/blob/master/docs/en/rfcs/1560_recall_sufficiency_gate.md) (merged as #1560).

Related: #1556. This PR deliberately does **not** write `Closes #1556`: the runtime implementation, observability sink and local regression coverage are included here, but the RFC's workload/OceanBase acceptance run and default-threshold calibration are still follow-up work before the feature should be enabled by default.

# Rationale for this change

`prepare_context` previously performed one recall pass per selected family, then selected and rendered whatever survived admission and fitting. If recall under-retrieved, the call still returned normally with a thin or underfilled context.

RFC 1560 adds a deterministic, model-free sufficiency gate between recall and rendering. When the first candidate set is thin, the runtime may run at most two bounded expansion rounds. Each expansion round keeps the caller-selected families, limits and search mode unchanged; its only action is lowering the internal admission floor so more already-retrievable evidence can compete for the final Builder selection.

The feature remains disabled by default.

# What changes are included?

**Pure gate module.** Adds `src/powercontext/builtin/runtime/recall_sufficiency.py` with `RecallSufficiencyPolicy`, `RecallSufficiencyGate`, `RecallExpander`, `RecallSignals`, `GateAssessment`, `RecallBudgetView`, `SearchPlan`, `RecallCandidate`, `RecallEffort` and helpers. The gate, expansion planner and candidate projection are pure: no I/O, no clock and no model call.

**Admission-floor plumbing.** Adds internal `AdmissionFloor` / `AdmissionCounts` values in `artifacts/search.py`. The default floor preserves the existing thresholds: lexical coverage `0.25`, minimum matched terms `2`, and semantic similarity `0.3`. `admission=None` is the historical path. The optional admission override is threaded through Memory, Topic Memory and Experience search paths.

**Bounded runtime expansion.** `ScopedContextApplication._prepare_build` now returns `(PreparedContextBuild, RecallEffort | None)`. With the gate enabled it performs round 0, probes the Builder budget, assesses sufficiency, and then runs up to `recall_gate_max_rounds` extra recall rounds. Candidates accumulate by family-specific evidence identity; later rounds can only add new identities, never remove an earlier candidate.

**Observability and cost accounting.** `PreparedContextBuild` no longer carries `recall_effort`. Instead, `open_builtin_runtime(..., recall_effort_sink=...)` can receive the final `RecallEffort`, mirroring the existing estimator/sink style. Sink failures are logged and fail open. The effort reports RFC-style fields including `assessment`, `expansion_actions`, `rounds` as search-pass count, `candidates_by_round`, `admission_by_family`, `added_embeddings`, `added_generation_calls`, `truncated_items`, `dropped_items`, `dropped_below_min_bytes` and `dropped_no_fitting_truncation`.

**Budget probe.** `PreparedContextBuilder.probe_budget(...)` runs the Builder's own pure selection/fitting path and returns aggregate budget pressure. The gate uses this view instead of a static `request.max_bytes <= 512` shortcut.

**Query-vector reuse.** Expansion rounds reuse round-zero query embeddings when the profile matches. Memory and Topic Memory use separate per-scope reuse caches so vectors are not shared across families.

**Aggregate tracing only.** `context.build` records only aggregate recall counters allowed by the RFC boundary: recall rounds, assessment, truncated item count and dropped item count. Per-family admission and cost details go to the optional in-process sink, not HTTP or persistence.

**Fail-open behavior.** Gate assessment, expansion and sink delivery failures cannot fail `prepare_context`. Gate/expansion errors return the round-zero candidates with `assessment="expansion-failed"`; sink errors are logged and the prepared context is still returned.

**Configuration.** Adds disabled-by-default runtime config fields for enabling the gate, setting max rounds, sufficiency thresholds, admission floors and rerank behavior.

# Scope and remaining work

This PR does not change `openapi/powercontext.yaml`, checked-in HTTP generated code, persisted schemas, public `PreparedContext`, public topic-memory search request/response shape, or default runtime behavior.

This PR also does not complete the RFC's acceptance/calibration work:

- The RFC 1229 workload comparison and RFC 0081 harness have not been run here.
- OceanBase live acceptance has not been run here.
- Default thresholds are intentionally not claimed as calibrated.
- #1556 should remain open until maintainers decide the rollout/evaluation criteria are satisfied.

# Calibration warning

The feature is disabled by default because the thresholds still need workload calibration. In particular, reciprocal-rank fused Memory scores can make the family-local `top - mean` gap small even when the hits are relevant, so enabling the gate without calibration can trigger extra bounded rounds. That is expected to be resolved through the RFC's unresolved-questions/evaluation path, not by silently enabling the feature in this PR.

# How was this tested?

Local validation on Windows from the PR worktree:

```text
uv run ruff check src tests
# All checks passed!

uv run ruff format --check src tests
# 579 files already formatted

uv run ty check src/powercontext/builtin/runtime/application.py src/powercontext/builtin/runtime/composition.py src/powercontext/builtin/runtime/relational.py src/powercontext/builtin/artifacts/memory/__init__.py src/powercontext/builtin/artifacts/topic_memory/models.py tests/e2e/test_recall_sufficiency_gate.py tests/e2e/test_builtin_runtime.py tests/e2e/test_context_text_assembly.py tests/builtin/artifacts/topic_memory/test_fusion.py tests/builtin/persistence/test_records.py
# All checks passed!

uv run pytest tests/builtin/artifacts/memory tests/builtin/artifacts/topic_memory tests/builtin/persistence/test_experience_index.py tests/builtin/persistence/test_records.py tests/builtin/persistence/test_provider.py tests/builtin/runtime/test_prepared_context.py tests/builtin/runtime/test_recall_sufficiency.py tests/builtin/runtime/test_topic_memory_application.py tests/builtin/runtime/test_topic_memory_processing.py tests/e2e/test_recall_sufficiency_gate.py tests/e2e/test_builtin_runtime.py tests/e2e/test_context_text_assembly.py tests/e2e/test_observability.py -q
# 282 passed, then 2 scheduled-worker observability tests failed in the long combined run

uv run pytest tests/e2e/test_observability.py::test_scheduled_family_worker_starts_an_independent_trace_root -q --tb=short
# 2 passed in 20.56s
```

The two long-run failures were:

- `tests/e2e/test_observability.py::test_scheduled_family_worker_starts_an_independent_trace_root[memory]`
- `tests/e2e/test_observability.py::test_scheduled_family_worker_starts_an_independent_trace_root[experience]`

Both passed when rerun directly. They exercise scheduled artifact-processing trace export rather than the recall-gate path, so they are recorded as local Windows timing/long-suite risk rather than fixed in this PR.

A broader `uv run pytest -q --tb=short --maxfail=20` was also attempted on Windows earlier and did not go green because of existing out-of-scope Windows environment/POSIX assumptions, including CRLF-sensitive package tests, process-spawn internals, systemd tests, POSIX mode/symlink checks and WSL/bash `python` lookup. Those are not changed by this PR.

# AI usage statement

Written with Codex/WorkBuddy under human direction. The RFC and current code paths were checked against the repository before implementation, the WIP PR head was repaired without rewriting existing commits, and the validation commands above were run locally against the resulting branch.


src/powercontext/builtin/persistence/topic_memory.py
```diff
@@ -442,12 +446,20 @@ async def search(
             mode=used_mode,
             query_vector=query_vector,
             embedding_profile=embedding_profile,
+            admission=admission,
         )
         channels = await self.index.search(connection, scope_id, request)
         await self._check_retrieval_shape(connection)
+        outcome = _fuse_topic_memory_rankings(query, channels, limit, mode=used_mode, admission=admission)
```
Teingi, line 453: [P2] Preserve pre-admission counts so FTS expansion can run

Both FTS indexes apply the coverage floor before returning these channels, so the fusion counters cannot see rejected candidates. With a Topic Memory detail containing `alpha` and query `alpha beta gamma`, round zero reports `retrieved=0, admitted=0`. The enabled gate then returns an empty context with `assessment=no-content` and `rounds=1`, although a direct search with the round-one floor returns that Topic. Please carry pre-admission accounting through the index boundary so the gate can distinguish recoverable FTS candidates from an empty result pool.

src/powercontext/builtin/runtime/application.py
```diff
@@ -785,31 +961,281 @@ async def _prepare(self, request: PrepareContextRequest, scope: ScopeDescriptor,
...
+        round's candidate is never removed. Any error returns the round-zero candidates unchanged
+        and records ``expansion-failed`` — the gate can never turn a successful prepare into a
+        failure. For a consistent signal basis, ``candidates_by_round`` counts the *un-truncated*
+        accumulated pool (no family is clamped while the gate is consulting it); the Builder
+        ceilings are applied once, on the returned candidates only.
+        """
+
+        gate = RecallSufficiencyGate()
+        expander = RecallExpander()
+        families_expected = _families_with_recoverable_candidates(families, round_zero.admissions)
```
Teingi, line 1023: [P2] Separate recoverability from expected family coverage

This counts only families with rejected candidates, while `signals.family_count` counts all families with admitted hits. With four fully matching Memory hits and one approved Experience matching only `alpha` in `alpha beta gamma`, the counts are Memory `4/4` and Experience `1/0`. `families_expected` becomes 1, so the Memory results satisfy the coverage check and the gate reports `sufficient` without expansion, even though the round-one floor retrieves the missing Experience. Please track expected family coverage separately from whether any family has recoverable candidates.

src/powercontext/builtin/runtime/application.py
```diff
@@ -785,31 +961,281 @@ async def _prepare(self, request: PrepareContextRequest, scope: ScopeDescriptor,
...
+                    topic_memory_hits=tuple(accumulated_topic),
+                    experience_hits=_flatten_scope_experience(experience_hits_by_scope, scope_ids),
+                )
+                candidates_by_round.append(len(candidates))
+                assessment = gate.assess(
+                    candidates,
+                    request.query,
+                    policy,
+                    scope_has_content=bool(candidates) or families_expected > 0,
+                    budget=budget,
```
Teingi, line 1112: [P2] Refresh the budget view after expansion

This reassessment still uses the round-zero budget. In a SQLite Memory-only request with `max_bytes=1000`, one full-query hit leaves 361 bytes free; round one admits a long single-term hit and fills all 1000 bytes, with one truncated item. A fresh Builder probe reports `budget_bounded=True`, but the stale view allows another search, producing `rounds=3` and candidate counts `(1, 2, 2)`. Please probe the accumulated candidates after applying the family ceilings before deciding to expand again, so a budget-bound result stops the additional recall work.

Teingi (commented):
Three P2 issues reproduced on this commit with isolated SQLite databases and the default gate thresholds. The focused test suite passes (230 tests), but these cases remain uncovered.

Teingi:
resolve conflicts

src/powercontext/builtin/persistence/topic_memory.py
AlexStocks: 已将 FTS 候选获取与 admission 判定分离。后端保留 MATCH 候选和每个 Topic 的最佳 coverage chunk，fusion 层统一应用 floor；默认结果保持不变，retrieved/admitted 现在能报告可恢复候选。

src/powercontext/builtin/runtime/application.py
AlexStocks: 已按 round-zero 实际检索到候选的 family 计算 expected coverage，不再只计算被当前 floor 拒绝的 family；已准入的 Memory 不会掩盖可在下一轮恢复的 Experience。

src/powercontext/builtin/runtime/application.py
AlexStocks: 每轮 expansion 合并候选并应用 family cap 后，都会重新调用 PreparedContextBuilder.probe_budget，再决定是否继续扩展。预算已绑定时会停止后续搜索。

src/powercontext/builtin/persistence/topic_memory.py
Teingi: [P2] Preserve default recall when moving admission after retrieval

On `eeb78c24`, removing the SQL coverage filter makes the [FTS query](https://github.com/oceanbase/powercontext/blob/eeb78c24fd2e487325c37ade5123778e893f507f/src/powercontext/builtin/persistence/sqlite/topic_memory_index.py#L149-L153) apply its candidate limit before admission. I reproduced this with 60 single-term Topics and one two-term Topic with a longer summary, querying `alpha beta gamma`: `f32b6efa` returns the valid two-term Topic, while the current head returns no hits (`retrieved=20, admitted=0`). The higher-ranked single-term matches fill the pool and are then all rejected by fusion. This also affects default search with the gate disabled. Could we preserve the eligible candidate pool for default admission and collect pre-admission counts separately?

src/powercontext/builtin/runtime/application.py
Teingi: [P2] Keep recoverability separate from expected family coverage

On `eeb78c24`, `families_expected` includes every family with retrieved candidates, but the [expansion condition](https://github.com/oceanbase/powercontext/blob/eeb78c24fd2e487325c37ade5123778e893f507f/src/powercontext/builtin/runtime/application.py#L1071) still uses it to decide whether another search can help. With one Memory containing `alpha beta gamma evidence` and query `alpha beta gamma`, SQLite reports `retrieved=admitted=1`. The previous head performs one search; the current head performs three, with candidate counts `(1, 1, 1)` and unchanged output. Lowering admission cannot recover anything from this fully admitted pool. Could we retain the corrected family-coverage count and add a separate recoverability check before expanding?

AlexStocks:
> resolve conflicts

done

src/powercontext/builtin/persistence/topic_memory.py
Teingi: Rechecked `eeb78c24` with a fresh SQLite database: the default-search regression still reproduces. The FTS search returns no hits (`retrieved=20, admitted=0`), although the dataset contains a Topic that meets the default two-term coverage rule. This also affects the gate-disabled path. The original expansion counterexample passes; preserving the default eligible candidate pool remains an outstanding P2.

src/powercontext/builtin/runtime/application.py
Teingi: Rechecked `eeb78c24` with a fresh SQLite database: one fully matching Memory still causes three search passes with candidate counts `(1, 1, 1)`, despite `retrieved=admitted=1` and unchanged output. The original missing-family counterexample passes, but the separate recoverability check is still needed to avoid these two redundant searches. This P2 remains reproducible on the current head.

src/powercontext/builtin/persistence/topic_memory.py
Teingi: September 17 recheck on `eeb78c24`: this P2 still reproduces with a fresh SQLite database. The default FTS search returns no hits (`retrieved=20, admitted=0`), even though one Topic satisfies the two-term coverage rule. Single-term matches fill the candidate limit before admission, excluding the eligible Topic. This affects the gate-disabled path as well. Please preserve the default eligible candidate pool while collecting pre-admission counts separately; the thread's resolved state does not match the current behavior.

src/powercontext/builtin/runtime/application.py
Teingi: September 17 recheck on `eeb78c24`: this P2 also remains reproducible with a fresh SQLite database. One fully matching Memory (`retrieved=admitted=1`) still triggers three search passes, with candidate counts `(1, 1, 1)` and unchanged output. The original missing-family case passes, but a separate recoverability check is still needed before expansion to avoid the two redundant searches. Please address this remaining case before treating the thread as resolved.

AlexStocks:
@Teingi please review it again.

Teingi:
Re-reviewed the current head, `eeb78c24`, and reran the previous counterexamples with fresh SQLite databases. The original three fixes still pass, but these two P2 regressions remain:

1. **Default Topic Memory search loses eligible results.** With 60 single-term Topics and one valid two-term Topic, querying `alpha beta gamma` returns no hits (`retrieved=20, admitted=0`). Single-term matches fill the candidate limit before admission, excluding the eligible Topic. This also affects the gate-disabled path. Please preserve the default eligible candidate pool and collect pre-admission counts separately. [Details](https://github.com/oceanbase/powercontext/pull/1596#discussion_r4032890952).

2. **Expansion runs when no candidates can be recovered.** One fully matching Memory (`retrieved=admitted=1`) still triggers three search passes with candidate counts `(1, 1, 1)` and unchanged output. Please keep expected family coverage separate from the recoverability check used to decide whether to expand. [Details](https://github.com/oceanbase/powercontext/pull/1596#discussion_r4032892429).

CI is green, but both cases still reproduce on the current head. I recommend addressing them before merging. This local recheck covered SQLite; OceanBase was not rerun locally.

src/powercontext/builtin/persistence/topic_memory.py
AlexStocks: Fixed in `cccad642`.

The public Topic Memory search limit remains 20, but the repository now requests a separate bounded internal channel pool (`max(limit * 4, 32)`, capped at 100) before fusion-time admission. That keeps default search from losing an eligible two-term Topic when single-term FTS matches fill the public limit, while `retrieved` still reflects pre-admission channel hits.

Added a SQLite regression with 60 single-term Topics plus one default-eligible two-term Topic; the default gate-disabled search now returns the eligible Topic and reports `retrieved > 20`.

Validated locally with:
`uv run pytest -q tests/builtin/persistence/test_topic_memory.py tests/builtin/persistence/test_topic_memory_oceanbase.py tests/builtin/artifacts/topic_memory/test_fusion.py tests/builtin/runtime/test_recall_sufficiency.py tests/builtin/runtime/test_prepared_context.py tests/e2e/test_recall_sufficiency_gate.py` (`167 passed, 1 skipped`).

src/powercontext/builtin/runtime/application.py
AlexStocks: Fixed in `cccad642`.

`families_expected` now remains the round-zero family coverage signal (`retrieved > 0`), while expansion uses a separate recoverability check that requires `retrieved > admitted`. A fully admitted family can still contribute to family coverage, but it no longer allows extra rounds when lowering admission cannot recover any candidates.

Added an E2E regression for one fully matching Memory (`retrieved=admitted=1`): the gate now stays at `rounds=1`, `expansion_actions=()`, and `candidates_by_round=(1,)`.

Validated locally with:
`uv run pytest -q tests/builtin/persistence/test_topic_memory.py tests/builtin/persistence/test_topic_memory_oceanbase.py tests/builtin/artifacts/topic_memory/test_fusion.py tests/builtin/runtime/test_recall_sufficiency.py tests/builtin/runtime/test_prepared_context.py tests/e2e/test_recall_sufficiency_gate.py` (`167 passed, 1 skipped`).

AlexStocks:
@Teingi current head \163a4436\ has addressed the latest SQLite regressions and the follow-up CI model-bound failure. There is no merge conflict, no current unresolved review thread, and all visible checks are green. Please continue review when you have time.

Teingi:
Re-reviewed `163a4436` with fresh SQLite databases. The fully admitted round-zero case is fixed: it now performs one search. The original three fixes also still pass. Two P2 issues remain:

1. **Default Topic Memory search still loses eligible results** ([code](https://github.com/oceanbase/powercontext/blob/163a4436c40a04722dd8c122e08acbdcc6f4fc78/src/powercontext/builtin/persistence/topic_memory.py#L446)). The original fixture—60 single-term Topics plus one two-term Topic with a longer summary—still returns no hits for `alpha beta gamma` with `limit=8` (`retrieved=32, admitted=0`). With 120 single-term Topics and `limit=20`, it returns `80/0`; `f32b6efa` returns the eligible Topic on the same dataset. Increasing the pool size moves the cutoff but still allows rejected matches to crowd out eligible results. This affects the gate-disabled path too. Please preserve eligible default-search candidates before truncating the pool and collect pre-admission counts separately.

2. **Recoverability is not refreshed after expansion** ([code](https://github.com/oceanbase/powercontext/blob/163a4436c40a04722dd8c122e08acbdcc6f4fc78/src/powercontext/builtin/runtime/application.py#L1034)). `families_recoverable` is computed only from round zero. With one Memory containing `alpha evidence` and query `alpha beta gamma`, admission changes from `1/0` to `1/1` after the first expansion, but the gate still runs a third search: candidates `(0, 1, 1)`. Allowing one versus two expansions produces identical output. Please recompute recoverability from each round's `issued.admissions` before deciding whether to continue.

Validation: 235 related tests passed, and CI is green. The counterexamples above still reproduce, so I recommend addressing them before merging. Local validation covered SQLite; OceanBase was not rerun locally.

AlexStocks:
@Teingi fixed in `bca57fbc`.

- Default Topic Memory FTS now applies the admission floor before truncating the backend candidate pool, while carrying separate pre-admission retrieved counts through `TopicMemorySearchChannels`. This preserves the eligible two-term Topic in the crowded default-search case and keeps the recall trace counts separate.
- Recall expansion now recomputes recoverability from each issued round's admissions before deciding whether to continue, so a round that changes `retrieved/admitted` from `1/0` to `1/1` stops after that expansion instead of running a redundant third search.

Validation on this head:

- `168 passed, 1 skipped` for the related Topic Memory / recall-sufficiency suites.
- `78 passed, 1 skipped` for the focused Topic Memory model/fusion/persistence/e2e subset.
- ruff check, ruff format --check, ty check, and git diff --check passed on the changed files.
- All visible GitHub checks are green on `bca57fbc`.

src/powercontext/builtin/artifacts/topic_memory/fusion.py
```diff
@@ -52,20 +53,53 @@ def fuse_topic_memory_rankings(
...
+        ("topic_fts", channels.topic_fts),
+        ("topic_vector", channels.topic_vector),
+        ("detail_fts", channels.detail_fts),
+        ("detail_vector", channels.detail_vector),
     )
     enabled_channels = _MODE_CHANNELS[mode]
+    retrieved = sum(
+        _retrieved_count(channels, channel, hits) for channel, hits in channel_inputs if channel in enabled_channels
+    )
+    admitted = sum(len(hits) for channel, hits in rankings if channel in enabled_channels)
```
Teingi, line 102: [P2] Separate candidate truncation from admission rejection

`retrieved` now includes all matching FTS rows, while `admitted` counts only the rows left after the backend candidate limit. Their difference therefore includes eligible rows discarded by the cap, but `_families_with_recoverable_candidates` treats it as evidence that lowering admission can recover candidates.

Reproduced on `bca57fbc` with SQLite: create 40 Topics titled `alpha beta`, with summaries `Useful evidence {i}` and unrelated details; prepare Topic Memory only with query `alpha beta gamma delta epsilon`, section limit 8, `max_bytes=8000`, and the gate enabled with default thresholds. All 40 Topics satisfy the default admission floor, yet the counts are `retrieved=40, admitted=32`. The gate performs three searches with candidate counts `(8, 8, 8)` and no additional candidates. The same fixture with 32 Topics performs one search.

Please measure admission rejection separately from candidate truncation, and use comparable counts for recoverability. Add this over-cap, fully eligible case as a regression test so lowering the floor does not trigger two redundant searches.

Teingi (commented):
Re-reviewed `bca57fbc` with fresh SQLite databases. The previously reported counterexamples now pass, including the crowded Topic Memory search and stopping after the first successful Memory expansion. One P2 remains in the Topic Memory admission accounting, detailed inline.

Validation: 258 passed, 1 skipped; Ruff lint and formatting checks passed. Local validation covered SQLite; live OceanBase was not rerun. I recommend addressing this remaining case before merging.

src/powercontext/builtin/artifacts/topic_memory/fusion.py
AlexStocks: Fixed in `87d98962`.

The Topic Memory admission accounting now carries an explicit `rejected` count for same-search admission-floor rejects. FTS backends separately report the unbounded eligible count, so `retrieved=40` and a capped delivered pool no longer makes recoverability look positive when all candidates already satisfy the default floor. `_families_with_recoverable_candidates()` now uses `rejected` when present and only falls back to `retrieved - admitted` for families whose counts remain comparable.

Added the over-cap regression with 40 eligible Topic Memories; the gate now performs one search with `candidates_by_round == (8,)` instead of redundant expansion.

Local validation:

- `170 passed, 1 skipped` for the related Topic Memory / recall-sufficiency suites.
- ruff check, ruff format --check, git diff --check, and scoped `ty check` passed on the changed files.
- Full `ty check` still reports pre-existing unrelated diagnostics in `artifact_processing.py`, `opencode.py`, and `test_artifact_processing.py`; none point at this change.

GitHub checks are running on `87d98962`.

Teingi (approved):
LGTM

Issue #1300, opened 2026-08-21
feat: add opt-in import of pre-install user prompts from one agent host


knqiufan:
### Feature description

PowerContext only captures user prompts **after** its plugin is installed. Older prompts from Codex, Claude Code, or DeepSeek Harness stay in that host's local session files. A project that already has months of history has to start from zero.

Add an explicit, default-off CLI that copies **only user prompts** from **one** host into the existing Source pipeline:

```text
powercontext import-sessions --host codex
powercontext import-sessions --host claude-code
powercontext import-sessions --host dsh
```

`--host` is required. This does not change Memory contracts, does not read full transcripts in v1, and does not run from a prompt hook.

This is not "inject history into the agent". It is a one-time or incremental backfill **into PowerContext**, so later `prepare` can use knowledge the user already wrote.

#### Why this issue exists

The official loop is:

```text
user prompt → Source → optional extract → Memory → later prepare
```

That loop starts only after `setup` and a running Server. Live hooks capture the current user prompt as `POST /v1/sources/content`. They never read host session files created before install. Scheduled extract cannot invent Sources that were never captured.

So the first day of PowerContext looks empty even when the user has already told Codex or DSH the project's decisions, constraints, and facts.

#### Gap

| Today | Missing |
| --- | --- |
| Hook captures the current user prompt | No reader for host session files created before install |
| `POST /v1/sources/content` and `flush` already exist | No official command that feeds old prompts into those APIs |
| Scope is per project and shared across hosts | Old prompts are stranded in host-specific files |

#### Value

- Users can reuse decisions, constraints, and facts already stated to an agent, instead of re-teaching PowerContext from the install day.
- Later recall still uses the current `prepare` path and the same project scope.
- Backfill stays explicit and host-scoped, so a Codex-only user does not have DSH or Claude files scanned by default.

#### Proposed user contract

Keep `server run` and live hooks unchanged. Add a client-side importer.

`--host` should stay required:

- Session paths and formats differ by host.
- Live capture already uses host-specific `source_id` and `origin` (`codex-user-prompt:`, `claude-code-user-prompt:`, `dsh-user-prompt:`).
- Users often want only one host. Scanning every host by default would be surprising.

Suggested flow:

```text
import-sessions --host <codex|claude-code|dsh>
  → host adapter reads local user prompts
  → derive the same scope as the plugin (cwd / git remote / override)
  → POST /v1/sources/content
  → optional flush
```

Rules:

[...]

PR #1602, opened 2026-09-14
feat: import codex sessions

curryxjh:

Part of #1300.
Refs #1574.

Closes #.

# Rationale for this change

This PR adds the first opt-in import path for pre-install agent session history, starting with Codex. It lets users import historical Codex user prompts as ordinary Content Sources so existing prompt history can participate in PowerContext processing without bypassing the Source pipeline.

The implementation follows the constraints from #1574: Codex-only, explicit/default-off, user prompts only, no assistant replies, no direct Memory writes, no default Scope fallback, and idempotent overlap with live Codex prompt capture.

- Adds `powercontext import-sessions --host codex`.
- Reads Codex JSONL session files from `CODEX_HOME` or `~/.codex`.
- Imports only real user prompts and skips synthetic environment context and assistant messages.
- Resolves Scope using Codex session/workspace bindings, or imports into `--scope-id` when provided.
- Writes prompts through the existing Content Source capture API.
- Uses the same Codex source identity and metadata shape as the live Codex hook to keep repeated imports and live/import overlap idempotent.
- Adds `--dry-run`, `--flush`, and `--checkpoint-file`.
- Records per-item accepted/skipped/failed results and keeps a persistent checkpoint for resume.
- Skips malformed session files while reporting file-level failures.
- Updates English and Chinese Sources documentation.
- Adds focused tests for reader behavior, scope resolution, checkpoint resume, dry-run, bad files, CLI JSON output, and live capture overlap.

Yes. A new explicit CLI command is available:
``` bash
powercontext import-sessions --host codex
```

The command is not run automatically. It only imports when the user invokes it. It writes ordinary Content Sources only; Memory extraction is requested only when --flush is passed.
There are no breaking API changes and no changes to existing public persisted formats.

# How was this change tested?

Added and updated tests in tests/test_session_import.py.
Commands run:
``` bash
UV_CACHE_DIR=/private/tmp/powercontext-uv-cache uv run pytest tests/test_session_import.py
UV_CACHE_DIR=/private/tmp/powercontext-uv-cache uv run pytest tests/codex_plugin/test_recall.py -k capture
UV_CACHE_DIR=/private/tmp/powercontext-uv-cache uv run ruff check src/powercontext/client/session_import.py src/powercontext/client/cli.py tests/test_session_import.py
UV_CACHE_DIR=/private/tmp/powercontext-uv-cache uv run ruff format --check src/powercontext/client/session_import.py src/powercontext/client/cli.py tests/test_session_import.py
UV_CACHE_DIR=/private/tmp/powercontext-uv-cache uv run ty check src/powercontext/client/session_import.py src/powercontext/client/cli.py tests/test_session_import.py
UV_CACHE_DIR=/private/tmp/powercontext-uv-cache uv run powercontext import-sessions --help
```

# AI usage statement

Implemented with assistance from OpenAI Codex.


src/powercontext/client/session_import.py
```diff
@@ -0,0 +1,632 @@
...
+    if not content:
+        return _PromptProcessingResult(
+            _item_result(prompt, SessionImportItemStatus.SKIPPED, reason="empty_after_redaction")
+        )
+    if len(content) > _MAX_CONTENT_LENGTH:
+        return _PromptProcessingResult(_item_result(prompt, SessionImportItemStatus.SKIPPED, reason="too_large"))
+    resolved_scope_id = scope_id or await _resolve_codex_scope(client, prompt)
+    if resolved_scope_id is None:
+        return _PromptProcessingResult(_item_result(prompt, SessionImportItemStatus.SKIPPED, reason="unresolved_scope"))
+    source_id = prompt.source_id_for_scope(resolved_scope_id)
```
AlexStocks, line 264: [P1] 修复 redaction 后的 live/import 重叠冲突
这里的 `source_id` 仍由原始 `prompt.content` 生成，但下面真正写入的是 `redact_known_secrets(prompt.content)`。live Codex hook 用同一组 `scope_id/session_id/turn_id/prompt` 生成相同 `source_id`，却提交原始 prompt；如果同一 prompt 已被 live capture 写入且包含会被 redaction 替换的 secret，导入器会用同一个 `source_id` 提交 `[REDACTED]` 后的不同 content，服务端会返回 `409 source_conflict`，该项会被记为 `capture_failed`，与 PR 描述中的 live/import overlap idempotent 不一致。建议让 identity 和实际写入内容使用同一份规范化文本，或显式处理这个 redaction conflict，并补一个“已存在 raw live capture + 导入 redacted prompt”的回归测试。

src/powercontext/client/session_import.py
```diff
@@ -0,0 +1,632 @@
...
+async def _process_prompt(
+    client: SessionImportClient,
+    prompt: ImportedPrompt,
+    *,
+    scope_id: str | None,
+    checkpoint: dict[str, object],
+    checkpoint_file: Path,
+    dry_run: bool,
+) -> _PromptProcessingResult:
+    content = redact_known_secrets(prompt.content).strip()
```
Teingi, line 254: [P1] Redact credentials from the selected Codex home

`redact_known_secrets()` reads `auth.json` from the environment's `CODEX_HOME` (or `~/.codex`), so it misses credentials when `--codex-home` points elsewhere. I put a synthetic token in the selected directory's `auth.json` and a historical prompt; the SQLite-backed import stored the token unchanged. Please pass the resolved Codex home into credential lookup so the directory being imported also supplies the known secrets to redact.

src/powercontext/client/session_import.py
```diff
@@ -0,0 +1,632 @@
...
+            payload = record.get("payload")
+            if not isinstance(payload, dict):
+                continue
+            typed_payload = cast(dict[str, object], payload)
+            if record.get("type") == "session_meta":
+                session_id = _string(typed_payload.get("session_id")) or _string(typed_payload.get("id")) or session_id
+                cwd = _string(typed_payload.get("cwd")) or cwd
+                continue
+            if record.get("type") == "turn_context":
+                cwd = _string(typed_payload.get("cwd")) or cwd
```
Teingi, line 352: [P2] Recover turn IDs from historical session records

Some real historical Codex logs store the turn ID in `turn_context`/`task_started`, without either turn-ID field on the user message. This reader discards that ID and gives both messages `turn_id=None`. With the same prompt in two distinct turns, the first was imported and the second was skipped as `checkpoint_accepted`; its identity also differed from the live hook's identity with a turn ID. Please carry the enclosing turn identity into these messages before calculating the source ID.

src/powercontext/client/session_import.py
```diff
@@ -0,0 +1,632 @@
...
+        return direct
+    metadata = payload.get("internal_chat_message_metadata_passthrough")
+    if isinstance(metadata, Mapping):
+        return _string(metadata.get("turn_id"))
+    return None
+
+
+def _is_synthetic_codex_user_message(content: str) -> bool:
+    stripped = content.strip()
+    return stripped.startswith("<environment_context>") and stripped.endswith("</environment_context>")
```
Teingi, line 576: [P2] Exclude automatically injected user-context messages

Real Codex logs also contain user-role messages starting with `# AGENTS.md instructions`, followed by repository instructions and environment context. This predicate only excludes a standalone `<environment_context>` block, so those automatic messages are imported as user prompts. I confirmed both the real-log parsing and persistence with a synthetic SQLite fixture. Please identify and exclude the host-injected context messages as well.

src/powercontext/client/session_import.py
```diff
@@ -0,0 +1,632 @@
...
+        if item.status == SessionImportItemStatus.FAILED and item.reason is not None:
+            _count_skip(failed_by_reason, item.reason)
+        if processed.imported:
+            imported += 1
+        if processed.touched_scope_id is not None:
+            touched_scopes.add(processed.touched_scope_id)
+    flushed_scopes: list[str] = []
+    if flush and not dry_run:
+        for resolved_scope_id in sorted(touched_scopes):
+            await client.flush_memory(FlushMemoryRequest(scope_id=resolved_scope_id))
```
Teingi, line 227: [P2] Flush through the imported source position

One flush processes at most `source_window_limit` Sources (100 by default), but this call discards the returned cursor and immediately reports the Scope as flushed. With the window set to 2 and three imported prompts, a subsequent manual flush still processed the third Source. Large history imports will therefore leave a backlog after `--flush`. Please retain the capture positions and continue flushing until the cursor reaches the highest position imported into each Scope, handling failures or lack of progress explicitly.

src/powercontext/client/session_import.py
```diff
@@ -0,0 +1,632 @@
...
+        return _PromptProcessingResult(
+            _item_result(prompt, SessionImportItemStatus.SKIPPED, reason="empty_after_redaction")
+        )
+    if len(content) > _MAX_CONTENT_LENGTH:
+        return _PromptProcessingResult(_item_result(prompt, SessionImportItemStatus.SKIPPED, reason="too_large"))
+    resolved_scope_id = scope_id or await _resolve_codex_scope(client, prompt)
+    if resolved_scope_id is None:
+        return _PromptProcessingResult(_item_result(prompt, SessionImportItemStatus.SKIPPED, reason="unresolved_scope"))
+    source_id = prompt.source_id_for_scope(resolved_scope_id)
+    if _checkpoint_status(checkpoint, source_id) == SessionImportItemStatus.ACCEPTED.value:
```
Teingi, line 265: [P2] Keep pending flush work separate from capture checkpoints

Accepted checkpoint entries return without adding their Scope to `touched_scopes`. Importing without `--flush` and then rerunning with it therefore requests no extraction, despite the CLI suggesting that workflow. The same happens after a flush failure: I injected a 503 after successful capture, and rerunning with `--flush` made no second flush attempt. Please preserve pending extraction work independently of accepted Source writes so adding or retrying `--flush` can finish it without recapturing Sources.

src/powercontext/client/session_import.py
```diff
@@ -0,0 +1,632 @@
...
+    cwd: str | None = None
+    environment_cwd: str | None = None
+    relative_session_file = _relative_session_file(session_file, session_root)
+    prompts: list[ImportedPrompt] = []
+    try:
+        stream = session_file.open(encoding="utf-8")
+    except OSError as error:
+        raise SessionImportError.unreadable_codex_session(session_file) from error
+    with stream:
+        for line_number, line in enumerate(stream, start=1):
```
Teingi, line 339: [P2] Contain decoding failures to the affected session file

Invalid UTF-8 raises `UnicodeDecodeError` while iterating this stream, outside the error handling that turns bad files into `SessionImportError`. With one non-UTF-8 `.jsonl` file and one valid session, the entire scan aborted instead of reporting a failed file and continuing. Please handle decoding failures at the file boundary and preserve the normal file-level failure reporting.

Teingi (commented):
Reviewed `dd2a9635`. The 53 focused tests, Ruff checks, and type checks passed. Additional SQLite/HTTP probes and checks against local Codex session formats reproduced the issues below.

src/powercontext/client/session_import.py
Teingi: [P2] Recover pending flush work after session archival

Rechecked at `eab34c16`: adding or retrying `--flush` works while the session file remains under `sessions`. Pending positions are still rebuilt only from the current scan, though, which excludes `archived_sessions`. In a SQLite-backed HTTP probe, I imported three Sources without flushing, moved the session to `archived_sessions`, and reran with `--flush`. It reported zero flushed Scopes and left all three Sources unprocessed; a manual flush then started processing them. Please restore pending Scope/position targets directly from the checkpoint, independently of whether the original session file is still discoverable, and cover this archive-and-retry case.

src/powercontext/client/session_import.py
```diff
@@ -0,0 +1,729 @@
...
+                    continue
+                typed_payload = cast(dict[str, object], payload)
+                record_type = record.get("type")
+                if record_type == "session_meta":
+                    session_id = (
+                        _string(typed_payload.get("session_id")) or _string(typed_payload.get("id")) or session_id
+                    )
+                    cwd = _string(typed_payload.get("cwd")) or cwd
+                    continue
+                if record_type in {"turn_context", "task_started"}:
```
Teingi: [P2] Read task_started from the event payload

Codex logs encode this as `type: "event_msg"` with `payload.type: "task_started"`, so the top-level check never updates the turn ID. When a user message has no turn ID of its own and the new ID is supplied by this event, the reader retains the previous turn ID. A SQLite-backed HTTP probe with the same prompt in two turns produced `turn-a, turn-a`: only one Source was stored, and the second prompt was skipped as `checkpoint_accepted`. Please recognize the nested event and use that structure in the regression fixture, which currently puts `task_started` at the top level.

Teingi (commented):
Reviewed `eab34c16`. All 60 related tests, Ruff checks, formatting, and targeted type checks passed. Additional SQLite-backed HTTP probes reproduced the turn-identity issue below and the archive-and-flush recovery issue described in the existing checkpoint thread.

PsiACE (changes requested):
Please keep this PR focused on importing historical user prompts as Content Sources. Two changes should block merging:

- Remove the `--flush` loop from the importer. It consumes the Scope’s shared journal, potentially processing unrelated pending Sources, without a total processing budget. Extraction should remain a separate operation.
- Validate the checkpoint’s destination before trusting accepted entries. Switching servers with the same Scope and Source IDs can otherwise skip Sources that the destination has never received. Stored positions also belong to the original destination.

```text
Read local sessions
        |
        v
Filter + redact + resolve Scope
        |
        v
Validate checkpoint destination
        |
        v
Capture Sources + save receipts
        |
        v
Report results
```

Please document that scanning happens on the CLI machine and supports a remote Server. Also clarify that omitting `--flush` does not prevent configured background processing.

Session/time filters and finer limits can follow separately. Host adapter separation is tracked in [#1680](https://github.com/oceanbase/powercontext/issues/1680) rather than expanding this PR.

src/powercontext/client/session_import.py
Teingi: Verified on `bac9c31f` with the original SQLite-backed HTTP probe. The two identical prompts now retain `turn-a` and `turn-b` and persist as two distinct Sources. This finding is fixed.

src/powercontext/client/session_import.py
Teingi: Still reproducible on `bac9c31f` with the original SQLite-backed HTTP probe: import three Sources without flushing, move the session file to `archived_sessions`, then rerun with `--flush`. The result reports zero flushed Scopes, and all three Sources remain unprocessed. A manual flush then starts processing them.

Please recover pending Scope/position targets directly from the checkpoint, independently of the current session scan, and add an archive-and-retry regression test. The 60 related tests and targeted lint, formatting, and type checks pass; this P2 remains open.

PR #1603, opened 2026-09-15
test(windows): make CLI fixtures and permission checks portable

Lianminx:

Refs #1501. Follow-up to #1595; OpenCode process-exit assertions and long temporary paths remain outside this PR.

# Rationale for this change

The remote-enrollment test compares Windows file modes to POSIX `0600`, and the OpenClaw fixture launches an extensionless shebang script on Windows. Both fail on master `545b1609`. The group-readable retry-state case also expects a POSIX rejection that the implementation deliberately does not apply on Windows; it can enter the real service launcher instead.

- Check that enrollment writes its configuration on every platform, and retain the exact `0600` assertion on POSIX.
- Skip only the group-readable retry-state parameter on Windows, matching the production permission boundary. Other retry-state validation and token-cleanup tests continue to run.
- Give the fake pnpm a Windows `.cmd` launcher using the current Python interpreter. Write the Python fixture with explicit LF newlines for its POSIX shebang. Preserve real subprocess execution, the `CI=true` requirement, and the generated bundle assertion.

No. Only three test files change; production behavior and permission checks are unchanged.

# How was this change tested?

- Before the change, the enrollment and OpenClaw nodes failed on Windows with `0666` versus `0600` and `WinError 193`, respectively.
- The unmodified group-readable case entered the launcher and was stopped. A controlled probe with the launcher patched to return success confirmed that Windows accepts this mode and reaches the launcher, rather than returning the expected permission error.
- Windows, CPython 3.12.9: `uv run --locked pytest -q --tb=short tests/test_cli.py::test_remote_enroll_can_install_automatic_service_in_one_command tests/test_service_bootstrap.py tests/test_openclaw_cli.py` — **27 passed, 2 skipped** (POSIX group bits and `O_NOFOLLOW`).
- Linux (WSL Ubuntu 20.04), CPython 3.12.13: the same selection — **29 passed**, including both POSIX permission cases.
- `uv run --locked ty check tests/test_cli.py tests/test_service_bootstrap.py tests/test_openclaw_cli.py` — passed.
- `uv run --locked prek run -a` — all hooks except full-repository `ty` passed. The Windows type check reports the same seven existing errors in unchanged artifact-processing files (`PipeConnection` versus `Connection`, and `os.WNOHANG`), plus two deprecation warnings in unchanged OpenCode code.
- `git diff --check` — passed.

# AI usage statement

OpenAI Codex (GPT-6) assisted with investigation, implementation, validation, and this PR description.


Teingi (approved):
LGTM

PR #1605, opened 2026-09-15
Feat/memory write tracing

### LINKED ISSUE #1242: feat: trace the Memory write path
## Feature description

Add bounded tracing for source capture and the Memory flush pipeline.

`flush_memory` currently covers extraction, embedding, and persistence without showing which stage completed or failed.

Possible trace structure:

```text
powercontext capture_content_source
└── memory.capture

powercontext flush_memory
└── memory.flush
    ├── invoke_agent memory_extraction
    ├── embeddings <model>
    └── memory.commit
```

Record only bounded counts, latency, outcome, and error category. Do not record Source content, Memory content, prompts, responses, or vectors.

## Alternatives considered

Keep one span for the complete flush operation.

## Additional context

Part of #1214. Tracing remains optional and must not change Runtime behavior.

Oxidaner:

Closes #1242

# Rationale for this change

Add bounded tracing for Source capture and the Memory write pipeline so operators can distinguish extraction, embedding, and persistence latency or failures.

Add memory.capture tracing for Source capture.
Add memory.commit tracing around the atomic Memory and cursor transaction.
Record bounded counts, outcomes, and error types without exporting content, prompts, responses, vectors, or IDs.
Add exported span-tree coverage, no-op coverage, privacy checks, and commit rollback tests.
Update English and Chinese tracing documentation.

When tracing is enabled, operators see additional internal spans. Runtime behavior, APIs, and persisted formats are unchanged.

# How was this change tested?

Focused observability, tracing, scheduler, and persistence tests: 55 passed
ty check and Ruff checks passed.
Manual HTTP validation with InMemorySpanExporter confirmed the capture and flush span trees.
Full suite: 1402 passed, 19 skipped; 3 Linux-only systemd tests fail on macOS.

# AI usage statement
OpenAI Codex was used to inspect the issue, implement the tracing changes, add tests, update documentation, and validate the result.


src/powercontext/builtin/runtime/relational.py
```diff
@@ -1252,6 +1264,7 @@ async def get(
         triggers: BuiltinTriggers = _RelationalTriggers(
             services=services,
             lock=self._activation_locks.setdefault(scope, asyncio.Lock()),
+            tracing=self._tracing,
```
Teingi, line 1268: **[P2] Pass tracing through `process_memory()` as well**

`get()` now passes `self._tracing` to `_RelationalTriggers`, but `process_memory()` constructs another trigger without it. The latter is the entry point used by `process_family_invocation()`, so it drops the configured tracing even when the contexts were opened with `tracing=ServerTracing(...)`.

I reproduced this with SQLite and `InMemorySpanExporter`: cursor-only commits and cancellation followed by retry emit no `memory.commit` spans through `process_memory()`, while the same cases through `context.triggers.flush()` emit the expected spans. Persistence and rollback still work. Please pass `self._tracing` in the other constructor call too and cover this entry point in the tracing tests.

Oxidaner: done

Teingi (approved):
LGTM

PR #1550, opened 2026-09-10
feat: unify artifact family reads and registration

Zxf-xufeng:
# Which issue or RFC does this PR close?

Related design: #1549.

This PR implements standard Artifact read support for Topic Memory and deliberately keeps the existing Artifact family composition and write-side logic.

# Rationale for this change

The standard Artifact list route rejected `topic-memory`, so clients could not use the unified Artifact read API to build the memory showcase page. Topic Memory also needs published-only ordering and display metadata when it is listed.

# What changes are included in this PR?

- Add a read-only `ArtifactReadFamily` contract containing `topic-memory`; keep `BaseArtifactFamily` unchanged for existing writable Artifact operations.
- Enable `topic-memory` for `list_artifacts`, `get_artifact`, `list_artifact_revisions`, and `get_artifact_revision`.
- Add the Topic Memory list adapter with published-time descending order, signed cursor pagination, published-only filtering, and optional title, summary, publication time, and source-count metadata.
- Reuse the standard Artifact response for detail and revision reads; keep the existing Topic Memory-specific APIs unchanged.
- Wire the Topic Memory list adapter explicitly while preserving the current master Artifact family tuple and management-writer composition.
- Remove the proposed Artifact family registry and its related generic registration abstractions.
- Regenerate the checked-in OpenAPI models and schema.

This PR does not add Topic Memory create, replace, or tag support, and does not introduce a new general-purpose capability framework.

# Are there any user-facing changes?

- Topic Memory can be read through the four standard Artifact read interfaces.
- Topic Memory list results are published-only and ordered by publication time descending.
- Artifact collection items support optional display metadata for families that provide it.
- Existing Artifact write/tag behavior and Topic Memory-specific APIs remain unchanged.

# How was this change tested?

- `make contract-test`: 48 passed.
- Focused Artifact, Topic Memory persistence, and API contract tests: 89 passed.
- Topic Memory HTTP and access-control regression tests with inherited auth/dashboard overrides removed: 25 passed.
- `uv run ruff check ...` passed.
- `uv run ty check` passed.
- `git diff --check` passed.

# AI usage statement

Implementation, tests, and PR text were produced with OpenAI Codex assistance.

PR #1614, opened 2026-09-15
docs(rfc): complete topic memory generic Artifact APIs

Zxf-xufeng:

Extends the Topic Memory read support delivered in #1550.

# Rationale for this change

Topic Memory can be read through generic Artifact APIs but cannot be created, replaced, or tagged through them. Generic cross-Scope copying also omits its required publication and search state.

Chinese and English RFCs define manual writes, tags, filtered pagination, atomic cross-Scope publication, Scope authorization, concurrency with automatic generation, and acceptance criteria.

Design documentation only. No runtime changes.

# How was this change tested?

Checked both documents for matching API semantics and ran git diff --check.

# AI usage statement

Written with OpenAI Codex assistance.


Teingi (approved):
LGTM

PR #1615, opened 2026-09-15
feat(topic-memory): complete generic Artifact APIs

Zxf-xufeng:

Implements the design proposed in #1614. The RFC and implementation are intentionally separate draft PRs.

# Rationale for this change

PR #1550 made Topic Memory readable through the generic Artifact APIs, but generic manual writes, tags, filtered listing, and complete cross-Scope publication were still unavailable. These operations must preserve the family's active head, immutable revisions, provenance, chunks, and configured search indexes as one consistent state.

- Add strict Topic Memory Create/Replace request models and regenerate the Python HTTP contract and public exports.
- Prepare deterministic chunks and optional embeddings before the write transaction, with configured timeout/concurrency limits; atomically commit manual provenance, content, and retrieval projections through the family repository. Recheck ETags after preparation and at commit.
- Add Topic Memory tag reads/writes, mixed-family tag queries, and SQL-level all/any tag filtering before pagination. Bind normalized filters to signed cursors and preserve tags across content revisions.
- Publish exact historical revisions into an independent target identity with complete retrieval state. Preserve publication provenance, omit direct Sources and tags, and resolve completed retries before embedding. Concurrent requests create only one complete target.
- Apply Scope-level authorization, including administration of both Scopes for publication, without Artifact-owner hooks for Topic Memory.
- Upgrade the tag family CHECK constraint on startup: transactional SQLite table rebuild preserving rows, foreign keys and indexes; OceanBase constraint alteration.
- Add HTTP lifecycle, restart, access-control (builtin/Casbin), concurrent CAS/publication, inference timeout, index-failure rollback, and legacy-schema migration coverage.

Topic Memory gains the seven capabilities described in #1614. Manual input requires complete nonblank title/summary/detail and is not rewritten by a generation model. Vector deployments perform synchronous embedding before success. Replacing existing content or tags requires `scope.admin`; creation requires `scope.contribute`. Default tag queries now include Topic Memory.

Before upgrading existing databases, back up the database and pause old writers while the tag CHECK constraint is upgraded. Downgrades require evaluating compatibility with newly stored Topic Memory tags. No automatic repair of historically incomplete Topic Memory copies is attempted.

# How was this change tested?

- `make check`: passed (integration manifest, lock consistency, pre-commit checks, Ruff, and both type-check targets).
- `make contract-test`: passed, 48 tests; generated Python and JavaScript API checks passed.
- `pytest tests/e2e/test_topic_memory_generic_api.py -q`: 9 passed, covering SQLite FTS and vector-enabled writes, history, tags, publication, restart, migration rollback, two-instance races, storage-failure atomicity, and configured inference timeout.
- Extended regression suite: 239 passed, 26 skipped. Covered generic API/access/tags/base access, Scope and Topic Memory server APIs, publication, Topic Memory persistence/integrity/application/processing, and the hermetic product chain.
- OceanBase/SeekDB integration cases require external test instances and were skipped. In particular, the OceanBase tag CHECK upgrade has not been verified against a live OceanBase instance.

Tests clear inherited local server authentication/dashboard settings and use temporary databases. Product-chain tests use loopback-only test services.

# AI usage statement

OpenAI Codex assisted with codebase analysis, RFC translation, implementation, regression tests, and verification. The changes are submitted as a draft for maintainer review.


src/powercontext/builtin/persistence/tag_schema.py
```diff
@@ -0,0 +1,58 @@
...
+    if connection.dialect.name == "mysql":
+        clause = await connection.scalar(
+            text(
+                "SELECT CHECK_CLAUSE FROM information_schema.CHECK_CONSTRAINTS "
+                "WHERE CONSTRAINT_SCHEMA = DATABASE() AND CONSTRAINT_NAME = 'ck_pc_artifact_tags_family'"
+            )
+        )
+        if clause is not None and "topic-memory" not in clause:
+            await connection.exec_driver_sql(
+                "ALTER TABLE pc_artifact_tags DROP CHECK ck_pc_artifact_tags_family, "
```
Teingi: **[P1] Fix the legacy tag migration before running it during startup**

Combining `DROP CHECK` and `ADD CONSTRAINT` in this `ALTER TABLE` fails on SeekDB v1.0.0.0 with `1235: Add/modify constraint together with other DDLs not supported`. I reproduced this with the Runtime at this commit: a fresh database starts successfully, restoring the previous family constraint makes startup fail here, and applying the migration as separate statements restores startup. Existing deployments with the old constraint therefore cannot start after upgrading, even if they do not use Topic Memory.

Please make the upgrade idempotent and add a real-backend legacy-schema startup test. My preferred approach is to remove the hard-coded family whitelist CHECK from both new and existing tables, while retaining the primary key, foreign key, and target-shape CHECK; the API/runtime already validate taggable families. Existing databases would still need a one-time migration: a standalone `DROP CHECK` for OceanBase/SeekDB and a transactional table rebuild for SQLite. If the whitelist is retained, use supported separate DDL statements and handle retries after an interruption between them. The blocking issue is the startup failure; removing the whitelist is a design recommendation.

Zxf-xufeng: Fixed in ce590718.
The migration now removes the hardcoded family whitelist CHECK from both new and existing tag tables, while preserving the primary key, foreign key, and target-structure CHECK. OceanBase/SeekDB execute a standalone DROP CHECK; SQLite rebuilds the table transactionally. Repeated startup safely skips the migration once the constraint is gone.
Added real-backend regression tests covering legacy databases, startup without Topic Memory usage, repeated startup, and preservation of existing tags and ETags. The original startup failure was reproduced on SeekDB 1.4.0.0, and the fix passed. All 57 relevant tests and repository checks passed.

Teingi (approved):
LGTM

PR #1616, opened 2026-09-15
perf(server): read POST /v1/stats in one transaction per selection

### LINKED ISSUE #1599: bug(server): POST /v1/stats latency degrades sharply under low concurrency on file-backed SQLite
## Summary

`POST /v1/stats` shows **superlinear latency growth under very low concurrency** on a **file-backed SQLite** deployment (the default), and frequently crosses the **10 s default client timeout**, surfacing as an opaque `TransportError` ("无有效 HTTP 响应").

On a small local database (~692 KB, dozens of scopes, Basic mode / no model), a single sequential request already varies between ~0.22 s and ~5.16 s, and **8 concurrent requests take 7.55 s – 10.26 s** — i.e. total latency scales roughly linearly with concurrency (≈ n × per-request cost). A control endpoint (`GET /`, which 404s) at the same n=8 concurrency stays at 0.115 s – 0.192 s, so this is **not** host/network starvation: it is specific to the DB-backed stats path.

## Environment

- PowerContext 1.0.0 (pip install), server launched via `powercontext server launchd` (uvicorn, Streamable HTTP MCP)
- Storage: file-backed SQLite (default; `journal_mode=WAL`, `busy_timeout=5000`)
- macOS 12.7.6, Intel x64
- Python 3.11
- Inspected against `master` @ `545b1609ee55a58348e734d2c27919c43b0a5051` (2026-09-14)

## Reproduction

Server with the default file-backed SQLite. Then:

```bash
# sequential baseline (n=1)
curl -s -o /dev/null -w "%{time_total}\n" -X POST http://127.0.0.1:8000/v1/stats \
  -H 'content-type: application/json' \
  -d '{"selection":{"root":{"mode":"all"}},"period":"FIELD_30D"}'

# concurrency n=8
for i in $(seq 1 8); do
  curl -s -o /dev/null -w "%{time_total}\n" -X POST http://127.0.0.1:8000/v1/stats \
    -H 'content-type: application/json' \
    -d '{"selection":{"root":{"mode":"all"}},"period":"FIELD_30D"}' &
done; wait
```

The client symptom is the same request failing with:

```
powercontext.errors.TransportError: 无有效 HTTP 响应 for POST /v1/stats
```

because the default client timeout is 10 s (`src/powercontext/client/settings.py:49`). A workaround exists via the global flag `powercontext --timeout 60 stats`, but that only masks the server-side regression.

## Observed data

| Scenario | p50 / range (wall) |
|---|---|
| `POST /v1/stats`, n=1 (sequential) | 0.22 s – 5.16 s (high variance) |
| `POST /v1/stats`, n=8 (concurrent) | 7.55 s – 10.26 s (some > 10 s → TransportError) |
| `GET /` (404 control), n=8 (concurrent) | 0.115 s – 0.192 s |

Server-side profiling showed per-request CPU ≈ 34 ms while wall reached ~10 s, i.e. **>99 % of the time is spent waiting, not computing** — consistent with request serialization at the SQLite read/concurrency layer rather than with aggregation cost.

## Analysis (with hypotheses ruled out)

The handler chain is:

- `src/powercontext/server/app.py:2095` `async def get_stats` → `application.statistics.overview(...)`
- `src/powercontext/builtin/runtime/application.py:679` `StatisticsApplication.overview` →
  ```python
  async with self._runtime._operation():
      resolved = await self._runtime.scopes.resolve_selection(selection)
      captured_at = self._runtime._clock()
      snapshots = tuple([
          await self._runtime._statistics(scope.scope_id).overview(period, captured_at)
          for scope in resolved
      ])
  ```
  i.e. **one transaction per resolved scope, run sequentially** (the `await` inside the comprehension forces per-scope serialization within a single request).
- `src/powercontext/builtin/runtime/statistics.py:95` `RelationalScopedStatistics.overview` → one `self._database.transaction()` per scope, issuing **5 sequential queries** (inventory, cursor load, memory entries, usage, recall usage).

Hypotheses I checked and excluded:

1. **In-memory shared-connection lock** — `src/powercontext/builtin/persistence/database.py:46` `_shared_connection_lock` is created only when `shared_connection=True`, and `src/powercontext/builtin/persistence/sqlite/profile.py:96` passes `shared_connection=config.is_in_memory`. On a **file** DB this lock is `None`, so it cannot be the cause.
2. **Global operation serializer** — `src/powercontext/builtin/runtime/application.py:2493` `_operation()` wraps bookkeeping in an `asyncio.Condition` (`self._lifecycle`) that only tracks depth/`_active_operations` and `notify_all()` when the count returns to zero. It does **not** hold a critical section around the work, so concurrent `_operation()` calls are not serialized by it.

Most likely remaining cause: the **file-backed SQLite connection path** itself. Each `overview` borrows a connection from the default SQLAlchemy `QueuePool` (pool_size 5 for file SQLite) and opens a read transaction; under n=8 concurrency the requests contend on that pool and/or on SQLite read-transaction lock acquisition, with `busy_timeout=5000 ms` (`src/powercontext/builtin/persistence/sqlite/profile.py:50`) letting contended locks wait up to 5 s. The per-scope sequential loop (`application.py:688`) compounds this because the cost grows with both scope count and concurrency.

Supporting code references (all pinned to `545b1609ee55a58348e734d2c27919c43b0a5051`, inspected 2026-09-14):

- `src/powercontext/server/app.py:2095` — `get_stats` handler
- `src/powercontext/builtin/runtime/application.py:679` — `StatisticsApplication.overview` (per-scope sequential loop)
- `src/powercontext/builtin/runtime/statistics.py:95` — `RelationalScopedStatistics.overview` (5 sequential queries / transaction)
- `src/powercontext/builtin/runtime/application.py:2493` — `_operation()` (bookkeeping only, not a serializer)
- `src/powercontext/builtin/persistence/database.py:46` — `_shared_connection_lock` (in-memory only)
- `src/powercontext/builtin/persistence/sqlite/profile.py:96` — `shared_connection=config.is_in_memory`
- `src/powercontext/builtin/persistence/sqlite/profile.py:50-51` — `busy_timeout_ms=5000`, `journal_mode="WAL"`
- `src/powercontext/client/settings.py:49` — client `timeout` default 10.0 s
- `src/powercontext/client/cli.py:284` — `stats` command (no per-command timeout; relies on global `--timeout`)
- `src/powercontext/client/client.py:1184` — raises `TransportError` on `httpx.HTTPError` (incl. timeout)
- `openapi/powercontext.yaml:2832` — `/v1/stats` is POST-only

## Impact

- The WorkBuddy integration (UserPromptSubmit hook + MCP) calls `/v1/stats`; under concurrent invocations it intermittently fails with an opaque `TransportError`, and the 10 s default masks the underlying slowdown.
- On multi-scope selections (`all` / subtree root) latency scales with scope count × concurrency, so larger deployments are affected worse.

## Suggested mitigations (for maintainer consideration, not prescribed)

1. Parallelize the per-scope `overview` in `StatisticsApplication.overview` with `asyncio.gather` (each scope already uses an isolated read transaction).
2. Right-size the file-SQLite connection pool, or serve stats from a consistent read snapshot, so concurrent stats calls don't queue on the default 5-connection pool.
3. Return a clear server-side timeout / `503` (or raise the default client timeout) instead of letting the client hit its 10 s socket timeout and emit an unhelpful message.
4. Add a concurrency benchmark for `POST /v1/stats` to CI to catch regressions.

## Questions

- Is the per-scope sequential loop in `StatisticsApplication.overview` intentional, or is parallelization safe given per-scope transaction isolation?
- Is the default 5-connection pool for file SQLite the contention point, or is it the WAL read-transaction lock under `busy_timeout`? Happy to help profile if a maintainer points me at the intended benchmark harness.

L4XB:

Closes #1599.

# Rationale for this change

I reproduced the reported shape and then measured where the time goes, and the measurement
disagrees with the issue's own root-cause guess, so the numbers come first.

**Reproduced.** File-backed SQLite, uvicorn, scopes holding Memory entries, `selection: all`,
`period: 30d`. Burst wall time is almost exactly `n x per-request cost` — concurrency buys
nothing — while the `GET /` 404 control stays flat, so it is not host starvation:

| scopes | n=1 | n=4 | n=8 | `GET /` at n=8 |
|---|---|---|---|---|
| 10 | 0.066 s | 0.250 s | 0.572 s | 0.009 s |
| 40 | 0.237 s | 0.976 s | 2.145 s | 0.008 s |
| 80 | 0.521 s | 1.958 s | 4.140 s | 0.011 s |

**It is not SQLite contention.** Sampling the server process during a burst, `cpu / wall` is
87 % at n=1, 132 % at n=8 and 137 % at n=16 — the server is busy, not blocked, so nothing is
waiting on a `busy_timeout` or a pool slot. Two supporting details: SQLAlchemy gives a
file-backed `sqlite+aiosqlite` URL an `AsyncAdaptedQueuePool` of size 5 with 10 overflow, so
15 connections are available against the 8 concurrent requests in the report; and each request
holds one connection at a time, so the pool is never the queue.

**What the time actually is.** One `POST /v1/stats` over 41 Scopes issued **810 SQL statements
in 42 transactions**, every one of them sequential:

| count | statement |
|---|---|
| 120 x 4 tables | the Memory Revision load (`pc_artifacts` + lineage + publications), **three times per Scope** |
| 41 each | source journal head, artifact head counts, candidate head counts, source cursor, Memory head revision, daily model usage, daily recall tokens |
| 40 | Memory entry versions |

Each round trip is an event-loop hop into aiosqlite's connection thread and back. At roughly
0.3 ms of overhead apiece, 810 of them are the request. That is why latency is linear in Scope
count, and why it is linear in concurrency too: the work is on one event loop, so requests
queue behind each other.

The three Memory Revision loads per Scope are the largest single block. `_memory_entries` needs
`(kind, state)` pairs; it reads the head, then `MemoryService.entries()` re-reads that Revision
to prove the caller's Memory matches storage, and `RelationalMemoryBackend.entries()` re-reads
it again to get the version ids.

Three steps, one commit each.

1. **`perf(persistence)`** — batched siblings for the four scoped reads
   (`inventory_many`, `usage_many`, `recall_usage_many`, `SourceCursorRepository.load_many`).
   They take the whole selection and group by `scope_id`; the single-Scope methods now delegate
   to them so the two paths cannot drift. Selections are chunked at `SELECTION_BATCH_SIZE` (500)
   so a large one cannot exceed a backend's bind-parameter ceiling, and duplicate Scope ids
   collapse before the query is built.
2. **`perf(memory)`** — `MemoryService.head_entries()` reads the head once and validates its
   entries against it. `entries()` is untouched and still re-reads, because that check is what
   makes it safe for a Memory that arrived from outside; only the head path stops proving a
   fresh backend read against itself.
3. **`perf(server)`** — `StatisticsApplication.overview` now calls `overview_selection`, which
   opens **one** transaction for the whole selection, issues the batched reads, and projects each
   Scope through the same assembly the single-Scope path already used. `RelationalScopedStatistics.overview`
   is split into `_read` and `_assemble` so both paths share one projection.

## Result

Same databases, same harness, `41` Scopes with Memory entries:

| | before | after |
|---|---|---|
| statements per request | 810 | **410** |
| transactions per request | 42 | **2** |
| runtime `overview(all)`, n=1 | 0.248 s | **0.117 s** |
| runtime `overview(all)`, n=8 | 2.22 s | **1.11 s** |

Over HTTP, `POST /v1/stats` burst p50:

| scopes | n | before | after |
|---|---|---|---|
| 10 | 1 / 4 / 8 | 0.066 / 0.250 / 0.572 s | **0.038 / 0.198 / 0.363 s** |
| 40 | 1 / 4 / 8 | 0.237 / 0.976 / 2.145 s | **0.187 / 0.563 / 1.324 s** |
| 80 | 1 / 4 / 8 | 0.521 / 1.958 / 4.140 s | **0.315 / 1.152 / 2.453 s** |

Because a single sweep can drift with machine load, I also ran the 40-Scope n=8 case as three
interleaved before/after pairs on one server-per-run: before `2.270 / 2.295 / 2.381 s`, after
`1.368 / 1.337 / 1.181 s`.

## What this does not fix

The endpoint is still **linear in Scope count** — the remaining 410 statements are two Memory
Revision loads per Scope (`RelationalMemoryBackend.entries()` re-reading the Revision it was
handed a reference to) plus one head-revision lookup and one entry-version read each. Removing
those needs a backend read that accepts an already-loaded `Memory`, which changes the
`MemoryBackend` contract, so I left it out of a fix for this issue. It is the obvious next step
and would take the same request to roughly 90 statements.

I also measured the issue's own suggestion, `asyncio.gather` over the per-Scope overviews: it
helps (n=8 3.11 s -> 1.91 s in-process) but less than batching, it does not reduce the work, and
it wants up to `scopes x concurrency` pool checkouts against 15 available connections. Batching
inside one transaction gives every Scope the same snapshot instead.

No API, schema or response change. Response bodies are byte-for-byte identical: I dumped the
full `overview(all)` result (minus `as_of`) before and after over the same database and diffed
them. `openapi/powercontext.yaml` is untouched.

`MemoryService.head_entries()` is new; nothing is removed or deprecated.

# How was this change tested?

Commands, all on macOS arm64 / Python 3.11:

- `uv run python -m pytest --doctest-modules --ignore=tests/e2e -q` — 2290 passed, 45 skipped,
  8 failed. The same 8 fail unmodified on `master` (`13518e4d`) in a clean worktree: three
  `tests/client/test_receiver_service.py` systemd cases and five `tests/test_scan_e2e_evidence.py`
  cases. The failure sets are identical before and after.
- `uv run python -m pytest tests/e2e -q` — 226 passed, 56 skipped.
- `uv run ruff check src/ tests/`, `uv run ruff format --check src/ tests/`, `uv run ty check` — clean.

Tests added:

- `test_statistics_selection_reads_each_table_once_regardless_of_scope_count` — the regression
  guard. It counts statements per table for a selection of 2 and of 6 Scopes and requires one
  query per table in both. On `master` it fails with `3` for every table (2 seeded Scopes plus
  the default one). This freezes a query-count budget deliberately: "a selection must not cost a
  query per Scope" is the external property the issue is about.
- `test_statistics_selection_reports_the_same_scope_rows_as_scoped_overviews` — the correctness
  guard: each `by_scope` row from a batched selection equals the row from that Scope's own
  `overview()`, over Scopes seeded with deliberately different content. Passes on `master` too.
- `test_memory_head_entries_matches_head_and_entries_read_separately` — `head_entries()` returns
  what `head()` + `entries()` return, including an entry that has been deactivated so `state`
  is not uniformly `active`.


Teingi:
good job!

Teingi (approved):
LGTM

PR #1617, opened 2026-09-15
feat(pi): add structured work continuity tools

wutongyuonce:
## Summary

Adds the missing structured work-continuity operations to the native Pi integration:

- `pc_work_contract` -> `create_work_contract`
- `pc_handoff_current` -> `handoff_current_work`
- `pc_handoff_acknowledge` -> `acknowledge_handoff`
- `pc_task_outcome` -> `record_task_outcome`

Refs #1338

## Final design

The implementation follows the original plan while keeping the scope at the Recommended Pi profile:

1. Create a Work Contract only for explicitly delegated work that has a stable inspected baseline. The contract records objective, grounded facts, scope, exclusions, completion criteria, authorization notes, and open questions; it grants no execution authority.
2. Use `pc_handoff_current` at a current-work transfer boundary. It prepares a temporary Handoff from the current Scope; the returned Handoff carrier is passed unchanged to the receiving task or to an explicitly requested durable commit.
3. The receiving task verifies the Handoff against live repository state, capabilities, and authorization, then acknowledges the prepared or exact target with the three receiver checks.
4. Record a Task Outcome only at an actual completion or interruption boundary, preserving status, observations, checks, produced artifacts, and remaining work.

All four tools use the existing OpenAPI contracts, including nested citation/reference and generation metadata schemas. They are marked mutating and included in the write-operation set, so interactive confirmation, no-UI refusal, Scope injection, and secret filtering apply consistently. The project-context Skill documents when to use each operation and requires exact pass-through of Handoff and citation references.

This PR does not claim the Full profile and does not add External Skills, Experience/Skill generation, candidate decisioning, Stats/Memory Changes, or Workstream persistence.

## Verification

- Pi package tests: 91 passed
- TypeScript typecheck: passed
- Integration manifest tests: 26 passed
- Generated manifest check: passed
- `ty check`: passed
- `git diff --check`: passed

Implementation and verification were performed with Codex (GPT-5.6 Sol). An independent read-only review was performed in a separate Codex session using GPT-5.6 Luna with medium reasoning; all actionable findings were resolved before submission.


Teingi (commented):
Two P2 findings reproduced on this commit using Pi 0.82.1 argument validation and a local SQLite-backed HTTP server.

integrations/pi/plugins/powercontext/src/tools.ts
```diff
@@ -55,6 +55,117 @@ const SEARCH_MODES = Type.Union([
...
+  content: HANDOFF_CONTENT,
+  generation: Type.Optional(Type.Union([HANDOFF_GENERATION_ENVELOPE, Type.Null()])),
+})
+const TASK_CHECK = Type.Object({
+  name: NON_EMPTY_STRING,
+  status: Type.Union([
+    Type.Literal('passed'), Type.Literal('failed'), Type.Literal('skipped'), Type.Literal('timed_out'),
+    Type.Literal('unavailable'), Type.Literal('cancelled'), Type.Literal('unknown'),
+  ]),
+  details: Type.Union([NON_EMPTY_STRING, Type.Null()]),
```
Teingi: [P2] Keep TaskCheck.details optional

The OpenAPI TaskCheck schema allows details to be omitted, but this declaration makes it required in Pi. A check such as {"name":"tests","status":"passed","basis":"declared","evidence":[]} causes Pi 0.82.1 to reject the entire pc_task_outcome call with "outcome.checks.0.details: must have required properties details" before sending a request. The same outcome succeeds with HTTP 202 when sent directly to the server; adding details: null also makes the Pi call succeed. Please wrap this union in Type.Optional so valid check results can pass through unchanged.

integrations/pi/plugins/powercontext/skills/project-context/SKILL.md
```diff
@@ -28,14 +28,31 @@ evidence. Do not call `pc_remember` merely to duplicate the current prompt.
...
+  grants no execution authority.
+- When handing off the current work, inspect the objective, state, disposition, next action, omissions, and exact evidence,
+  then call `pc_handoff_current` once with a unique `source_id`. Pass its returned `handoff` member unchanged to the
+  receiving task or to `pc_handoff_commit` when the user explicitly requests a durable milestone.
+- The receiving task must verify the Handoff evidence against the current repository, instructions, capabilities, and
+  authorization before calling `pc_handoff_acknowledge`. Use the same prepared or exact target, all three receiver check
+  states, and `accepted`, `needs_clarification`, or `declined`. Never accept without all checks confirmed.
+- At an actual completion or interruption boundary, call `pc_task_outcome` with the exact status, observations, checks,
+  produced Artifacts, and remaining work. Do not treat every session stop as completion. Preserve failed, skipped,
+  timed-out, unavailable, cancelled, and unknown checks exactly; pass any exact Handoff receipt reference unchanged.
```
Teingi: [P2] Limit outcome receipt links to accepted committed handoffs

The default workflow above transfers a prepared Handoff and allows it to be acknowledged, but this instruction does not distinguish that receipt from one for a committed Handoff. Following pc_handoff_current -> pc_handoff_acknowledge(selection: "prepared", status: "accepted") and copying receipt.source into outcome.handoff_receipt_ref produces HTTP 422: _validate_outcome_receipt requires an accepted exact acknowledgement with a selected revision. Omitting the reference, or using an accepted committed-Handoff receipt, succeeds with HTTP 202. Please explicitly instruct callers to omit this field for prepared acknowledgements and only link receipts for accepted committed Handoffs.

integrations/pi/plugins/powercontext/src/tools.ts
wutongyuonce: @Teingi Fixed in 7a7240d3. `details` is now `Type.Optional(Type.Union([NON_EMPTY_STRING, Type.Null()]))`, so an omitted `details` no longer fails Pi's argument validation before the request is sent.

I added the regression test you effectively described to `tests/tools.spec.ts`: it validates the registered `pc_task_outcome` parameter schema with TypeBox `Value.Check`, and asserts a check without `details` passes, `details: null` passes, and a check missing `basis` still fails. Confirmed the test fails against the previous declaration.

I also diffed every other structured-work TypeBox declaration against the OpenAPI `required`/`nullable` lists (`WorkClaim`, `WorkContract`, `CurrentWorkHandoff`, `ReceiverChecks`, `HandoffStatement`, `HandoffOmission`, `HandoffGenerationMetadata`, `HandoffGenerationEnvelope`, `HandoffContent`, `PreparedHandoff`, `TaskOutcome`). `TaskCheck.details` was the only mismatch.

integrations/pi/plugins/powercontext/skills/project-context/SKILL.md
wutongyuonce: @Teingi Fixed in 7a7240d3. The Skill now states the field may only be linked when the work closes an accepted committed Handoff: pass that acknowledgement's exact `receipt.source` as `handoff_receipt_ref`, and omit it for a `prepared` acknowledgement or when no Handoff is covered.

This also restores the wording the `claude-code` and `codex` project-context Skills already use ("When the work continues an accepted committed Handoff, include that exact Receipt SourceRef"), so the Pi Skill no longer diverges from them.

`docs/en/docs/integrations/pi.md` and `docs/zh/docs/integrations/pi.md` carried the same "pass returned ... references ... unchanged" ambiguity without the committed-Handoff qualifier, so I tightened those two lines as well; say the word if you would rather keep this commit smaller.

wutongyuonce:
Follow-up commit `0e93f1fe` after a full self-review of this PR (schema parity, payload routing, write gating, capability manifest, Skill against the Server validators).

Skill fixes found in that pass:

- **Restored the receiver-side resolution step.** The restructure introduced by this PR dropped the instruction that the receiving task resolves the transferred value with `pc_handoff_continue` (`selection: "prepared"` with that exact value, or `selection: "exact"` with the committed revision) before verifying and acknowledging it. `tests/package.spec.ts` was changed to assert `pc_handoff_current` in place of `pc_handoff_continue`, so the removal was invisible to the test surface.
- **Stated the Server's claim rules.** `WorkClaim` and `TaskCheck` reject both `verified` without exact evidence and `declared` with evidence (HTTP 422). The Skill never said so, while the `claude-code` and `codex` Skills do ("Use `declared` for claims without an exact same-scope citation; never invent `verified` evidence"). The Skill now states the rule once for Work Contracts, Handoffs, and Task Outcomes.

Everything else I re-checked is consistent and needed no change:

- All twelve new TypeBox structures match the OpenAPI `required`/`nullable`/enum/`minItems`/`maxItems`/pattern constraints (`WorkClaim.evidence` 31, `TaskCheck.evidence` 32, `HandoffStatement.citations` 1..32, generation metadata digests, `MemoryCitation` reference patterns); `TaskCheck.details` was the only mismatch and is fixed in `7a7240d3`.
- The four payloads match `CreateWorkContractRequest`, `HandoffCurrentWorkRequest`, `AcknowledgeHandoffRequest`, and `RecordTaskOutcomeRequest`; `handoff_current_work` returns a `PreparedWorkHandoff` whose `handoff` member is a `PreparedHandoff`, so the documented pass-through to `acknowledge_handoff` and `commit_handoff` is schema-compatible.
- Write gating: all four are `mutates: true` and join `WRITE_OPERATIONS` in `invoke.ts`, which is the plugin's only global operation list.
- `capabilities.toml`: the added capabilities make the derived profiles `minimal` + `recommended`, matching the declared `profiles` list; `generate_integration_manifest_docs.py --check` and `tests/test_integration_manifest.py` (26 passed) are green.

Local verification: `pnpm test` 92 passed, `pnpm typecheck` clean, `git diff --check` clean.

Teingi (approved):
LGTM

PR #1618, opened 2026-09-15
docs(rfc): add declarative artifact families and generation profiles

MaoMengww:

Refs #1424

# Rationale for this change

Family identity is already a free string at the domain-value level (`ArtifactRef.family`), but every execution path keeps its own closed set
— artifact classes, authorization profiles, ID prefixes, processing bindings, contract enums — so changing the set of families requires
platform source and contract changes and "extend without forking" does not hold today. This RFC defines two governed extension points to
close that gap without executing extension code: what class of derived artifact is produced (declarative Artifact families), and how that
content is generated (Scope-owned generation profiles).

- Declarative Artifact families: a data-only extension package (`powercontext.extension.json` plus JSON Schema documents), closed manifest
validation, fail-fast activation with per-family failure isolation, and no extension code on the production path.
- Multi-version content schemas: several `schema_version`s per family, `current_schema_version` for new writes, the `schema` content marker
as the version authority, and `schema_version` columns on `pc_artifacts` and `pc_artifact_candidate_versions` with `RESTRICT` composite
foreign keys into the new `pc_extension_families` description table.
- Content validation through thin wrapper types (`RootModel[dict[str, Any]]`, `ArtifactDraft`/`Artifact` subclasses) reusing the existing
strict decode seam, with the manifest schema as the only authority and document-scoped `$ref` resolution.
- Fixed platform rules for retrieval projection and context rendering, plus projection rebuild on activation and upgrade.
- Generation profiles: the built-in `generation-profile` family, immutable revisions, profile content model (target family, prompt
reference, model catalog entry, bounded settings, output cap, empty-result field, failure policy), resolution keyed by profile key and
frozen per operation.
- Model catalog as server configuration, with the existing `InferenceConfig.generation_*` normalized into a `default` entry, and profile
limits min'd against catalog bounds.
- Generation provenance: a structured record plus a derived digest whose input contract is frozen by `digest_input_version`, covered by a
new `pc_artifact_generation_provenance` table and candidate-side columns carried into Revision lineage at approval.
- A generic `POST /v1/generation/generate` endpoint (profile-driven generation, or a manual proposal when only `family` is given), plus a
read-only `GET /v1/extensions` discovery endpoint that makes default-deny explicable through four per-family states.
- Registry-driven family dispatch replacing the closed family sets, contract changes (read-path family enums opened, write-path unions kept
closed), and the deactivate/downgrade/reinstall state machine with historical readability from the family description table.
- Author tools (`extension init`/`validate`/`lock`/`diff`) and administrator tools (`extension enable`/`print-config`,
`generation-profile`), plus a config wizard step; `validate` performs structural lint and a skeleton dry run so an unconstrained schema
cannot pass.
- Documents drawbacks, alternatives, prior art, unresolved questions, and acceptance criteria covering external behavior.

No released behavior changes. This PR adds design documents only.

Design decisions worth calling out for reviewers:

- The platform performs no content migration: historical Revisions are validated and rendered under the version each recorded.
- Opening the family enums on read paths is an intentional breaking change for typed integrations when implemented, and must ship with `make
api-generate` and `make contract-test`.
- Not in the first slice: code-hook extension points, automatic triggering, tags and cross-Scope publication for extension families, and
Dashboard presentation.

# How was this change tested?

make docs-test: no issues.
uv run prek run --files ...: all hooks passed.

# AI usage statement
Claude Code assisted me with research, design, and review.


docs/en/rfcs/1618_declarative_families_and_generation_profiles.md
```diff
@@ -0,0 +1,477 @@
...
+
+`_candidate_draft` carries it into the new Artifact draft at approval time, making "generated by whom" visible in the Revision's lineage.
+
+## Registry-driven family dispatch
+
+Family dispatch becomes registry-driven.
+
+Approval still completes inside a single database transaction: validate the proposal under the candidate's recorded `schema_version`, enforce declarative lineage, write the Artifact, update derived indexes, and `mark_approved` — five steps in one transaction, rolling back entirely if any step fails. With `cardinality = "singleton"`, a candidate that does not explicitly give a target implicitly points at the current head and performs a CAS; `collection` keeps explicit semantics.
```
Teingi: [P2] Freeze the singleton target when the candidate is created

Please specify when the implicit target is resolved. If a candidate is based on revision 1, revision 2 is committed while it awaits review, and approval resolves the latest head, the old proposal can overwrite revision 2 while still passing CAS. I reproduced this using the current ArtifactRepository on SQLite: resolving latest at approval commits revision 3, while retaining revision 1 correctly raises a conflict. Resolve and persist the exact target when preparing the candidate, including an explicit expected-absence state for first creation, and check that same expectation at approval.

docs/en/rfcs/1618_declarative_families_and_generation_profiles.md
```diff
@@ -0,0 +1,477 @@
...
+- A caller may only name a profile key and **cannot submit profile content**: the body is always written by an administrator and subject to authorization and budget validation.
+
+## Model catalog
+
+The catalog is server configuration declared by the deployment, shaped as `name → {provider, model, base_url, credential env name, bounds}`. The existing `InferenceConfig.generation_*` is normalized at configuration load time into a catalog entry named `default`, so existing deployments keep working unchanged. A profile's limits are min'd item by item against the catalog entry's bounds, and any clamped item records a content-free diagnostic.
+
+## Generation provenance
+
+`ArtifactLineage` gains a pair of nullable non-evidence fields, `generation_source: ArtifactRef | null` and `generation_provenance: GenerationProvenance | null`, following the existing all-or-nothing validation. `GenerationProvenance` is **structured fields plus a derived digest**: `prompt_ref`, `effective_limits`, `target_schema_version`, `digest`, `digest_input_version`.
```
Teingi: [P2] Retain the resolved generation configuration behind the digest

A catalog entry can change models, or an extension upgrade can replace a template under the same ID, without changing the profile revision. The digest detects a difference, but the listed fields do not preserve enough information to recover the old model configuration or template after replacement/uninstall. Please retain a resolved configuration snapshot or references to durably stored immutable versions, including model identity, effective settings, template version, and extension package version. Acceptance criterion 2 also needs alignment: catalog/package changes do not currently have a defined mechanism that creates a new profile revision.

docs/en/rfcs/1618_declarative_families_and_generation_profiles.md
```diff
@@ -0,0 +1,477 @@
...
+
+`dump_model` uses `model_dump_json(by_alias=True)`, which is **order-preserving but not normalizing**: the payload's key order determines the stored bytes. Built-in families gain stability from model field declaration order, while a declarative family's key order comes from the author or the model output, so semantically identical content can produce different bytes. S1's approach is recursive key sorting (array order preserved) inside the wrapper model's `mode="before"` validator, which the generation and approval entry points both inherit by construction.
+
+## Default rules for retrieval projection and context contribution
+
+A declarative family cannot carry projection or rendering functions, so the platform decides both with **one fixed rule**, introducing no configurable dialect:
+
+- **Projection**: take every string field in the content, exclude the version marker `schema`, and concatenate them in a stable order into searchable text. This rule happens to reproduce the sample family's expected projection and adds no configuration.
+- **Rendering**: render scalar fields one per line as `Label: value`, with the section title taken from `prepared_context.display_name`.
```
Teingi: [P2] Define rendering for arrays and nested objects

The Runbook example stores its useful content in steps[], symptoms[], and failure_handling[], but this rule renders only scalar fields. Following it literally produces valid, approved content whose actual instructions disappear from PreparedContext. Please define recursive rendering for arrays and nested objects, preserving step order and field labels under the Runtime's byte budget. An acceptance case should verify that the Runbook steps reach the assembled context.

docs/en/rfcs/1618_declarative_families_and_generation_profiles.md
```diff
@@ -0,0 +1,477 @@
...
+
+- **Activation-time fail-fast.** An invalid or self-contradictory manifest (unknown fields, illegal schema, duplicate family, incompatible naming, undeclared `target_family`) is rejected at startup; that extension stays inactive and appears with a content-free reason code in readiness and on `GET /v1/extensions`. It neither blocks Runtime startup nor affects other families.
+- **Runtime fail-soft.** When a generation fails, output does not satisfy the schema, a timeout or request bound is exceeded, or evidence is insufficient, it fails before persisting a candidate and affects only that generation; unrelated capabilities keep working.
+
+## Relationship to existing users
+
+- Scope-level custom instructions for Prompts stay as they are and remain the only entry point for built-in family guidance. Generation profiles do not replace them and can only reference them.
+- The `family=profile` user-profile family keeps its per-scope `activation_mode` and does not migrate to the generic Review policy.
+- No built-in family is migrated: they keep using the existing typed generation endpoints, and `generation_profile` is always empty for built-in candidates.
```
Teingi: [P2] Reconcile the Experience example with the S1 boundary

Example 1 says that writing an experience.generate profile changes subsequent Experience generation and records its exact revision, while this paragraph says built-in candidates always have an empty generation_profile. These describe different observable behavior. If S1 covers extensions only, use an extension family in the example and defer the Experience claim. If Experience is included, define how its typed endpoint selects a profile, when it takes effect, and how provenance is recorded.

Teingi (commented):
The Family/Profile split and reuse of Candidate → Review → Revision make sense. I suggest resolving the four contract gaps below before implementation.

For S1, I would focus on one Runbook plus one generation profile, covering exact evidence → Candidate → human Review → retrieval/context contribution → exact reads after deactivation. The lock/diff tools, config wizard, and built-in family integration could follow separately.

I checked the wrapper/codec round trip and the target-version example against the repository on SQLite. These comments concern the proposed design; extension support itself is not implemented in this PR.

MaoMengww:
@Teingi  Done. singleton target frozen at candidate creation (including the expected-absence state), provenance as a complete resolution-time snapshot with a frozen digest input contract, acceptance 2 split by domain as 2a–2d, one fixed recursive rule for array and nested-object rendering (acceptance 25–27), and Example 1 switched to an extension family — and built-in family integration moved to Future possibilities because its payoff is low (parameter tuning is already covered by deployment configuration and Prompt customization). Slices: S1 = one Runbook extension family + one generation profile; S2 = lock/diff, enable / print-config, the config wizard, and discovery-endpoint refinements

docs/en/rfcs/1618_declarative_families_and_generation_profiles.md
```diff
@@ -0,0 +1,489 @@
...
+
+## Persistence change list
+
+| Table | Nature | Change |
+| --- | --- | --- |
+| `pc_artifact_generation_provenance` | **new** | Records Artifact Revision generation provenance by `(scope_id, family, artifact_id, revision)`: the profile reference and the structured generation provenance. Shaped like the existing `pc_artifact_publications`, with a composite foreign key into `pc_artifacts` on the reference columns. |
+| `pc_extension_families` | **new** | Family description: primary key `(family, schema_version)`, storing that version's JSON Schema text, the source extension identifier, and the activation time, and recording the `prompt_templates[].id` and template text declared by that version's manifest; append-only, referenced by the artifact and candidate tables through `RESTRICT` composite foreign keys. It is what allows historical Revisions to be validated and rendered after deactivation, and it is what resolves generation provenance's template anchor, as well as `diff`'s authoritative input. |
```
Teingi: [P2] Version template text independently of the content schema

Acceptance 2c permits changing a template under the same ID without changing the content schema, but this append-only table has only `(family, schema_version)` as its primary key. A template-only package upgrade therefore has no distinct row for the new text: inserting conflicts, overwriting loses the old template, and ignoring the conflict leaves the new template unavailable after uninstall. A stored digest cannot recover the missing text. Please retain immutable template snapshots keyed by a template digest or package version and bind provenance to that exact snapshot. Cover two package versions with the same schema and template ID but different text, with both texts recoverable after uninstall.

docs/en/rfcs/1618_declarative_families_and_generation_profiles.md
```diff
@@ -0,0 +1,489 @@
...
+
+## Default rules for retrieval projection and context contribution
+
+A declarative family cannot carry projection or rendering functions, so the platform decides both with **one fixed rule**, introducing no configurable dialect:
+
+- **Projection**: traverse the content tree depth-first, take every string value, exclude the version marker `schema`, and concatenate them in a stable order consistent with payload normalization into searchable text; a newline boundary is inserted between array entries so a phrase query never matches across two array elements. This rule happens to reproduce the sample family's expected projection and adds no configuration.
```
Teingi: [P2] Define an actual element boundary for phrase matching

A newline does not provide this guarantee. Projecting `steps = ["restart", "database"]` as `restart\ndatabase` still matches the phrase `"restart database"` in SQLite FTS5 with the current `unicode61` tokenizer; I reproduced this with both raw and analyzed projection text. The existing Analyzer also turns that newline into a space, and the current query compiler treats the words as OR terms. Thus the proposed projection cannot satisfy acceptance 26 while reusing the existing analysis/query path. Please specify how indexing and queries preserve array-element boundaries, or remove the cross-element phrase guarantee from the rule and acceptance criteria.

Teingi (commented):
Rechecked at `751c82f1`. The singleton target/expected-absence rule, recursive rendering, and built-in-family scope are now addressed. Two P2 contract gaps remain below. I replayed the CAS checks against ArtifactRepository on SQLite and verified the projection counterexample with SQLite FTS5. These are RFC design findings; extension support is not implemented by this PR.

MaoMengww:
@Teingi Done:
 - Dropped the cross-element phrase guarantee 
- Template text now has its own version axis: pc_extension_prompt_templates keyed by template_digest, with provenance bound to it by a RESTRICT foreign key;

PR #1621, opened 2026-09-15
feat(pi): add experience and skill generation tools

wutongyuonce:
## Summary

Adds the missing Pi entry points for generating reviewable Experience and Skill candidates.

- `pc_experience_generate` -> `generate_experience`
- `pc_skill_generate` -> `generate_skill`

Refs #1338

## Plan and final design

The original plan was to complete the next Pi capability gap after artifact inspection and structured work continuity, while keeping candidate governance separate.

The final implementation follows that plan:

1. `pc_experience_generate` accepts exact Source and Artifact references plus optional target and reason.
2. `pc_skill_generate` accepts the same evidence model and requires the direct provenance `origin`: `experience`, `source`, or `usage`.
3. Both tools route through the existing Scope-aware operation path. The request schemas match the OpenAPI contracts, including the combined maximum of 32 evidence references.
4. Both operations are durable writes: they require interactive confirmation, refuse writes without a UI, inject the resolved Scope, and reject secret-bearing payloads.
5. Generation returns a candidate for human review. It does not approve, publish, install, activate, or execute the resulting Experience or Skill. Candidate decisions remain outside the Pi tool surface.
6. The project-context Skill, capability manifest, bilingual Pi documentation, and focused tests describe and verify these boundaries.

This PR intentionally does not add candidate approval/rejection/revision or External Skills. Pi therefore gains the Experience/Skill generation part of the Full-profile contract, but this PR alone does not claim Full support.

## Verification

- Pi package tests: 95 passed
- TypeScript typecheck: passed
- Integration manifest tests: 28 passed
- Generated manifest check: passed
- `ty check`: passed
- `git diff --check`: passed

Implementation and local verification were performed with Codex (GPT-5.6 Sol). An independent read-only review was performed in a separate Codex session using GPT-5.6 Luna with medium reasoning; all actionable findings were resolved before submission.


integrations/pi/plugins/powercontext/src/tools.ts
```diff
@@ -180,6 +180,31 @@ const TASK_OUTCOME = Type.Object({
...
+const GENERATION_FIELDS = {
+  target: Type.Optional(Type.Union([ARTIFACT_REFERENCE, Type.Null()])),
+  reason: Type.Optional(Type.Union([Type.String({ minLength: 1, maxLength: 2000 }), Type.Null()])),
+}
+const EVIDENCE_REF_COMBINATIONS = Array.from({ length: 33 }, (_, sourceMax) => Type.Object({
+  source_refs: Type.Array(SOURCE_REFERENCE, { maxItems: sourceMax }),
+  artifact_refs: Type.Array(ARTIFACT_REFERENCE, { maxItems: 32 - sourceMax }),
+  ...GENERATION_FIELDS,
+}))
+const EXPERIENCE_GENERATION = Type.Union(EVIDENCE_REF_COMBINATIONS)
```
Teingi: [P1] Keep generation parameters in a top-level object

These unions have no top-level `properties` or `required`. In the locked Pi 0.82.1, the default Anthropic adapter constructs its tool schema from those two fields, so both generation tools reach the actual request payload as `{"type":"object","properties":{},"required":[]}`. I reproduced this through the provider's `onPayload` callback: the evidence fields, `origin`, `target`, and `reason` disappear, while Pi's argument validator rejects `{}`. Claude therefore receives an argument contract that cannot satisfy the tools.

Please expose each tool as a top-level `Type.Object`, validate the combined evidence count before dispatch, and add a test of the provider-converted schema. The current `Value.Check` test does not cover this conversion.

integrations/pi/plugins/powercontext/src/tools.ts
```diff
@@ -547,6 +572,51 @@ export function registerTools(pi: ExtensionAPI, runtime: PluginRuntime): void {
...
+  registerOperationTool(pi, runtime, {
+    name: 'pc_experience_generate',
+    label: 'PowerContext Experience Generate',
+    description:
+      'Generate an Experience candidate from exact Source and Artifact evidence when the user requests ' +
+      'candidate generation. The result remains pending human review; generation does not approve, ' +
+      'publish, install, or activate it. Preserve exact returned references and report the returned ' +
+      'status. Do not use this as a routine Memory write.',
+    parameters: EXPERIENCE_GENERATION,
+    operationId: 'generate_experience',
```
Teingi, line 647: [P2] Give synchronous generation its own request timeout

Both new generation operations use the shared client and inherit its default 1,000 ms timeout. With this commit's real HTTP server and disposable SQLite, I supplied deterministic Experience/Skill generators that take 1.5 seconds. After confirmation, the Pi tools returned `unavailable` after 1,007/1,013 ms, but the server subsequently persisted both candidates as `pending`. The caller loses the successful generation result and may retry an operation that already wrote a candidate.

Please give explicit generation a separate configurable deadline suitable for model calls, without increasing the automatic recall budget, and handle timeout as an unknown write outcome with a way to retrieve the result.

Teingi (commented):
Two issues reproduced on be6b60c5: generation parameters are lost in Pi's Anthropic request conversion, and generation can persist a candidate after the tool reports a timeout.

integrations/pi/plugins/powercontext/src/tools.ts
wutongyuonce: @Teingi Fixed in dedfa0d9. Generation has a separate POWERCONTEXT_PI_GENERATION_TIMEOUT_MS deadline (30s default), while recall keeps its existing budget. A timeout now reports unknown_write_outcome and directs callers to pc_review_list before retrying; added coverage for that path.

integrations/pi/plugins/powercontext/src/tools.ts
wutongyuonce: @Teingi Fixed in dedfa0d9. Both generation tools now use top-level Type.Object schemas, and the combined 32-reference limit is checked before confirmation or dispatch. Added a schema regression test covering the top-level fields Pi uses for Anthropic conversion.

Teingi (approved):
LGTM

Teingi:
resolve conflicts

PR #1619, opened 2026-09-15
docs(rfc): propose Git repository understanding for PreparedContext

Teingi:
# Which issue or RFC does this PR close?

Introduces RFC 1619, building on #1489 (PreparedContext text assembly):

- [English RFC](https://github.com/Teingi/powercontext/blob/codex/git-repository-understanding-rfc/docs/en/rfcs/1619-git-repository-understanding.md)
- [Chinese RFC](https://github.com/Teingi/powercontext/blob/codex/git-repository-understanding-rfc/docs/zh/rfcs/1619-git-repository-understanding.md)

# Rationale for this change

Coding tasks need both current code facts and historical constraints. This proposal lets `prepare_context`
retrieve relevant code and deliver it with existing context under one budget, reducing repeated navigation
and manual context assembly.

# What changes are included in this PR?

- English and Chinese RFCs organized using the standard nine-section template.
- An opt-in `include_code` preparation option, a replaceable CodeGraph adapter, and one local Python repository
  configured per existing Scope, with shared byte and entry budgets.
- Index/content consistency, authorization, fallback behavior, independent read-only queries, optional Source
  capture, and task-based acceptance criteria.
- Existing Artifact families remain intact; the proposed integration uses rebuildable local caches without
  adding Code Artifacts or business database tables.

# Are there any user-facing changes?

Documentation only. The RFC proposes behavior and API changes; this PR does not implement them or require a
data migration.

# How was this change tested?

- `make docs-test`: passed on the final numbered documents, including website tests, link validation,
  compilation, and static export verification of 808 public pages.
- Verified matching bilingual heading hierarchy, tables, numeric policy values, configuration, commands, API
  examples, and diagram flow. Checked local links and the required content exclusions.
- `git diff --cached --check`: passed.
- `.venv/bin/prek run -a`: formatting and lint hooks passed. `ty check` reports three unresolved
  `scripts.integration_guidance_*` / `scripts.evaluate_integration_guidance` imports in
  `tests/test_integration_guidance_evaluation.py`. The same three errors were reproduced in an unmodified
  upstream checkout with the same Python environment. Only this known baseline check is skipped by the commit
  hook.

# AI usage statement

OpenAI Codex (GPT-6) assisted with drafting, translation, editing, and validation of these RFC documents.

PR #1624, opened 2026-09-16
feat(code): add Git repository context to prepare_context

Teingi:

Implements the first-phase design in RFC #1619.

# Rationale for this change

When an agent resumes a coding task, historical Memory and Handoff material can explain decisions without showing the current implementation. This change lets `prepare_context(include_code=True)` add current Python definitions and bounded structural relationships alongside historical context, under one entry and byte budget.

- Add an optional CodeGraph 1.6.0 adapter, explicit local `powercontext code index/status` commands, and a configured mapping from existing Scopes to Git directories. The index is a rebuildable local cache; no new artifact family, workspace entity, or database table is introduced.
- Capture the working tree with content fingerprints, validate freshness before delivery, publish complete cache generations atomically, and enforce path, authorization, time, size, and subprocess boundaries. Raw Git capture avoids running repository clean filters.
- Add `include_code` to context preparation, share the existing assembly budget, and retain historical context when code is unavailable or changes during preparation. Normal preparation does not persist code as Sources or Artifacts.
- Add the HTTP/SDK code-query API and the read-only `powercontext_code_query` MCP tool for status, tree, symbols, callers, callees, impact, affected tests, and bounded source reads. Continuation queries validate the expected fingerprint.
- Support opt-in Codex injection with older-Server fallback. Explicitly saved code-query Sources remain readable as evidence but are excluded from automatic Memory/Experience/TopicMemory/Profile processing.
- Add bilingual usage documentation, real-engine regression tests, and opt-in configured-service and native Codex acceptance tests.

The feature is disabled by default. Deployments opt in by configuring a local repository and the separately installed CodeGraph executable, building its index explicitly, and enabling code context for requests or the Codex integration. Existing requests retain their default behavior; there is no database migration.

The initial implementation supports local Python repositories. Refresh is explicit, cache operations use a cache-wide lock, and static relationships are incomplete. Ambiguous same-name relationships are rejected or omitted. The usage guide documents these limits and the supported engine build.

The 12-task paired evaluation ran both settings twice (48 real model runs). Correct definition-file sets were 22/24 without code and 23/24 with code; caller sets were 20/24 and 22/24. All 16 patch checks and 48 regression checks passed. Median end-to-end time increased from 16.32 s to 22.36 s and median input tokens for completed outputs increased from 6,097.5 to 14,240. These small fixture results support keeping the feature opt-in; they do not establish an efficiency gain or justify reducing tests.

# How was this change tested?

- `make check`: passed, including type checks. Pre-commit hooks also passed over all 80 changed files, including new files.
- `make contract-test`: 48 passed.
- `make docs-test`: passed; verified 811 public pages and internal links.
- Real CodeGraph/Git/HTTP regression run: 18 passed. Final Git capture/process checks: 10 passed, including clean filters, branch changes, SHA256 repositories, and child-process cleanup. The final context-statistics regression passed separately.
- The full non-E2E run initially reported 2,362 passed, 53 skipped, and 7 failures. Tool metadata/manifest issues were fixed; host checks were rerun with supported Node 24 and CLI checks with inherited test configuration cleared. The affected-module rerun had 155 passed, 1 skipped, and 2 remaining manifest ordering failures; the subsequent manifest suite passed all 28 tests after the ordering fix. A second full-suite run was not performed.
- `POWERCONTEXT_TEST_NATIVE_CODEX=1 uv run pytest tests/e2e/real_experience_skill/test_code_context.py --run-real-e2e --real-e2e-env-file .env -q -s`, with `POWERCONTEXT_TEST_CODEGRAPH_EXECUTABLE` pointing to the installed engine: passed in 305.81 s. This used the configured OceanBase, LLM and embedding services, actual HTTP/SDK/MCP paths, and native Codex. It verified repair, executable checks, Handoff continuation after a code change, refreshed fingerprints, explicit Source retention after cache removal, automatic-processing exclusion, and subsequent ordinary Source processing. The temporary database was removed and user Codex state was unchanged.
- Real PowerContext repository probe: 791 Python files indexed in 28.476 s; three prepare queries completed in 4.467/2.695/3.684 s. This is an observed single-host result, not a latency guarantee.
- One-off paired evaluation: 48 configured-model runs, with results, failures, and usage-coverage limits summarized above. Experiment scripts and run reports are kept outside the source tree.
- Wheel build passed; the package includes the engine bridge and CLI entry point. Credentials, downloaded engine binaries, caches, and local test reports are excluded from the commit.

Real engine and configured-service tests require explicit local setup and are skipped when their prerequisites are absent. Local validation does not imply that upstream CI has completed.

# AI usage statement

Implemented and tested with OpenAI Codex (GPT-6), including code generation, test development, documentation, and command execution. The paired task evaluation used the configured `openai:qwen3.7-plus` model. Results and limitations are summarized in this PR.


Teingi (commented):
I reproduced the four issues below on 250a2473 using real CodeGraph 1.6.0 and Git fixtures, plus SQLite/HTTP checks for context preparation. OceanBase and native Codex acceptance were not rerun in this review.

src/powercontext/builtin/runtime/application.py
```diff
@@ -724,22 +755,25 @@ async def prepare(
...
                 return PreparedContextBuilder().empty()
             if authorize_scopes is not None:
-                await authorize_scopes((self.scope_id, *scope.context_references))
+                await authorize_scopes((self.scope_id,) if code_only else (self.scope_id, *scope.context_references))
             return await self._prepare(request, scope)
 
     async def _prepare(self, request: PrepareContextRequest, scope: ScopeDescriptor, /) -> PreparedContext:
         builder = PreparedContextBuilder()
-        scope_ids = [self.scope_id, *scope.context_references]
+        assembly = request.assembly or (ContextAssembly() if request.include_code else None)
```
Teingi: [P2] Preserve the default historical families when enabling code

When `assembly` is omitted, enabling `include_code` creates a `ContextAssembly` whose defaults contain only Memory and Experience, dropping Topic Memory from the existing default retrieval. I reproduced this through SQLite/HTTP: a query with a matching Topic Memory returns `ready`, but the same request with only `include_code: true` added returns `empty` when code is not configured. This also breaks the promised fallback to available history. Please preserve the existing default historical families when adding code context.

src/powercontext/builtin/code/engine.cjs
```diff
@@ -0,0 +1,214 @@
...
+  let truncated = starts.length > 500;
+  for (let offset = 0; offset < queue.length; offset++) {
+    const current = queue[offset];
+    if (current.path.length >= depth) continue;
+    for (const edge of graph.getIncomingEdges(current.node.id)) {
+      if (++examined > 1000 || seen.size >= 500) return { items, omitted, truncated: true };
+      if (edge.kind !== 'calls') continue;
+      const caller = graph.getNode(edge.source);
+      if (!caller || caller.language !== 'python' || ambiguous(graph, caller)) { omitted++; continue; }
+      if (seen.has(caller.id)) continue;
```
Teingi: [P2] Keep affected tests that are also in changed_paths

All definitions from `changed_paths` are placed in `seen` before traversal, so a test included in that list is skipped here even when it calls changed production code. With the real engine, `changed_paths: ["budget.py"]` returns `test_budget.py`, while `["budget.py", "test_budget.py"]` returns no items with `truncated: false`. A normal change that updates both implementation and tests therefore loses a known affected test. Please separate traversal deduplication from result selection and retain affected tests that are also starting nodes.

src/powercontext/builtin/code/engine.cjs
```diff
@@ -0,0 +1,214 @@
...
+    }
+  }
+  return { items, omitted, truncated };
+}
+
+function symbols(graph, operation, limit) {
+  const options = { languages: ['python'], kinds: ['function', 'method', 'class', 'variable', 'constant', 'property'], limit };
+  if (operation.path) options.includePatterns = [operation.path];
+  return graph.searchNodes(operation.query, options).map(result => result.node).filter(node =>
+    node.kind !== 'file' && (!operation.path || node.filePath === operation.path)
```
Teingi, line 147: [P2] Apply the file constraint before limiting symbol results

CodeGraph 1.6.0 does not apply `includePatterns` in this search path, so `limit` is applied to repository-wide results before the adapter filters by file. With `shared` defined in both `alpha.py` and `beta.py`, querying `shared` with `path: "beta.py", limit: 1` returns an empty list and `truncated: false`; increasing the limit to 20 returns the definition. Other files can therefore hide a valid match in the requested file. Please constrain the candidate set by path before applying the requested limit.

src/powercontext/builtin/code/repository.py
```diff
@@ -0,0 +1,367 @@
...
+        path: b"D " if path not in indexed else b"A " if path not in head_entries else b"M "
+        for path in indexed.keys() | head_entries.keys()
+        if indexed.get(path) != head_entries.get(path)
+    }
+    if config.include_untracked:
+        untracked = await _git(root, ("ls-files", "--others", "--exclude-standard", "-z"), config, deadline)
+        entries.extend(
+            _GitEntry(path.decode("utf-8", errors="surrogateescape"), "100644", "")
+            for path in untracked.split(b"\0")
+            if path
```
Teingi, line 194: [P2] Exclude untracked symlinks before reading file contents

Every untracked entry is labeled `100644`, so an untracked Python symlink bypasses the symlink exclusion and reaches the `O_NOFOLLOW` read. In a repository containing tracked `core.py` and untracked `linked.py -> core.py`, capture succeeds with the default settings but raises `code_repository_unavailable` with `include_untracked: true`. One link thus prevents the whole repository from being indexed instead of being omitted as documented. Please classify untracked entries without following links and skip symlinks.

Teingi:
native support: https://github.com/oceanbase/powercontext/pull/1708

PR #1625, opened 2026-09-16
feat(pi): add artifact candidate review tools

wutongyuonce:
## Summary

Add the missing Pi tools for reviewing generated Artifact candidates:

- `pc_review_approve` → `approve_artifact_candidate`
- `pc_review_reject` → `reject_artifact_candidate`
- `pc_review_revise` → `revise_artifact_candidate`

Refs #1338.

## Dependency

Depends on #1621 and should be reviewed/merged after #1621. This PR builds on the Pi candidate generation and inspection surface so the flow is generation → inspection → explicit review decision.

## Plan and final design

The original plan was to complete the candidate governance seam identified during the Pi profile audit, while keeping approval separate from installation, publication, activation, and execution.

The final design matches that plan:

1. Inspect the exact candidate and version with the existing `pc_review_list` / `pc_review_get` tools.
2. Require explicit user intent for the decision:
   - `pc_review_approve` approves the exact pending candidate/version only.
   - `pc_review_reject` rejects the exact candidate/version and requires a non-empty reason.
   - `pc_review_revise` submits an explicitly requested content/provenance revision using the exact expected version; the result remains a new reviewable candidate.
3. Keep the operations scoped and safe: all three are marked mutating, included in Pi's write-operation set, require interactive confirmation, refuse writes without a UI, inject the resolved Scope, and retain secret filtering.
4. Align the Pi TypeBox schemas with the OpenAPI limits, including closed objects, candidate/version bounds, proposal shapes, Skill package digests and size limits, nullable memory citations, and a combined maximum of 32 source/artifact references.
5. Document the decision boundary in `project-context` and the Pi guidance. Returned candidate references, versions, and Handoff values must be passed unchanged; assessment or generation never counts as approval.

This PR intentionally does not add installation, publication, activation, execution, profile promotion, or candidate generation. It only completes explicit candidate review decisions.

## Verification

- Pi tests: 14 files, 95 tests passed
- Pi TypeScript typecheck passed
- Integration manifest tests: 28 passed
- Integration manifest check passed
- Python typecheck (`ty check`) passed
- `git diff --check` passed

## Review disclosure

Implemented and verified with Codex using GPT-5.6 Sol. A separate GPT-5.6 Luna review (medium reasoning) found three schema mismatches during review; all were fixed and the final review was clean. Luna did not modify files or run formatters.


integrations/pi/plugins/powercontext/src/tools.ts
```diff
@@ -180,6 +180,54 @@ const TASK_OUTCOME = Type.Object({
...
+  instructions: Type.String({ maxLength: 131072 }),
+  validation: Type.Array(Type.String({ minLength: 1, maxLength: 2000, pattern: '^\\S(?:.*\\S)?$' }), { maxItems: 32 }),
+  package: Type.Optional(Type.Union([SKILL_PACKAGE_REFERENCE, Type.Null()])),
+  license: Type.Optional(Type.Union([Type.String({ minLength: 1, maxLength: 512 }), Type.Null()])),
+  compatibility: Type.Optional(Type.Union([Type.String({ minLength: 1, maxLength: 500 }), Type.Null()])),
+  metadata: Type.Optional(Type.Record(Type.String(), Type.String(), { maxProperties: 64 })),
+  allowed_tools: Type.Optional(Type.Union([Type.String({ minLength: 1, maxLength: 2000 }), Type.Null()])),
+}, { additionalProperties: false })
+const CANDIDATE_PROPOSAL = Type.Union([EXPERIENCE_PROPOSAL, SKILL_PROPOSAL])
+const REVISE_CANDIDATE = Type.Union(Array.from({ length: 33 }, (_, sourceMax) => Type.Object({
```
Teingi: [P2] Keep the revision tool's root schema an object

This produces a root schema containing only `anyOf`. I reproduced with the pinned Pi 0.82.1 that its Anthropic adapter reads the root `properties` and `required` fields and converts this tool to `input_schema: { type: 'object', properties: {}, required: [] }`. Claude therefore receives none of the required revision parameters, and a call following that empty definition fails local argument validation. The current tests pass complete arguments directly, so they miss this conversion. Could we use a top-level `Type.Object`, enforce the combined reference limit before execution, and add a regression test for the provider-facing tool schema?

wutongyuonce: @Teingi Thanks for catching this. Fixed and pushed in 1f9da4a1. `pc_review_revise` now exposes a provider-compatible top-level object schema, while preserving the per-array API limits and enforcing `source_refs + artifact_refs <= 32` before execution. Added regression coverage for the provider-facing root schema and the combined reference limit. Local Pi tests (95), TypeScript typecheck, manifest tests (28), manifest check, and `ty check` all pass.

Teingi:
resolve conflicts

Teingi (approved):
LGTM

PR #1632, opened 2026-09-16
fix(codex): verify native MCP authentication on Windows

### LINKED ISSUE #1628: bug(codex): setup reports success while native MCP tools are unavailable
## Describe the bug

On Windows, `powercontext setup codex` and `powercontext doctor codex` can report a healthy, enabled plugin while the Codex host has no PowerContext MCP tools such as `remember_memory` and `search_memory`.

The PowerContext service itself is healthy and an authenticated direct Streamable HTTP MCP client can connect to `http://127.0.0.1:8000/mcp` and discover the server tools. The failure is at the Codex host integration boundary: native MCP authentication depends on the environment of the process that starts Codex.

***I use chatGPT desktop 26.908.70816, maybe this bug only occures in chatGPT desktop***

## Steps to reproduce

1. On Windows, install/setup the Codex integration with an authenticated local PowerContext server.
2. Install codex plugin (the official `powercontext setup codex` is not supported for chatGPT desktop)
3. Start or restart Codex Desktop without ensuring that `POWERCONTEXT_CODEX_AUTHORIZATION` is present in the Desktop process environment.
4. Ask Codex to use PowerContext or inspect the available MCP tools.
5. Observe that the PowerContext tools are unavailable, although the setup/doctor result reports the plugin as enabled.

Additional observations from the same installation:

- `codex mcp list --json` reports the PowerContext server as enabled with `auth_status: "unsupported"` when the referenced environment variable is absent.
- Loading the project `.env` into the process before running the same command changes the status to `auth_status: "bearer_token"`.
- An authenticated direct MCP client successfully discovers 33 PowerContext tools.
- The focused plugin tests pass: `88 passed, 4 skipped` for `tests/codex_plugin/test_contract.py` and `tests/codex_plugin/test_client_transport.py`.

## Expected behavior

`setup codex`/`doctor codex` should verify the native Codex MCP connection (including authentication and tool discovery), or clearly report the missing host environment/authentication. A setup-configured credential should either be usable by the native MCP client or the required Desktop configuration should be made explicit and actionable.

## Actual behavior

The plugin is reported as installed and enabled, but native MCP tools are unavailable. Restarting the service or Codex does not provide the credential to a Desktop process that did not inherit the terminal environment.

## Code analysis

- `src/powercontext/cli/system.py:429-440` calls `run_codex_diagnostics()` and reports success after plugin registration only.
- `src/powercontext/cli/system.py:1331-1377` implements `run_codex_diagnostics()` using only `codex plugin list`; it does not inspect `codex mcp`, perform an MCP initialize/list-tools probe, or validate the native MCP authentication path.
- `src/powercontext/cli/system.py:1088-1092` persists setup authorization through `configure_stored_authorization()`.
- `integrations/codex/plugins/powercontext/settings.py:80-93` lets the Python Hook read that PowerContext-owned credential file, but the native MCP client reads `integrations/codex/plugins/powercontext/.mcp.json`, which only references the process environment variable `POWERCONTEXT_CODEX_AUTHORIZATION`.
- `src/powercontext/cli/system.py:1102-1115` updates only the installed MCP URL; it does not bridge the persisted credential into the native Codex MCP process.
- The MCP entry is `required: false`, so a native MCP connection/authentication failure is non-blocking and can remain invisible during setup.

The documentation already notes that Desktop apps may not inherit terminal environment variables, but the setup/diagnostic path does not surface this as a verifiable failure mode.

## Proposed direction

- Add a native MCP smoke check to setup/doctor (at least configuration, environment resolution, initialize, and list-tools), with redacted actionable diagnostics.
- Surface `authorization_state` and missing/invalid `POWERCONTEXT_CODEX_AUTHORIZATION` in the normal setup output.
- Define a supported credential path for Codex Desktop, or make the host-level environment requirement explicit in the setup flow while preserving the current no-token-in-static-config security rule.
- Add a Windows end-to-end acceptance test that verifies authenticated native MCP tool discovery, not just plugin registration and static `.mcp.json` validation.

Related: #1538

## Environment

- OS: Windows / PowerShell
- ChatGPT: 26.908.70816
- PowerContext revision: `b0ec6290`

beatenevo:

Closes #1628 .

# Rationale for this change
On Windows, `powercontext setup codex` and `powercontext doctor codex` could report that the PowerContext plugin was installed, enabled, and healthy even when the Codex host could not load any PowerContext MCP tools.

The PowerContext Server itself could be healthy, and an authenticated direct Streamable HTTP MCP client could successfully initialize the server and discover its tools. The failure occurred at the Codex host integration boundary:

- `setup codex` persisted authorization in PowerContext-owned credential storage.
- The Python Hook could read that stored credential.
- The native Codex MCP client instead resolved `POWERCONTEXT_CODEX_AUTHORIZATION` from the environment inherited by the Codex process.
- Codex Desktop launched outside the setup terminal did not necessarily inherit that environment variable.
- The MCP entry is optional, so authentication or connection failures did not prevent Codex from starting.
- The previous diagnostics only checked plugin registration through `codex plugin list`; they did not verify the native MCP configuration, authentication path, initialization, or tool discovery.

As a result, setup and doctor could produce a false-positive result while tools such as `remember_memory` and `search_memory` were unavailable.

This change makes setup and diagnostics verify the same native MCP path used by Codex and provides an explicit Windows Desktop credential path without placing secrets in `.mcp.json` or static MCP headers.

## Windows Codex Desktop authorization

- Continue storing setup-managed authorization in the existing URL-bound credential file at `~/.codex/powercontext/credentials.json`.
- On Windows, synchronize the normalized bearer header to the current user's `POWERCONTEXT_CODEX_AUTHORIZATION` environment variable under `HKCU\Environment`.
- Broadcast a Windows environment-change notification after updating the user environment.
- Preserve the requirement to restart Codex Desktop so the new process inherits the updated environment.
- Keep `.mcp.json` token-free. It continues to reference the environment variable:

```json
{
  "env_http_headers": {
    "Authorization": "POWERCONTEXT_CODEX_AUTHORIZATION"
  }
}
```

- Validate authorization values before writing or comparing them.
- Report Windows environment update failures without exposing credential values.

## Native Codex MCP diagnostics

`powercontext doctor codex` now verifies:

1. The Codex CLI is available.
2. The PowerContext plugin is installed and enabled.
3. `codex mcp list --json` contains an enabled PowerContext MCP entry.
4. The MCP entry uses the expected environment-backed authorization.
5. The current process or Windows user environment matches the URL-bound setup credential.
6. Codex's native `app-server` initializes successfully.
7. `mcpServerStatus/list` reports the PowerContext server.
8. The native MCP connection discovers `remember_memory` and `search_memory`.

The diagnostic output now includes separate redacted checks for:

- `mcp_configuration`
- `authorization`
- `mcp_tools`

The native probe does not treat `auth_status` alone as proof of a working connection. This matters because `codex mcp list --json` can report `auth_status: "unsupported"` when the current shell lacks the variable, even though a restarted Desktop process can resolve it from the Windows user environment.

## Setup behavior

- `powercontext setup codex` synchronizes a matching stored credential into the Windows user environment.
- Setup continues to update the installed plugin's MCP endpoint.
- Plain-text setup output now reports the authorization state.
- Setup fails when the native MCP connection cannot initialize or does not expose the required tools, rather than succeeding after plugin registration alone.

## Hook credential resolution

- Fix URL matching between a stored base Server URL and the effective `/mcp` endpoint.
- Keep the setup-managed credential available to the Python Hook.
- Ensure an explicit process-level authorization value takes precedence over the stored credential.

## Documentation

Update the English and Chinese Codex integration documentation and plugin README to describe:

- the setup-managed credential path;
- Windows user-environment synchronization;
- the required Codex Desktop restart;
- the native MCP diagnostic behavior;
- the no-token-in-static-config security rule;
- PowerShell commands for authenticated Windows setup;
- the distinction between updating the user environment and updating already running processes.

## Tests

- Add focused coverage for Windows user-environment authorization persistence and resolution.
- Add diagnostics coverage for valid and invalid MCP configuration, missing authorization, URL mismatches, missing tools, and native app-server initialization.
- Add plugin settings coverage for base URL versus `/mcp` URL matching and process-level credential precedence.
- Add an opt-in Windows end-to-end acceptance test that removes the process authorization variable and verifies native Codex MCP tool discovery through the Windows user environment.

Yes.
On Windows, ```powercontext setup codex``` now writes the matching complete bearer header to the current user's ```POWERCONTEXT_CODEX_AUTHORIZATION``` environment value. Users must restart Codex Desktop after setup so the new Desktop process inherits it.
```powercontext doctor codex``` now performs a real native MCP smoke check and can fail when:
- the native MCP entry is missing or disabled;
- environment-backed authorization is not configured;
- the setup credential does not match the effective host environment;
- Codex cannot initialize the MCP server;
- required PowerContext tools are not discovered.
The setup output also reports the authorization state.
There are no breaking public Python API changes and no persisted credential format changes. ```.mcp.json``` remains token-free, and diagnostics do not expose credential values.
On non-Windows platforms, Codex must still be started from an environment containing ```POWERCONTEXT_CODEX_AUTHORIZATION``` when authentication is enabled

# How was this change tested?
- Ran the focused Codex integration and authorization test suite:

  ```powershell
  .\.venv\Scripts\pytest.exe `
    tests\codex_plugin\test_contract.py `
    tests\codex_plugin\test_client_transport.py `
    tests\test_authorization.py `
    tests\test_system_cli.py `
    tests\test_setup_transport.py `
    tests\e2e\test_codex_native_mcp_live.py `
    -q
  ```

  Result: `176 passed, 8 skipped`.

- Ran the opt-in Windows native MCP acceptance test:

  ```powershell
  .\.venv\Scripts\pytest.exe `
    tests\e2e\test_codex_native_mcp_live.py `
    --run-real-e2e `
    -q
  ```

  Result: `1 passed`.

- Ruff linting, Ruff formatting checks, Ty type checks, and Codex plugin validation all passed.

- Manually removed the process-level authorization variable and ran:

  ```powershell
  Remove-Item Env:POWERCONTEXT_CODEX_AUTHORIZATION -ErrorAction SilentlyContinue
  uv run --frozen powercontext doctor codex --json
  ```

  The native Codex MCP probe returned `ok: true` and discovered 33 PowerContext tools. After restarting Codex Desktop, `remember_memory` and `search_memory` were available to the host.

- A full pytest run was attempted on Windows , `455 passed, 33 skipped`.

# AI usage statement
OpenAI Codex was used.

### CODE REVIEW

**Comment by Teingi** [comment]
tests falied

File: src/powercontext/cli/system.py [hunk 1]
```diff
@@ -1375,6 +1391,292 @@ def run_codex_diagnostics() -> dict[str, Diagnostic]:
...
+    if expected_authorization is None:
+        native_authorization = process_authorization or desktop_authorization
+        authorization_ok = authorization.status == "not_configured"
+        authorization_detail = (
+            "no setup-managed credential; the native probe will verify the effective host environment"
+            if authorization_ok
+            else f"stored credential state is {authorization.status}; rerun `powercontext setup codex`"
+        )
+    else:
+        process_matches = process_authorization == expected_authorization
```
Teingi: [P2] Respect the explicit process credential in the native probe

If credentials.json contains an old token while POWERCONTEXT_CODEX_AUTHORIZATION contains the valid replacement, doctor reports authorization failure and skips MCP discovery. I reproduced this with Codex CLI 0.153.4 and a real SQLite-backed server: the native client discovers 33 tools with the process credential, while doctor rejects the same configuration. Conversely, a matching Windows user value causes the probe to replace a conflicting process value, masking its failure. Please give an explicit process credential precedence and probe that effective value; report stale saved credentials separately, and distinguish a current-process check from a check of the environment expected after restarting Desktop.

src/powercontext/cli/authorization.py
```diff
@@ -188,13 +190,75 @@ def setup_server_url(host: str, default: str) -> str:
...
+        return normalize_authorization(value)
+    except ValueError:
+        return None
+
+
+def _write_windows_user_environment(name: str, value: str) -> None:
+    import ctypes
+    import winreg
+
+    with winreg.CreateKeyEx(winreg.HKEY_CURRENT_USER, "Environment", access=winreg.KEY_SET_VALUE) as key:
```
Teingi: [P2] Guard the Windows APIs inside these helpers

The platform checks in the callers do not narrow the platform when ty analyzes these helper bodies. On Linux, ty reports 13 unresolved-attribute errors for winreg and ctypes here and in _read_windows_user_environment, so the required quality job fails. This also reproduces locally with the locked dependencies. Please add platform guards that the type checker can recognize inside the helpers, or isolate these calls in an appropriately guarded implementation.

tests/test_system_cli.py
```diff
@@ -155,6 +393,81 @@ def test_setup_codex_persists_setup_only_token_in_codex_owned_storage(
...
+                    "pluginId": "powercontext@powercontext",
+                    "installed": True,
+                    "enabled": True,
+                }
+            ]
+        },
+    )
+    credential = tmp_path / "codex" / "powercontext" / "credentials.json"
+    credential.parent.mkdir(parents=True)
+    credential.write_text(
```
Teingi: [P2] Give the credential fixtures owner-only permissions

These write_text calls create credentials.json with mode 0644 under the usual POSIX umask. read_stored_authorization therefore returns unsafe_permissions before reaching the behavior the tests intend to exercise. This affects this test, the URL-mismatch test, and the unavailable-to-host test, and fails every Python 3.11–3.14 CI job. The focused suite reproduces all three failures locally. Please use write_stored_authorization to create the fixtures, or explicitly set mode 0600 as the plugin contract tests already do.

Teingi (commented):
Found three issues below. Validation used Codex CLI 0.153.4 with a disposable SQLite-backed PowerContext server on Linux. The focused suite reports 176 passed, 3 failed, and 5 skipped; ty reports 13 errors. Windows registry behavior was simulated rather than tested on a native Windows host.

beatenevo:
> tests falied


宿舍熄灯了明早起来修嘿嘿

Teingi:
> > tests falied
> 
> 宿舍熄灯了明早起来修嘿嘿

good night!

tests/test_system_cli.py
beatenevo: The affected credential fixtures now use write_stored_authorization instead of Path.write_text, which applies the production owner-only permission behavior on POSIX and avoids the unintended unsafe_permissions result.

src/powercontext/cli/system.py
beatenevo: An explicit valid POWERCONTEXT_CODEX_AUTHORIZATION value now takes precedence and is always used for the native MCP probe. The setup-managed credential and the Windows user environment are reported separately through current_process, setup_managed, and desktop_restart checks.
A stale stored credential no longer blocks tool discovery, and a conflicting Windows user value no longer replaces the current-process credential. Added regression coverage for all three combinations, including redaction assertions.

src/powercontext/cli/authorization.py
beatenevo: The Windows-only winreg and ctypes members are now resolved dynamically inside the guarded helpers, so Linux type checking no longer attempts to resolve platform-specific attributes.

Teingi (commented):
The previous three findings are fixed. The focused suite now passes (182 passed, 5 skipped), and the type checks pass. One native authentication diagnostic issue remains below.

src/powercontext/cli/system.py
```diff
@@ -1375,6 +1391,398 @@ def run_codex_diagnostics() -> dict[str, Diagnostic]:
...
+
+    authorization = read_stored_authorization(
+        credential_path("codex"), server_url=mcp_url.rstrip("/").removesuffix("/mcp")
+    )
+    process_value = os.environ.get("POWERCONTEXT_CODEX_AUTHORIZATION")
+    process_authorization: str | None = None
+    process_state = "not_configured"
+    if process_value is not None:
+        try:
+            process_authorization = normalize_authorization(process_value)
```
Teingi: [P2] Validate the process header without repairing it for the probe

If POWERCONTEXT_CODEX_AUTHORIZATION contains only the token, normalize_authorization silently adds the Bearer prefix, and _probe_codex_mcp_status injects that repaired value into its child process. The actual Codex client reads the original environment header unchanged, so doctor can report a working connection while native MCP authentication fails. With Codex CLI 0.153.4 and a real HTTP/SQLite-backed server, the original bare-token header discovers 0 tools, while doctor reports authorization and MCP checks as OK and discovers 33 tools. Please require a complete Bearer header here and probe the actual value without rewriting it; malformed process configuration should fail validation.

src/powercontext/cli/system.py
```diff
@@ -1375,6 +1391,402 @@ def run_codex_diagnostics() -> dict[str, Diagnostic]:
...
+    process_value = os.environ.get("POWERCONTEXT_CODEX_AUTHORIZATION")
+    process_authorization: str | None = None
+    process_state = "not_configured"
+    if process_value is not None:
+        try:
+            normalized_process_authorization = normalize_authorization(process_value)
+        except ValueError:
+            process_state = "invalid"
+        else:
+            if normalized_process_authorization != process_value:
```
Teingi: [P2] Accept case-insensitive Bearer schemes without rewriting the header

normalize_authorization always emits capitalized Bearer, so this equality check rejects a valid header such as bearer <token>. I reproduced this with Codex CLI 0.153.4 and a real HTTP/SQLite-backed server: the native client discovers 33 tools using the lowercase scheme, while doctor reports authorization=failed and skips MCP discovery with the same environment. The existing hook also accepts the scheme case-insensitively. Please validate the complete header syntax with a case-insensitive scheme check, then pass the original value unchanged to the probe. Comparing the entire header with its canonicalized form confuses valid casing differences with malformed input.

Teingi (commented):
The bare-token false positive is fixed, and the earlier credential-precedence checks still pass. The focused suite reports 183 passed and 5 skipped, with type checks passing. The new equality check introduces one regression below.

src/powercontext/cli/system.py
beatenevo: done.

Teingi: I rechecked, but the PR head is still `8e9c8e1` and the equality check at `system.py:1578` is unchanged. The lowercase `bearer` case still reproduces. Has the fix been pushed to this PR branch?

beatenevo: sorry, my bad. I thought I had committed the changes. I even checked my local `git status` after sending “done” and was wondering what was going on. I just realized that I forgot to commit my changes on my win...3333

Teingi (approved):
LGTM

PR #1633, opened 2026-09-16
feat(pi): add Memory Changes and Stats tools

wutongyuonce:
## Summary

Add two native read-only Pi tools to expose existing PowerContext operations:

- `pc_memory_changes` → `list_memory_changes`, with an optional non-negative exclusive `since_revision` cursor.
- `pc_stats` → `get_stats`, with an optional `today`, `7d`, or `30d` period (default `30d`).

## Design: plan and final implementation

The plan was to expose the existing generated operations without adding a second transport or state layer, while keeping Scope resolution authoritative in the shared Pi invoke path. The final implementation follows that plan:

- TypeBox schemas are closed and aligned with the OpenAPI input limits.
- `list_memory_changes` receives the resolved current Scope through `scopeMode: current`.
- `get_stats` accepts only the period at the Pi boundary; its `scopeMode: selection` metadata causes the shared invoke layer to replace any selection with the resolved exact Scope. This keeps the native tool current-Scope-only and prevents caller-supplied cross-Scope selection.
- Both tools are read-only: they are not in `WRITE_OPERATIONS`, require no confirmation, create no durable state, and do not alter profile capabilities.
- Returned history and statistics remain untrusted observations. No response transformation or new caching was added.
- Pi guidance, the project-context Skill, the capability manifest, and English/Chinese integration docs describe the same boundary.

## Non-goals

- No new slash command; `/pc stats` already exists.
- No cross-Scope stats query from native Pi tools.
- No profile promotion or changes to the server operations/OpenAPI.

## Verification

- Pi tests: 14 files, 98 tests passed.
- TypeScript typecheck passed.
- Integration manifest tests: 28 passed.
- `integration_manifest.py --check` passed.
- `ty check` passed.
- `git diff --check` passed.
- GPT-5.6 Luna medium read-only review found no actionable issues.

Implementation and verification used Codex with GPT-5.6 Sol; the separate Luna review was read-only.

Refs #1338


integrations/pi/plugins/powercontext/src/tools.ts
```diff
@@ -53,6 +53,12 @@ const SEARCH_MODES = Type.Union([
   Type.Literal('vector'),
   Type.Literal('hybrid'),
 ])
+const STATS_PERIOD = Type.Union([
+  Type.Literal('today'),
+  Type.Literal('7d'),
+  Type.Literal('30d'),
+])
+const NON_NEGATIVE_REVISION = Type.Union([Type.Integer({ minimum: 0 }), Type.Null()])
```
Teingi: [P2] Preserve the cursor value through Pi's argument validation

Pi 0.82.1's `validateToolArguments` coerces `{ since_revision: null }` to `{ since_revision: 0 }` with this schema. These values have different server semantics: `null` returns only the latest revision, while `0` requests the entire history. With three revisions in an authenticated HTTP/SQLite fixture, the direct server request returned `[3]`, but the same input through the Pi tool returned `[1, 2, 3]`. This silently expands the query and can produce increasingly large tool responses. The validator also converts `1.5` to `1`, despite the intended integer constraint.

Please use a schema that preserves these values through Pi's actual validator and add a regression test through `validateToolArguments`; the current `Value.Check` tests miss this conversion. I verified that `type: ["integer", "null"]` with `minimum: 0` preserves both `null` and `0` and rejects fractions. Simply reversing the Union members converts `0` to `null` instead.

wutongyuonce: @Teingi Fixed in 86f5cf0. The nullable cursor now uses the JSON Schema type array `['integer', 'null']` with `minimum: 0`, so Pi 0.82.1 preserves `null` versus `0` and rejects fractional values instead of coercing them. I added a regression through `validateToolArguments` covering `null`, `0`, and `1.5`; the Pi test suite now passes 99 tests.

Teingi (approved):
LGTM

PR #1636, opened 2026-09-17
Codex/fix windows command encoding

### LINKED ISSUE #1627: bug(windows): icacls decoding failure crashes protected environment-file validation
## Describe the bug

On Chinese Windows, loading a protected service environment file can crash with an unhandled `AttributeError` when Python UTF-8 mode is enabled and `icacls.exe` emits CP936 output.

`_validate_windows_protection()` uses `capture_output=True, text=True` without specifying an encoding. On the tested runtime, decoding fails in a subprocess reader thread with `UnicodeDecodeError`, but the caller receives `returncode == 0` and `stdout is None`. Calling `result.stdout.splitlines()` then crashes instead of returning a `ProtectedEnvironmentFileError`.

UTF-8 mode alone is not sufficient to reproduce: the command's actual output and console environment also matter.

## Steps to reproduce

On Chinese Windows with ACP/OEMCP 936, run this from the repository root after installing development dependencies. The isolated child uses `CREATE_NO_WINDOW`; no existing service or scheduled task needs to be changed.

```powershell
@'
import subprocess
import sys

for mode in ["utf8", "utf8=0"]:
    result = subprocess.run(
        [sys.executable, "-X", mode, "-m", "pytest",
         "tests/test_service_environment.py", "-q", "--tb=short"],
        capture_output=True,
        creationflags=subprocess.CREATE_NO_WINDOW,
    )
    print("MODE", mode, "EXIT", result.returncode)
    print(result.stdout.decode("utf-8", errors="backslashreplace"))
    print(result.stderr.decode("utf-8", errors="backslashreplace"))
'@ | uv run python -
```

Observed results:

| Execution environment | Result |
| --- | --- |
| Inherited console, Python UTF-8 mode enabled | 5 passed, 7 skipped |
| No-window child, Python UTF-8 mode enabled | 5 failed, 1 passed, 6 skipped |
| No-window child, Python UTF-8 mode disabled | 5 passed, 7 skipped |

The inherited console reported output code page 65001 and the probed command output was ASCII. The no-window child reported output code page 936; captured `icacls` output contained non-ASCII bytes that failed strict UTF-8 decoding and decoded successfully with `oem`.

## Expected behavior

- A correctly protected environment file loads successfully in supported Windows execution environments, including no-window processes with Python UTF-8 mode enabled.
- Files granting access to unexpected accounts remain rejected.
- Command execution or decoding failures produce a clear domain error without bypassing ACL validation or raising an unhandled `AttributeError`.

## Actual behavior

```text
src/powercontext/service/environment.py:147
    for line in result.stdout.splitlines():
AttributeError: 'NoneType' object has no attribute 'splitlines'
```

A separate minimal probe of the real `icacls.exe` command captured:

```text
returncode=0
stdout_is_none=True
thread_errors=['UnicodeDecodeError']
```

The test setup also invokes `icacls` with `text=True`, so it can emit reader-thread decoding warnings independently of the production call.

## Environment

- OS: Chinese Windows 10, build 19045.
- Python: CPython 3.13.12, 64-bit.
- System ACP/OEMCP: 936 / 936.
- Reproduced on checkout `1dd71908`.
- The same faulty code was also verified on upstream master `b0ec6290f6ce2fce9d26e768e30e4c67f79b506e`.

## Related code and repair considerations

- [Protected environment-file validation](https://github.com/oceanbase/powercontext/blob/b0ec6290f6ce2fce9d26e768e30e4c67f79b506e/src/powercontext/service/environment.py#L125-L147).
- The `schtasks.exe` runner has the same implicit decoding assumption. In a read-only query for a nonexistent task under the reproduction environment, stderr decoding failed, leaving `stderr=None` and `stdout=''`; `_command_detail(result.stderr or result.stdout)` returned an empty diagnostic. This observation did not produce a second `AttributeError`.
- Both `whoami.exe` call sites have the same assumption, but their output was ASCII for the tested account; a failure there has not been reproduced.

Consider capturing bytes and decoding explicitly in the calling thread, using a verified encoding contract for each command and mapping failures to the existing service/environment-file exceptions. ACL and account data must not be silently decoded with ignored errors or accepted when validation is incomplete.

`encoding="oem"` worked for the CP936 reproduction, but OEM code page and current console output code page are not interchangeable. A blanket replacement has not been established as a general fix.

Regression coverage should protect valid-file loading, rejection of unexpected ACL principals, and clear decoding-failure diagnostics. Existing Windows test helpers with the same implicit decoding assumption should be addressed as needed.

Related but distinct: #1585 / #1591 concern Task Scheduler query timeouts, and #1305 / #1326 concern UTF-8 hook input.

Doar999:

Closes #1627.

# Rationale for this change

`icacls.exe`, `whoami.exe`, and `schtasks.exe` were invoked with `capture_output=True, text=True` and no explicit encoding. Python then decodes with `locale.getpreferredencoding(False)`, which is unrelated to the code page the child console actually writes.

Two failures follow from that mismatch:

1. **Crash instead of a domain error.** In a no-window process under Python UTF-8 mode, the console reports the OEM code page (936 on a Chinese Windows). Strict UTF-8 decoding then fails inside the `subprocess` reader thread, where the exception is lost. The caller receives `returncode == 0` with `stdout is None`, and `_validate_windows_protection()` crashes on `result.stdout.splitlines()` with `AttributeError: 'NoneType' object has no attribute 'splitlines'` — the ACL check never completes and never reports why.

2. **Silent misclassification on any system where ACP differs from OEMCP.** On, for example, an English Windows (ACP 1252 / OEMCP 437), decoding succeeds but produces mojibake. An account name containing non-ASCII characters then fails to match the allowed principals, and a correctly protected file is rejected with `--env-file ACL grants access to an unexpected account`. This second case is not in the issue report; it was found while fixing the first.

Choosing an encoding is not sufficient on its own, because the child's code page depends on the console it inherits: the same call yields 65001 under a terminal running `chcp 65001` and 936 in a no-window process. The child's console has to be made deterministic before its output can be decoded reliably.

**New helper — `src/powercontext/service/_windows_command.py`**

`run_windows_command(command, *, timeout)` is the single entry point for the Windows service's native command-line tools:

- Passes `CREATE_NO_WINDOW`, so the child always gets a fresh console using the system OEM code page instead of inheriting the caller's `chcp` setting.
- Captures bytes rather than text, then decodes `stdout` and `stderr` with the `oem` codec **on the calling thread**. A decoding failure therefore reaches the caller as `subprocess.SubprocessError` instead of being swallowed by a reader thread.
- Also maps `LookupError`, so a call from a non-Windows path reports the missing OEM codec as a command failure rather than escaping the caller's `except` clause.
- Normalises `\r\n` and lone `\r` to `\n`, preserving the previous `text=True` universal-newlines behaviour. This matters in practice: `schtasks /Query /XML` emits `\r\r\n`.

**Call sites migrated (all four native Windows command invocations in `src/`)**

| File | Command |
| --- | --- |
| `service/environment.py` | `icacls.exe <path>` |
| `service/environment.py` | `whoami.exe /user /fo csv /nh` |
| `service/adapters/windows.py` | `schtasks.exe` (all subcommands, via `_run`) |
| `service/adapters/windows.py` | `whoami.exe /user /fo csv /nh` |

A repository-wide grep confirms no other `subprocess` call in `src/` targets a Windows native command; the remaining ones are `git`, `systemctl`, and fixed CLI executables.

**Tests**

- `test_windows_env_loader_handles_native_output_encoding` — parametrised over `-X utf8=0/1` × console output code page `default/65001`, run in a child process. This is the regression test for the reported crash and covers the console-inheritance dimension that a single-mode test would miss.
- `test_windows_env_loader_reports_undecodable_acl_output` — undecodable `icacls` output must surface as `ProtectedEnvironmentFileError`, not `AttributeError`.
- `test_windows_support_preserves_native_command_error` — an OEM-encoded `schtasks` diagnostic must survive into `support()` detail.
- Test helpers that themselves shell out to `icacls`/`whoami` no longer pass `text=True`; they were emitting reader-thread decoding warnings independently of the production call (noted in the issue).
- `test_service_install_cli_expands_the_environment_file_home_directory` now also sets `USERPROFILE`. Windows `expanduser()` ignores `HOME`, so this test failed on Windows before this PR as well — see "Out of scope" below.

**Diff:** 5 files, +177 / −41.

No API, CLI, configuration, or persisted-format change. Behaviour changes, all Windows-only:

- Loading a protected `--env-file` now succeeds in execution environments that previously crashed — notably any no-window process (scheduled task, service host, GUI launcher) running under Python UTF-8 mode.
- Command execution and decoding failures now produce `ProtectedEnvironmentFileError` / `ServiceError` with the tool's own diagnostic text, instead of an unhandled `AttributeError`. **ACL validation is never skipped on a decoding failure** — the file is rejected.
- Non-ASCII account names and diagnostics are now decoded correctly, including on systems where ACP differs from OEMCP.
- Native commands no longer flash a console window when invoked from a process without a console.

No documentation update appears necessary: this restores documented behaviour rather than changing it. Happy to add a note if maintainers prefer one.

# How was this change tested?

Environment: Windows 10 22H2, `zh-CN`, ACP/OEMCP 936, Python 3.13.12, `uv`.

**Regression reproduced on the merge base first.** In a temporary worktree at `ee5c076e`, running `tests/test_service_environment.py` inside a `CREATE_NO_WINDOW` child under `-X utf8=1` produces the reader-thread failure from the issue:

```
PytestUnhandledThreadExceptionWarning: Exception in thread Thread-59 (_readerthread)
  File "subprocess.py", line 1615, in _readerthread
    buffer.append(fh.read())
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd2 in position 0: invalid continuation byte
```

These warnings are gone on this branch.

**Test suite** — `tests/test_service.py` + `tests/test_service_environment.py`, executed inside a `CREATE_NO_WINDOW` child so the console code page matches the failing environment:

| Mode | Result |
| --- | --- |
| `-X utf8=1` | 108 passed, 12 skipped |
| `-X utf8=0` | 108 passed, 12 skipped |

(Before this PR, the same command reported 107 passed / 1 failed in both modes; the failure was the pre-existing `USERPROFILE` issue described below.)

**Static checks**

- `ruff check` — passed
- `ruff format --check` — passed
- `uv run ty check src/powercontext/service` — passed
- `prek` pre-commit hooks — passed (case/merge-conflict, end-of-file, trailing-whitespace, ruff check, ruff format)

One caveat, stated explicitly: the repository-wide `ty check` hook **cannot run clean on Windows**, and I skipped it locally. It reports 10 diagnostics, all in files this PR does not touch (`builtin/runtime/artifact_processing.py`, `cli/opencode.py`, `tests/builtin/runtime/test_artifact_processing.py`), all of the form `Module 'os' has no member 'WNOHANG'` — POSIX-only APIs that do not resolve on a Windows host. They are present on `ee5c076e` too and should pass on Linux CI. The scoped check over the changed package passes.

**Manual probes**

- `schtasks /Query /XML` output on this host: single-byte ASCII carrying an `encoding="UTF-16"` declaration, with `\r\r\n` line endings. Verified that `ET.fromstring` accepts the resulting `str` (CPython's ElementTree tolerates the declaration) and that the newline normalisation reproduces the previous `text=True` result exactly.
- Verified `CREATE_NO_WINDOW` introduces no interactive hang: every `schtasks` mutation already passes `/F`.

**Not covered:** the `LookupError` branch has no test. The `oem` codec always exists on Windows, so triggering it would require patching `bytes.decode`; the branch is a one-line defensive guard and I judged the test noise not worth it. Happy to add one if a maintainer disagrees.

# Out of scope

`test_service_install_cli_expands_the_environment_file_home_directory` failed on Windows before this PR — it sets `HOME`, which `ntpath.expanduser()` ignores in favour of `USERPROFILE`, so `~/powercontext.env` resolved outside `tmp_path` and the CLI exited with a usage error. Confirmed failing at `ee5c076e` in a clean worktree, so it is unrelated to the encoding defect.

The one-line fix is included here (commit `605fdc77`) because it is Windows-specific and otherwise leaves this branch red on Windows, which would obscure the regression tests above. It is isolated in its own commit and can be split into a separate PR on request.

# AI usage statement

OpenAI Codex (GPT-6) assisted with reproduction, implementation, regression tests, and validation.


tests/test_service_environment.py
```diff
@@ -104,13 +109,77 @@ def test_secure_env_loader_accepts_owned_0600_regular_file(tmp_path: Path) -> No
         assert loaded.identity.owner_sid is None
 
 
+@pytest.mark.skipif(os.name != "nt", reason="Exercises Windows command output in an isolated console")
```
Teingi, line 121: [P2] Run the new encoding regressions in Windows CI

All six new encoding cases are Windows-only, but neither Windows Python test job includes these files. The full suite runs on Linux and skips them, so the current green checks do not exercise these regressions. Please add `tests/test_service.py` and `tests/test_service_environment.py` to the Windows test job.

Doar999: Done — and it turned up two things worth reporting.

Both files are now in windows-unit-portability (fe9c805).

That immediately failed 13 cases, and not for encoding reasons. Hosted Windows runners execute jobs under an elevated token whose default owner is the Administrators group, so the env files pytest creates are not owned by the account itself, and load_protected_environment_file rejected every fixture before reaching a single encoding assertion. Fixed by setting the owner before restricting the ACL (d63a74f), mirroring _environment_file in tests/native/test_personal_service_lifecycle.py, which has carried that exact call since #1441 for the same reason — the two unit fixtures were written alongside it but never ran on Windows CI, so they missed it.

Running these cases in CI turned out to be necessary but not sufficient. Hosted runners report an OEM code page of 437, where icacls writes the fixture's non-ASCII path as question marks and every byte decodes successfully — so the four parametrized cases pass against the pre-fix text=True capture just as well as against the fix. And test_windows_env_loader_reports_undecodable_acl_output, the one case that does force a decode failure, skips on 437 for the same reason. The regressions were only ever exercised on a DBCS code page.

So I added a code-page-independent case (58685a7): it drives run_windows_command with a child process that writes 0x81 — undefined in the ANSI code page an English console reports, an incomplete lead byte in a DBCS one — and asserts the caller gets either a decoded string or a reported command failure, never returncode == 0 with stdout as None. I verified it fails against the pre-fix implementation and passes against this one, on either code page, with no skip.

The only red is macos-launchagent, It is unrelated to this change: that job installs a built wheel and runs only tests/native/test_personal_service_lifecycle.py, which does not collect either file I touched. Could someone re-run that job?

Teingi (approved):
LGTM

PR #1640, opened 2026-09-17
fix(sqlite): batch vector completeness checks

### LINKED ISSUE #1631: bug(sqlite): avoid N+1 physical-vector probes during vector completeness checks
### Describe the bug

SQLite vector search performs N+1 physical-vector probes while checking whether the vector projection is complete.

`MemoryService._select_search_mode()` checks whether vector search is available, and `SQLiteMemoryVectorIndex.search()` checks completeness again immediately before executing the vector query. These checks protect separate Service and index-layer contracts, so the issue is not that one of them exists. The issue is that each `vector_complete()` call performs one query per vector metadata row.

For one selected Memory whose vector projection is complete, with `N` active Entries and `N` corresponding metadata rows, the current completeness check performs:

```text
1 query: load active Entry heads
1 query: load vector metadata
N queries: probe physical vectors by vector_id
---------------------------------------------
N + 2 queries
```

A Vector search invokes the check twice:

```text
2 × (N + 2) = 2N + 4 queries
```

In the local single-Memory measurement below, the rest of the Vector-search path issued 27 fixed statements, so the full request total was `2N + 31`.

### Steps to reproduce

1. Start the built-in Runtime with the SQLite vector backend and a deterministic local embedding model.
2. Warm up Runtime initialization and database connection setup before counting.
3. Create one Memory containing N active Entries.
4. Register a SQLAlchemy `before_cursor_execute` listener.
5. Start counting immediately before calling:

   ```python
   await memory_service.search("query", memories=(memory,), mode="vector")
   ```

6. Stop counting immediately after `MemoryService.search()` returns.
7. Repeat with increasing values of N.

### Expected behavior

Vector completeness should preserve all existing integrity guarantees while issuing a number of application-level SQL statements that does not grow with the number of Entries.

The implementation must continue to verify:

- Active Entry heads and vector metadata represent the same Entry set.
- Entry IDs, entry version IDs, and content hashes match.
- Embedding content hashes match the active embedding profile.
- Every vector metadata row has a physical vector.
- The requested embedding profile matches the index profile.

Both existing completeness checks may remain because they enforce separate layer contracts.

The set-based query must expose both missing metadata for active heads and orphan or stale metadata that is not represented by an active head. A heads-driven `LEFT JOIN` alone is insufficient to prove set equality.

### Actual behavior

`vector_complete()` first loads N metadata rows, then issues one separate SQL statement for every row to verify that its physical vector exists.

This produces SQLAlchemy, aiosqlite, cursor, statement-execution, and async context-switch overhead for each Entry. Search results remain correct, but request overhead grows linearly with Memory size.

A local, uncommitted prototype replaced the per-row probes with a set-based query. Under its local measurement configuration, SQL statements changed as follows:

| Active Entries | Current implementation | Local prototype |
|---:|---:|---:|
| 100 | 231 | 29 |
| 1,000 | 2,031 | 29 |

These are preliminary local measurements, not CI results or a published benchmark contract. The acceptance criterion should be that SQL statement count remains constant as N increases, rather than requiring the specific value 29.

### Environment

 - PowerContext revision: `534460e068ef733ec22a672b838734635fa1677e`
    (local fork; branch: `master`; no release tag)
 - Python version: 3.12.3
 - OS: Ubuntu 24.04.4 LTS
 - Kernel/runtime: Linux 6.18.33.2-microsoft-standard-WSL2, x86_64

notepix:

Closes #1631 

# Rationale for this change

SQLite Vector search performed N+1 physical-vector probes during completeness validation. The validation runs once while selecting the search mode and again immediately before SQLite vector search, causing SQL statement count to grow with the number of active Memory Entries.

- Replace the N+1 physical-vector probes in `SQLiteMemoryVectorIndex.vector_complete()` with a set-based SQLite query.
- Add regression coverage for the SQL budget and existing vector-completeness validation behavior.

No API, configuration, or persisted-schema changes.
SQLite Vector search uses fewer database round trips for Memories with many active Entries.

# How was this change tested?

- `make check` — passed.
- `uv run python -m pytest tests/e2e/test_sqlite_vec.py -q` — 8 passed.
- `make unit-test` — known failures were reproduced on clean `6ffc5263` master: 49 failed, 135 passed, 1 skipped.
- `make e2e-test` — known failures were reproduced on clean `6ffc5263` master: 15 failed, 13 passed.

# AI usage statement

OpenAI Codex (GPT-5.6 sol) assisted with regression-test design, reviewed the generated changes and PR content.


src/powercontext/builtin/persistence/sqlite/memory_index.py
```diff
@@ -163,6 +163,33 @@
...
+     AND r.head_revision = h.head_revision
+    WHERE h.scope_id = :scope_id
+    UNION ALL
+    SELECT 'vector' AS row_kind, m.memory_artifact_id, m.head_revision,
+           m.entry_id, m.entry_version_id, m.entry_content_hash,
+           m.embedding_content_hash,
+           CASE WHEN v.rowid IS NULL THEN 0 ELSE 1 END AS vector_present
+    FROM pc_memory_vector_entries AS m
+    JOIN requested AS r
+      ON r.memory_artifact_id = m.memory_artifact_id
```
Teingi, line 196: [P2] Preserve stale-head retries when an append advances the revision

The metadata branch joins only on artifact ID, while the heads branch also filters by the requested revision. If an append commits after the backend validates the requested head but before the index query, the heads branch returns no rows while the metadata branch returns the new revision. `vector_complete()` then returns `False`, and the index raises `CapabilityNotSupportedError("vector")`, which `_is_stale_memory_search()` does not retry. A healthy index now causes concurrent vector/hybrid searches to fail.

Reproduced with the existing tests on real SQLite:
```sh
python -m pytest tests/e2e/test_memory_search_concurrency.py -k 'sqlite and (vector or hybrid)' -q
```
Both cases pass on base `6ffc5263` and fail on head `3fe8d7be` with the capability error.

Please distinguish an advanced head from an incomplete projection and preserve the existing stale-head retry behavior, while retaining the orphan/stale metadata checks.

notepix: Fixed in a86e178f.

The completeness query now reads the authoritative artifact head in the same SQL statement. When the requested revision has advanced concurrently, it raises the existing `head` capability signal so the runtime retries with the latest head. Stale or orphaned vector metadata still causes the vector completeness check to fail.

Validated with:
- python -m pytest tests/e2e/test_memory_search_concurrency.py -k 'sqlite and (vector or hybrid)' -q
- python -m pytest tests/e2e/test_sqlite_vec.py -q
- make check

Teingi (approved):
LGTM

PR #1653, opened 2026-09-18
feat(inference): add MiniMax native embedding adapter

### LINKED ISSUE #1642: feat: support MiniMax embedding backend (non-OpenAI-compatible /v1/embeddings)
## Feature description

Add first-class support for MiniMax as a PowerContext embedding backend.

Today PowerContext's embedding adapter only speaks the OpenAI embeddings contract. MiniMax exposes `/v1/embeddings` under an OpenAI-looking path but uses a different request/response shape, so it cannot be used as an "OpenAI-compatible" embedding endpoint. Users who already run generation on `openai-chat:MiniMax-M3` cannot complete their inference configuration because no compatible embedding provider is available.

### The incompatibility (reproduced)

PowerContext builds the embedding request through Pydantic AI's OpenAI client, which sends the standard OpenAI body:

```
POST /v1/embeddings
{ "model": "embo-01", "input": ["hello"], "dimensions": 1536 }
```

MiniMax rejects this and returns **HTTP 200** (not 4xx) with a vendor error envelope:

```json
{ "vectors": null, "base_resp": { "status_code": 2013, "status_msg": "invalid params, binding: expr_path=texts, cause=missing required parameter" } }
```

Two divergences from the OpenAI embeddings API:

1. **Request shape**: MiniMax requires `texts` + `type` (e.g. `{"texts":["hello"],"type":"db"}`), not `input`.
2. **Response shape**: MiniMax returns `{vectors: [...], base_resp: {...}}` instead of OpenAI's `{data: [{embedding: [...]}]}`. On error it returns `vectors: null` together with a `base_resp.status_code`.

Because the response never contains a `data` array, Pydantic AI raises during parsing and PowerContext surfaces a generic `InferenceError` (`InferenceUnavailableError` / `InvalidInferenceOutputError`). The failure is **loud** — no malformed vectors are stored in the index — but the message ("provider rejected" / "did not return a valid result") does not tell the operator that the endpoint is simply not OpenAI-compatible, which makes the root cause hard to diagnose.

Generation works correctly: `openai-chat:MiniMax-M3` against `https://api.minimaxi.com/v1` returns valid results, so the gap is specific to embeddings.

### Proposed solution

Add a MiniMax embedding adapter (or a small, provider-keyed request/response transform) so that when the embedding provider is MiniMax:

- the request body uses `texts` + `type` instead of `input`;
- the response is parsed from `vectors`, and a non-zero `base_resp.status_code` is mapped to a clear, named error.

Minimal alternative: at minimum, detect a non-OpenAI embeddings response (missing `data`, or a vendor error envelope such as `base_resp`) and raise a diagnostic `InferenceError` that explicitly states the endpoint is not OpenAI-compatible, so operators stop guessing.

## Alternatives considered

- **Use a different OpenAI-compatible embedding provider** (OpenAI `text-embedding-3-small`, DashScope `text-embedding-v4`, or any OpenAI gateway). This works today and is the current workaround, but it forces users who standardize on MiniMax to run a second vendor just for embeddings.
- **Generic custom-transform hook** for embeddings (user-supplied request/response mapping). More flexible but a larger surface; a MiniMax-specific adapter is a lower-risk first step.

## Additional context

- PowerContext version observed: `powercontext-v1.0.0-46-gee5c076e` (CLI self-report `1.0.1.dev20+ged11e5b65`).
- Evidence captured via `curl` against `https://api.minimaxi.com/v1/embeddings` (key omitted). Native MiniMax call `{"texts":["hello"],"type":"db"}` returns 1536-dim vectors, confirming the key has embedding permission and only the API shape differs.
- Generation config that works:
  - `POWERCONTEXT_SERVER_INFERENCE_GENERATION_MODEL=openai-chat:MiniMax-M3`
  - `POWERCONTEXT_SERVER_INFERENCE_GENERATION_BASE_URL=https://api.minimaxi.com/v1`
- This is distinct from `integrations/minimax/`, which is about MiniMax Code connecting to PowerContext as an MCP client. This issue is about PowerContext *consuming* MiniMax as an inference backend.

I am willing to contribute the code change.

AlexStocks:

Closes #1642.

# Rationale for this change

MiniMax exposes `/v1/embeddings` under an OpenAI-looking path but speaks a different dialect, so it cannot be used as an "OpenAI-compatible" embedding endpoint. PowerContext's only embedding path goes through Pydantic AI's OpenAI client, which sends `{"model","input","dimensions"}` and parses OpenAI's `{data:[{embedding}]}`. MiniMax requires `{"model","texts","type"}` and answers `{"vectors","total_tokens","base_resp"}` — returning **HTTP 200** with a non-zero `base_resp.status_code` on business errors. The OpenAI client therefore fails while parsing, and the operator only sees a generic "provider rejected" message that does not point at the real cause.

The result was that users who already run generation on `openai-chat:MiniMax-M3` could not complete their inference configuration. This PR adds the missing adapter, which is the solution proposed in the issue.

**New adapter — `src/powercontext/builtin/inference/minimax.py`**

`MiniMaxEmbeddingModel` is a self-contained `EmbeddingModel` implementation that:

- issues the native request body (`texts` + `type: "db"`, with the provider prefix stripped from `model`);
- parses `vectors` and validates the returned vector count against the number of inputs;
- maps HTTP failures to `InferenceUnavailableError`, and a non-zero `base_resp.status_code` (including the `2013` envelope from the issue) to `InferenceUnavailableError` / `InvalidInferenceOutputError` — so a null or partial vector is never handed to the index;
- owns an `httpx.AsyncClient` registered on the runtime exit stack, so it is closed with the rest of the resources;
- exposes `is_minimax_embedding()` for host/prefix detection.

**Routing — `src/powercontext/builtin/runtime/composition.py`**

`_embedding_models()` detects a MiniMax endpoint — `api.minimaxi.com` / `api.minimax.io` host, or an explicit `minimax` model prefix — and routes to the new adapter via a small `_minimax_embedding_models()` helper that mirrors the existing provider path (operational + readiness instances, `EmbeddingProfile`, batch size, timeout). The existing OpenAI/Anthropic path is untouched, and no new configuration knob is introduced. The existing base-URL guard in the helper was widened to also require the embedding model, which narrows the type for the profile construction.

**Tests**

- `tests/builtin/inference/test_minimax_embedding.py` (new) — request body shape, response parsing, the `base_resp` business-error envelope, HTTP errors, and vector-count validation.
- `tests/builtin/runtime/test_composition_embedding.py` — routing test asserting that a MiniMax host selects `MiniMaxEmbeddingModel` and that the request actually reaching the transport is the native shape.

No change to the OpenAI embedding path, to any persisted format, or to the `EmbeddingProfile` contract.

Yes, one — MiniMax can now be configured as an embedding backend using the existing generic variables:

```ini
POWERCONTEXT_SERVER_INFERENCE_EMBEDDING_MODEL=openai:embo-01
POWERCONTEXT_SERVER_INFERENCE_EMBEDDING_BASE_URL=https://api.minimaxi.com/v1
POWERCONTEXT_SERVER_INFERENCE_EMBEDDING_HEADERS='{"Authorization":"Bearer <key>"}'
POWERCONTEXT_SERVER_INFERENCE_EMBEDDING_PROFILE_ID=minimax-embo-01-v1
POWERCONTEXT_SERVER_INFERENCE_EMBEDDING_DIMENSION=1536
```

Notes for reviewers:

- The provider prefix stays `openai:` — selection is by base-URL host / model prefix, so no new provider name is needed and existing configs keep working.
- `MODEL` + `PROFILE_ID` + `DIMENSION` remain a three-way contract, exactly as for every other embedding provider.
- No breaking change to public APIs or persisted formats.
- This does not touch `docs/`. A note listing MiniMax among supported embedding backends may be worth a follow-up.

# How was this change tested?

```bash
# Tests (19)
.venv/Scripts/python.exe -m pytest \
  tests/builtin/inference/test_minimax_embedding.py \
  tests/builtin/runtime/test_composition_embedding.py -q
# 19 passed

# Lint / format / types
.venv/Scripts/ruff.exe check        <the 4 changed files>   # All checks passed!
.venv/Scripts/ruff.exe format --check <the 4 changed files> # 4 files already formatted
.venv/Scripts/ty.exe check --python-version 3.11 <the 4 changed files>  # All checks passed!
```

Beyond the unit tests, the change was validated end to end against the live MiniMax endpoint, on a personal server built from this branch:

1. started the server with the configuration shown above — `/health/ready` reports `inference.embedding: ready`;
2. wrote a Memory entry, then issued a `POST /v1/memory/search` with `mode: "vector"` using a query that shares **no** meaningful literal overlap with the stored text;
3. the entry came back with `"matched_by": ["vector"]`.

Because retrieval only matched through the vector channel, this confirms that both the write-side and the query-side embeddings were really produced by MiniMax (1536 dimensions) rather than by any keyword path. The failing behaviour described in #1642 no longer reproduces.

# AI usage statement

AI assistance: code. The adapter, the runtime routing, and the tests were drafted with an AI coding agent (WorkBuddy). The full diff was reviewed, and every command listed above was executed locally against the commit in this PR before submission.


src/powercontext/builtin/inference/minimax.py
```diff
@@ -0,0 +1,196 @@
...
+        self.profile = profile
+        self._model = model
+        self._batch_size = batch_size
+        self._timeout = timeout_seconds
+        self._endpoint = f"{base_url.rstrip('/')}/embeddings"
+        request_headers: dict[str, str] = {"Content-Type": "application/json"}
+        if headers:
+            request_headers.update(headers)
+        self._headers = request_headers
+        self._client = http_client or httpx.AsyncClient()
```
Teingi: [P2] Honor the configured embedding timeout in the HTTP client

This client retains HTTPX's default 5-second timeout, so the outer `asyncio.wait_for` cannot honor a longer `embedding_timeout_seconds`. With the setting at 30 seconds and a local provider returning valid vectors after 6 seconds, the adapter raised `InferenceUnavailableError` caused by `ReadTimeout` after 5.01 seconds. The same request succeeded in 6.03 seconds with the client timeout set to 30 seconds. Slow but valid requests can therefore leave Memory writes without vectors or trigger retrieval fallback. Please align the transport timeout with the configured deadline (or let the outer deadline control it), and map transport timeouts to `InferenceTimeoutError`.

src/powercontext/builtin/inference/minimax.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+def is_minimax_embedding(base_url: str | None, model: str | None) -> bool:
+    """Detect a MiniMax embedding endpoint by host or explicit model prefix."""
+
+    host = (urlparse(base_url or "").hostname or "").lower()
+    if any(host == domain or host.endswith(f".{domain}") for domain in ("minimaxi.com", "minimax.io")):
+        return True
+    provider_prefix, separator, _model_name = (model or "").partition(":")
+    return bool(separator) and provider_prefix.lower() == "minimax"
```
Teingi, line 195: [P2] Accept the MiniMax prefix in Topic Memory worker validation too

The adapter accepts `minimax:embo-01`, but `validate_topic_memory_provider_settings()` still rejects the `minimax` prefix. Consequently, `processing_capabilities()` drops `topic-memory`; with `topic_memory_schedule_seconds=300`, runtime startup fails with `Topic Memory processing requires a configured generation model` even when generation is configured. I reproduced this with file-backed SQLite and `generation_model='test'`: the same setup starts successfully with `openai:embo-01`. Please update capability detection and worker validation together with this new prefix, and cover full runtime startup so the supported embedding configuration does not disable Topic Memory or prevent the server from starting.

Teingi (commented):
Two P2 issues remain, both reproduced locally at this revision. The focused test suite passed (106 tests); the additional probes used a delayed localhost HTTP server and disposable SQLite databases, without calling the live MiniMax service.

PsiACE (approved):
lgtm, thanks

PR #1660, opened 2026-09-19
fix(workbuddy): extend prompt hook timeout

### LINKED ISSUE #1659: bug: WorkBuddy prompt hook can be killed before fail-open completes
## Describe the bug

The WorkBuddy integration registers the PowerContext `UserPromptSubmit` hook with an outer host timeout of 10 seconds. In real WorkBuddy sessions, a cold or degraded local PowerContext Server can make the hook exceed that host timeout even though the hook is designed to fail open internally.

When this happens, WorkBuddy blocks the submitted prompt with a hook timeout instead of receiving the hook's normal fail-open JSON response.

## Steps to reproduce

1. Install the WorkBuddy integration with `powercontext setup workbuddy`.
2. Use a PowerContext Server that is cold-starting, degraded, or slow to resolve scope/prepare context.
3. Submit a prompt in WorkBuddy while the hook is still resolving scope or preparing context.
4. Observe that WorkBuddy terminates the hook at the registered 10-second timeout.

## Expected behavior

The WorkBuddy prompt should continue. If PowerContext is unavailable or slow, the hook should fail open and return an empty `additionalContext` response, matching the documented integration behavior.

## Actual behavior

The host can terminate the hook before the fail-open path returns, producing a `UserPromptSubmit operation blocked by hook` timeout. In one observed run, the hook returned fail-open output only after the 10-second host limit; later warm runs completed faster.

The hook already uses a much smaller internal HTTP wall-clock budget, so increasing the outer host timeout gives process startup, local server cold-start jitter, and scope-resolution overhead enough room without making individual PowerContext HTTP calls unbounded.

## Environment

- Host: WorkBuddy desktop
- OS: Windows
- Integration: PowerContext WorkBuddy hook installed through `powercontext setup workbuddy`
- Registered hook timeout: 10 seconds

Yes, I would like to submit a PR.

AlexStocks:

Closes #1659.

# Rationale for this change

The WorkBuddy integration currently registers the PowerContext `UserPromptSubmit` hook with a 10-second host timeout. In real desktop sessions, cold startup, degraded local services, or scope resolution overhead can push the hook past that outer timeout even though the hook itself has a smaller HTTP budget and is designed to fail open.

A longer outer timeout gives the hook process enough room to return its fail-open response instead of letting WorkBuddy kill the hook and block the prompt.

- Raises the WorkBuddy `UserPromptSubmit` hook timeout from 10 seconds to 30 seconds in the CLI installer.
- Updates the WorkBuddy hook template to match the installer.
- Updates WorkBuddy integration docs in English, Chinese, and the integration README.
- Updates WorkBuddy CLI tests so installed and refreshed hooks assert the new timeout.

Yes. New or refreshed WorkBuddy installations now register the PowerContext prompt hook with a 30-second outer timeout. The hook's internal HTTP request timeout and shared HTTP budget remain unchanged, so PowerContext HTTP calls still fail open under their bounded budget.

There are no public API, persisted format, or MCP contract changes.

# How was this change tested?

- `rg -n 'WORKBUDDY_HOOK_TIMEOUT = 10|"timeout": 10|Hook timeout \(10 s\)|Hook 超时（10 秒）|hook\["timeout"\] == 10' src/powercontext/cli/workbuddy.py integrations/workbuddy docs/en/docs/integrations/workbuddy.md docs/zh/docs/integrations/workbuddy.md tests/test_cli_workbuddy.py`
- `uv run pytest tests/test_cli_workbuddy.py tests/test_integration_manifest.py -q`
- `git diff --check`

# AI usage statement

Used OpenAI Codex to investigate the WorkBuddy hook timeout, prepare the patch, update tests/docs, and run the validation commands above.


Teingi (approved):
LGTM

PR #1670, opened 2026-09-19
fix(mcp): allow Markdown handoff report output

### LINKED ISSUE #1664: bug(mcp): Markdown handoff reports fail output-schema validation
## Describe the bug

The MCP `get_handoff_report` tool accepts `format=markdown`, but the result fails output-schema validation even when the same report succeeds over HTTP. Agents cannot use the advertised Markdown variant.

## Steps to reproduce

1. Start a Server and create a Scope accessible to the caller. A Scope without a current handoff is sufficient.
2. Through a real MCP client connected to `/mcp`, call `get_handoff_report` with:

```json
{"selection":{"mode":"exact","scope_ids":["<scope_id>"]},"format":"markdown"}
```

3. Repeat with `format=json`.
4. POST the same two payloads to `/v1/handoff-reports/get` over HTTP.

## Expected behavior

The Markdown tool result is valid against the advertised MCP output contract and contains the report, as the JSON variant does.

## Actual behavior

| Transport / format | SQLite | OceanBase | seekdb |
|---|---|---|---|
| MCP / Markdown | `is_error=true` | `is_error=true` | `is_error=true` |
| MCP / JSON | Success, structured content | Success, structured content | Success, structured content |
| HTTP / Markdown | 200, `text/markdown` | 200, `text/markdown` | 200, `text/markdown` |
| HTTP / JSON | 200 | 200 | 200 |

The MCP error is identical across all three backends:

```text
Output validation error: outputSchema defined but no structured output returned
```

The Markdown result has no structured content. HTTP report generation and Scope access both work in the controls.

Inspect the OpenAPI-to-MCP projection in [src/powercontext/server/mcp.py:224](https://github.com/oceanbase/powercontext/blob/fa581eeca400bf3116e1191fcf4add0dd702f5cf/src/powercontext/server/mcp.py#L224): the text response and the advertised structured output schema disagree. Preserve HTTP Markdown behavior while giving MCP a conforming output. Regress with a real MCP client for both JSON and Markdown.

## Environment

- FastMCP 3.4.4; real Streamable HTTP MCP client calls.

Teingi:

Closes #1664.

# Rationale for this change

`get_handoff_report(format="markdown")` returned text while advertising an object output schema, so MCP rejected a successful report with `outputSchema defined but no structured output returned`.

Omit the optional object output schema for this mixed-format tool. JSON continues to return structured content and Markdown returns text. Other tools retain their output schemas and the report retains its read-only annotations. Add a Streamable HTTP MCP regression covering both formats alongside HTTP controls.

Agents can read Markdown handoff reports over MCP. HTTP formats and request contracts are unchanged.

# How was this change tested?

- Reproduced the Markdown validation failure before the fix; the JSON control passed.
- `python -m pytest tests/test_mcp.py tests/e2e/test_mcp_transport.py -q`: 14 passed. Re-ran the two format regressions after the final test adjustment: 2 passed.
- Real loopback HTTP and Streamable HTTP MCP clients against SQLite, OceanBase from the configured environment, and embedded seekdb, with authentication and enforced Access: both formats passed on all three backends. Dedicated databases/directories were removed and cleanup verified. No LLM is involved in this transport path.
- Ruff and focused `ty check` passed. `ty check --exclude tests/test_integration_guidance_evaluation.py` passed.
- `make check`: lock validation and all formatting/lint hooks passed; full type checking reports four existing unresolved `scripts.*` imports in `tests/test_integration_guidance_evaluation.py`, also present on the unchanged base checkout.

# AI usage statement


Zxf-xufeng (approved):
LGTM

happy-v587:
LGTM

PR #1671, opened 2026-09-19
fix(context): bound optional Topic recall embedding

### LINKED ISSUE #1665: bug(context): slow embedding failure prevents available FTS memory from reaching Codex hooks
## Describe the bug

When Embedding is configured but fails slowly, the default `prepare_context` assembly waits for it even though useful Memory is already available through FTS. The Server eventually returns the correct context, but the Codex Hook times out first and injects nothing.

## Steps to reproduce

1. On an isolated SQLite Server, configure an OpenAI-compatible Embedding endpoint with matching model/profile/dimension settings. Make the endpoint accept requests, wait 20 seconds, then return HTTP 503. Set `POWERCONTEXT_SERVER_INFERENCE_EMBEDDING_TIMEOUT_SECONDS=12`. This is an explicit fault-injection setup, not a production latency measurement.
2. Seed a fresh Scope through `POST /v1/memory/remember` with `kind=fact` and this text. During seeding, let the endpoint return 503 immediately; enable the 20-second delay before the read comparisons.

```text
For the synthetic ORCHID project the release codename is ORCHID-728 and the required validation command is pytest -q.
```

3. Use the same Scope and query in two rounds:

```text
What are the ORCHID release codename and required validation command?
```

- POST `/v1/memory/search` with `scope_id`, `query`, and `mode=fts`.
- POST `/v1/context/prepare` with `scope_id`, `query`, `max_bytes=4000`, and `assembly={"sections":[{"family":"memory","limit":3}]}`.
- Repeat prepare with the `assembly` field omitted (default assembly).

4. Point the current Codex plugin's effective `.mcp.json` URL and authorization at this Server, use the same Scope, and invoke its actual `UserPromptSubmit` Hook with the query. Compare default budgets with `POWERCONTEXT_CODEX_REQUEST_TIMEOUT_SECONDS=3` and `POWERCONTEXT_CODEX_HTTP_BUDGET_SECONDS=8`.

## Expected behavior

A failed optional Embedding dependency should not prevent already available FTS Memory from being delivered within the online recall budget. The Hook should remain fail-open, but useful context should be returned promptly when it is available.

## Actual behavior

| Request | Round 1 | Round 2 | Result |
|---|---:|---:|---|
| Direct FTS search | 0.129 s | 0.133 s | Correct fact |
| Memory-only prepare | 0.208 s | 0.201 s | `ready`, correct fact |
| Default prepare | 12.185 s | 12.232 s | Eventually `ready`, correct fact |

The actual Hook subprocess has no `additionalContext` and reports `server_unavailable` under both default budgets (1-second request / 4-second HTTP budget) and the increased budgets (3 / 8 seconds). Hook process durations are 1.872 and 3.382 seconds, including startup. A short-query Hook control against an FTS-only Server successfully injects two facts.

The Topic Memory search path awaits Embedding before falling back: [src/powercontext/builtin/runtime/application.py:2615](https://github.com/oceanbase/powercontext/blob/fa581eeca400bf3116e1191fcf4add0dd702f5cf/src/powercontext/builtin/runtime/application.py#L2615). Bound dependency fallback by the online recall budget, skip unnecessary embedding work where appropriate, or return available Memory promptly. A regression should assert actual Hook injection, not only eventual HTTP 200.

This is separate from #1594: it occurs during context preparation with a slow dependency and an already selected Scope, rather than WorkBuddy's cold-start Scope resolution.

## Environment

- SQLite and an actual Codex Hook subprocess; disposable databases and synthetic data; real HTTP services with authentication and enforced Access.
- No new native Codex model conversation was run for this replay. The controlled 20-second fault / 12-second adapter timeout does not describe the current health of any external gateway.
- Recall gate remains at its default setting; enabling it was not tested as a mitigation for this issue.

Teingi:

Closes #1665.

# Rationale for this change

Default context preparation could wait for Topic query Embedding long enough for the Codex Hook to time out, even after FTS Memory was available.

Skip query Embedding when a Scope has no current Topics. Give Topic query Embedding during context preparation a 250 ms budget, cancel it on expiry, and use FTS. Cache the fallback within that prepare request so admission and policy-floor expansion rounds do not repeat the timed-out attempt. Successful vectors remain reusable, and the next prepare request can attempt Embedding again. Direct Topic search retains the provider timeout. Add cancellation, gate-expansion, and actual Codex Hook regressions with empty/populated Topics, plus English/Chinese documentation.

Slow optional Topic Embedding no longer delays available FTS context until the provider timeout. Fast hybrid retrieval remains available; a Topic Embedding call exceeding the online budget uses FTS. No HTTP contract changes.

# How was this change tested?

- `python -m pytest tests/e2e/test_recall_sufficiency_gate.py tests/builtin/runtime/test_topic_memory_application.py tests/e2e/test_codex_service_chain.py -q`: 45 passed. The expansion regression forces all three rounds and checks one external Embedding attempt per prepare across two requests.
- Real authenticated SQLite HTTP with enforced Access, the recall gate enabled, and an OpenAI-compatible fault endpoint delayed by 20 seconds before returning 503. Two rounds with empty/populated Topic collections cover direct FTS, prepare, and actual Codex Hook subprocesses at default 1/4-second and increased 3/8-second budgets. All eight Hook calls injected the expected facts. This is controlled fault injection, not a remote provider latency measurement.
- Dedicated SQLite data and the HTTP fault server were cleaned up.
- Ruff and `ty check --exclude tests/test_integration_guidance_evaluation.py` pass. Full `prek run -a` passes formatting/lint hooks and encounters four pre-existing unresolved `scripts.*` imports in that test file.

# AI usage statement


Zxf-xufeng:
Consider avoiding repeated embedding attempts after a timeout within the same prepare request. With the recall gate enabled, the initial retrieval plus up to two expansion rounds can each wait 250 ms, accumulating approximately 750 ms of embedding wait time. Database retrieval and context assembly add further latency, potentially exceeding the PowerContext Codex plugin’s configurable 1-second default per-request HTTP timeout. This applies when both expansion rounds are triggered and all three embedding attempts time out; the recall gate is disabled by default.

happy-v587 (commented):
LGTM

Teingi:
@Zxf-xufeng Addressed in ee2da512. Prepare now caches the FTS fallback for the rest of that request, so the recall gate cannot repeat the timed-out Topic embedding attempt. Successful vectors are still reused, and a new prepare request can attempt embedding again.

The regression forces both expansion rounds and verifies one embedding attempt per prepare, including a second request. All 45 focused tests passed. With the gate enabled, a real HTTP embedding endpoint delayed by 20 seconds also passed the empty/populated Topic checks and all eight Codex Hook calls, including the default 1-second request / 4-second total budgets. Test resources were cleaned up.

Zxf-xufeng (approved):
LGTM

PR #1672, opened 2026-09-19
fix(artifacts): preserve registered Source provenance

### LINKED ISSUE #1666: bug(artifacts): registered custom sources make generated Profile list and detail return 500
## Describe the bug

A dynamically registered `note` Source can be accepted, used in real Profile generation, and approved. Reading the resulting Profile through the generic Artifact list/detail endpoints then returns HTTP 500 because its Source reference is mapped through a closed enum.

## Steps to reproduce

Generate the registered remote-source input from the repository's existing `NoteSource` fixture (run from the checkout with test dependencies installed):

```python
from powercontext.sources import SourceDefinitionRegistry, SourceMaterialization
from powercontext.sources.observations import manifest_for_definition, project_source_for_transport
from tests.builtin.persistence.contract import SOURCE_ADAPTERS, NoteSource

registry = SourceDefinitionRegistry.from_adapters(SOURCE_ADAPTERS)
manifest = manifest_for_definition(registry.definition_for_name("note"))
observation = project_source_for_transport(
    registry,
    NoteSource(
        name="retest-note",
        materialization=SourceMaterialization.CAPTURED,
        body="I am a Python backend engineer. I always prefer concise Chinese engineering reports and use pytest. Always validate configuration before deployment.",
    ),
)
register_body = {"manifest": manifest.model_dump(mode="json", by_alias=True)}
observation_body = {"scope_id": scope_id, "observation": observation.model_dump(mode="json", by_alias=True)}
```

`scope_id` is a newly created, writable Scope. POST `register_body` to `/v1/source-definitions/register`, then POST `observation_body` to `/v1/source-observations`. Both succeed. The adapter above is used by the client to produce the manifest/observation; do not install a custom local adapter into the Server.

Also POST this built-in Content to `/v1/sources/content` in the same Scope:

```json
{
  "scope_id": "<scope_id>",
  "source_id": "builtin-control",
  "content": "Synthetic release configuration bypassed validation and the deployment failed. We added strict schema validation before deployment and ran pytest. All 12 checks passed and the retry succeeded. Stable engineering preference: always validate configuration and run pytest before production release; use concise Chinese reports."
}
```

1. Read `/v1/scopes/<scope_id>/profile-policy`, then PUT it with `generation_enabled=true`, `activation_mode=review_required`, and the current `expected_version` (0 for the tested fresh Scope).
2. POST `/v1/scopes/<scope_id>/subject-sources` with `{"subject_key":"synthetic-user","content":"I prefer concise Chinese engineering reports and use pytest."}`.
3. POST `/v1/profile/flush` with `{"scope_id":"<scope_id>"}` using a working generation model. It returns `review_pending` and a `candidate_id`.
4. Read the candidate via `/v1/artifact-candidates/get`; confirm `source_refs` includes `{"name":"note",...}`. Approve it via `/v1/artifact-candidates/approve` with `scope_id`, `candidate_id`, and its `expected_version`.
5. GET `/v1/scopes/<scope_id>/artifacts/profile` twice, then GET `/v1/scopes/<scope_id>/artifacts/profile/profile`.

## Expected behavior

An approved Profile containing a valid registered Source reference can be listed and read without losing its provenance.

## Actual behavior

On SQLite, OceanBase, and seekdb, real Profile generation and approval succeed (HTTP 200; candidate becomes `approved` with artifact `profile/profile`, revision 1). Each backend then returns **500 / 500 / 500** for the repeated list and detail calls. A readable empty Scope on the same Server returns 200 for its Profile list.

```text
ValueError: 'note' is not a valid SourceType
```

The HTTP body is `{"error":{"code":"internal_error","message":"The Server failed.","details":null}}`. Repeated reads use fresh HTTP connections.

[src/powercontext/server/mapping.py:1107](https://github.com/oceanbase/powercontext/blob/fa581eeca400bf3116e1191fcf4add0dd702f5cf/src/powercontext/server/mapping.py#L1107) still calls:

```python
SourceTypeReference(source_type=SourceType(value.source_type), source_id=value.source_id)
```

This rejects dynamically registered Source names. Align generic Artifact provenance with the open Source-name contract, updating generated types through the canonical OpenAPI contract if needed. Cover registration → Observation → generation → approval → list/detail, preserving Source references.

## Environment

- Profile generation used a real `qwen3.7-plus` model; no mocked Profile result.

Teingi:

Closes #1666.

# Rationale for this change

Profiles generated from accepted remote Source observations could be approved but then fail generic Artifact list/detail reads because provenance was converted through the closed `SourceType` enum.

Make `SourceTypeReference.source_type` an open string and allow Unicode/interior spaces in `source_id`, retaining its 1–256 character bounds. Regenerate Python models/schema from the canonical OpenAPI contract and preserve the Source name directly in response mapping. The Profile review/CRUD regression registers a remote `note` definition, submits `preferences`, `用户偏好`, and `release notes` without a Server adapter, approves the Profile, and verifies exact provenance in list, detail, and revision responses. Update SDK usage and bilingual documentation.

Generic Artifact responses preserve registered Source names such as `note` and accepted Unicode/interior-space Source identities. The JSON shape stays the same. Python Client `SourceTypeReference.source_type` is now a string, so callers should use it directly instead of `.value`. The built-in Content creation contract is unchanged.

# How was this change tested?

- Reproduced `ValueError: 'note' is not a valid SourceType` before the fix, both with and without enforced Access.
- `make api-generate` and `make contract-test`: generated artifacts current; 48 tests passed.
- `python -m pytest tests/e2e/test_profile_api.py tests/test_dashboard_profile_handoff.py -q`: 23 passed.
- `python -m pytest tests/e2e/test_base_access_api.py -q`: 4 passed.
- Real `qwen3.7-plus` generation from the configured environment on SQLite, OceanBase, and embedded seekdb: register definition → submit all three remote Source IDs plus built-in Content/subject evidence → generate Profile → inspect Candidate → approve → repeated list, detail, and revision reads. Repeated list reads, detail, and exact revision all returned 200 on each backend and preserved every source reference, including `用户偏好` and `release notes`. Dedicated databases/directories were removed and cleanup verified.
- Ruff and `ty check --exclude tests/test_integration_guidance_evaluation.py` passed. Full pre-commit hooks encounter only the same four pre-existing unresolved `scripts.*` imports in that test file.

# AI usage statement


Zxf-xufeng:
[P2] Align source_id validation with the remote Source contract
Opening source_type fixes serialization for registered types such as note, but SourceTypeReference.source_id still enforces ^[\x21-\x7E]+$. Remote observations and internal SourceRef values accept IDs containing Unicode or internal spaces.
I verified that observations with IDs such as 用户偏好 and release notes pass both transport and runtime observation validation, but source_type_reference() raises a Pydantic string_pattern_mismatch error. Artifacts carrying those references would therefore still fail serialization in list, detail, and revision responses. The new regression uses the ASCII ID preferences, so it does not cover this case.
Please align the OpenAPI source_id constraints with the accepted remote Source identity contract, regenerate the models, and add an end-to-end regression covering a Unicode or space-containing ID through submission, approval, and Artifact reads.
This is an existing contract mismatch that remains unresolved by this fix, rather than a newly introduced regression.

happy-v587 (commented):
LGTM

happy-v587 (commented):
Correction to my earlier LGTM: the existing [P2] is confirmed independently. `SourceRef` accepts Unicode and interior-space IDs, but the generated `SourceTypeReference.source_id` still enforces `^[\x21-\x7E]+$`. On the current head, `source_type_reference(SourceRef(source_type="note", source_id="用户偏好"))` and the same call with `"release notes"` both fail with `string_pattern_mismatch`.

A registered remote Source using either valid identity can therefore still make generic Artifact response mapping fail. Please align the OpenAPI `source_id` constraint with the accepted SourceRef contract, regenerate, and cover submission through list/detail/revision reads.

Teingi:
@Zxf-xufeng @happy-v587 Confirmed and fixed in 65e6cf04. `SourceTypeReference.source_id` now accepts the same Unicode and interior-space identities as remote Observation ingestion, retaining the 1–256 character bounds. The OpenAPI contract and generated models/schema are updated.

The regression submits `preferences`, `用户偏好`, and `release notes`, then approves the Profile and checks exact provenance in list, detail, and revision responses with Access both disabled and enforced. All 48 contract tests and 27 related API tests passed. Real qwen3.7-plus generation/approval on SQLite, OceanBase, and seekdb also passed repeated list/detail/revision reads with both reported IDs preserved. The isolated databases and directories were cleaned up.

Zxf-xufeng (approved):
LGTM

PR #1673, opened 2026-09-19
fix(topic-memory): process persisted remote source evidence

### LINKED ISSUE #1667: bug(topic-memory): registered remote Source observations block processing after restart
## Describe the bug

A Scope containing a registered remote `note` Observation and built-in Content is accepted by the public APIs, but Topic Memory processing repeatedly fails with `SourceAdapterNotFoundError`. Restarting the Server with the same database does not recover it. Built-in Content in the mixed window cannot produce Topics either.

## Steps to reproduce

Generate the registered remote-source input from the repository's existing `NoteSource` fixture (run from the checkout with test dependencies installed):

```python
from powercontext.sources import SourceDefinitionRegistry, SourceMaterialization
from powercontext.sources.observations import manifest_for_definition, project_source_for_transport
from tests.builtin.persistence.contract import SOURCE_ADAPTERS, NoteSource

registry = SourceDefinitionRegistry.from_adapters(SOURCE_ADAPTERS)
manifest = manifest_for_definition(registry.definition_for_name("note"))
observation = project_source_for_transport(
    registry,
    NoteSource(
        name="retest-note",
        materialization=SourceMaterialization.CAPTURED,
        body="I am a Python backend engineer. I always prefer concise Chinese engineering reports and use pytest. Always validate configuration before deployment.",
    ),
)
register_body = {"manifest": manifest.model_dump(mode="json", by_alias=True)}
observation_body = {"scope_id": scope_id, "observation": observation.model_dump(mode="json", by_alias=True)}
```

`scope_id` is a newly created, writable Scope. POST `register_body` to `/v1/source-definitions/register`, then POST `observation_body` to `/v1/source-observations`. Both succeed. The adapter above is used by the client to produce the manifest/observation; do not install a custom local adapter into the Server.

Also POST this built-in Content to `/v1/sources/content` in the same Scope:

```json
{
  "scope_id": "<scope_id>",
  "source_id": "builtin-control",
  "content": "Synthetic release configuration bypassed validation and the deployment failed. We added strict schema validation before deployment and ran pytest. All 12 checks passed and the retry succeeded. Stable engineering preference: always validate configuration and run pytest before production release; use concise Chinese reports."
}
```

1. Create a second control Scope containing only the same built-in Content.
2. With a working generation model, POST `/v1/topic-memory/flush` for both Scopes and observe worker completion/errors; HTTP `accepted` alone is not a completion signal.
3. Repeat flush for the mixed-source Scope.
4. Restart the Server against the same database, flush the mixed-source Scope again, and check new worker failures plus `/v1/scopes/<scope_id>/artifacts/topic-memory` for both Scopes.

## Expected behavior

Persisted remote Observation content should be consumable through the supported Source projection path, including after restart. Accepted Source input should not leave the entire mixed-source Topic window in repeated adapter lookup failures.

## Actual behavior

Flush returns 200 `accepted`, but the mixed-source Scope has zero Topics and the worker records `SourceAdapterNotFoundError`. The failure happens before generation, during source-window projection.

| Backend | New failures observed after restart | Mixed-source Topics | Content-only control Topics |
|---|---:|---:|---:|
| SQLite | 2 | 0 | 2 |
| OceanBase | 1 | 0 | 1 |
| seekdb | 1 | 0 | 2 |

These failure counts are from the observation window, not a claim about total retries. The control Topics were generated by the real model and remained readable after restart.

Relevant stack:

```text
topic_memory_processing.py:1237, _project_evidence
    materialized = await sources.read_value(stored.value)
builtin/persistence/sources.py:345, read_value
    return await self._registry.read(source)
sources/definitions.py:155, definition_for_source
    raise SourceAdapterNotFoundError("source", type(source)) from None
```

Inspect [src/powercontext/builtin/runtime/topic_memory_processing.py:1237](https://github.com/oceanbase/powercontext/blob/fa581eeca400bf3116e1191fcf4add0dd702f5cf/src/powercontext/builtin/runtime/topic_memory_processing.py#L1237) and [src/powercontext/sources/definitions.py:155](https://github.com/oceanbase/powercontext/blob/fa581eeca400bf3116e1191fcf4add0dd702f5cf/src/powercontext/sources/definitions.py#L155). The worker resolves a local adapter for a remotely submitted Observation rather than successfully consuming its persisted content/projections. Regress mixed built-in/remote windows, generated output, provenance, cursor advancement, and restart recovery.

## Environment

- Real `qwen3.7-plus` generation succeeds for the built-in Content control. Mixed-source failures occur before model generation.

Teingi:

Closes #1667.

# Rationale for this change

An accepted remote Source Observation caused Topic window selection and processing to fail with `SourceAdapterNotFoundError`. The worker tried to resolve the Observation through the Server's local adapter registry, blocking built-in Content in the same window as well.

Build Topic evidence directly from persisted Observations: prefer the standard text-evidence projection, otherwise use the captured payload without envelope identity fields. Keep the existing canonicalization and evidence limits, preserve original Source references, and leave native Source materialization unchanged. Add restart regressions for both remote payload forms in a mixed window, checking published lineage and cursor advancement, and document the behavior in English and Chinese.

Registered remote Sources can participate in Topic processing without installing their adapters on the Server. Topic revisions retain their remote and built-in provenance. Generic Artifact response serialization for open Source names is addressed independently in #1672.

# How was this change tested?

- Both new remote-source regressions reproduced `SourceAdapterNotFoundError` before the fix.
- `python -m pytest tests/e2e/test_topic_memory_processing.py tests/builtin/runtime/test_topic_memory_processing.py tests/builtin/runtime/test_topic_memory_security.py -q`: 46 passed.
- Real configured `qwen3.7-plus` with authenticated HTTP and enforced Access on SQLite, OceanBase, and embedded seekdb: register remote Note, submit Note plus built-in Content, restart, flush, wait for publication, retrieve exact Topic revisions, verify both Source names and completed cursors, restart and read/flush again. A separate built-in-only Scope is the control. Dedicated resources were removed after validation.
- Full pre-commit hooks pass apart from the four baseline unresolved `scripts.*` imports in `tests/test_integration_guidance_evaluation.py`. `ty check --exclude tests/test_integration_guidance_evaluation.py` passes.

The OceanBase live run recorded one model `invalid_reconciliation` failure and then recovered through the normal worker retry. All three backends completed both cursors, preserved provenance, and passed the restart checks; no `SourceAdapterNotFoundError` remained.

# AI usage statement


happy-v587 (commented):
LGTM

Zxf-xufeng (approved):
LGTM

PR #1674, opened 2026-09-19
fix(access): distinguish missing Memory from pending ownership

### LINKED ISSUE #1668: bug(access): missing Memory in a readable Scope returns artifact_owner_pending 503
## Describe the bug

An exact Memory lookup in the wrong but readable Scope returns a temporary Access Control outage (`503 artifact_owner_pending`), even though that Scope is valid, accessible, and simply contains no such Memory. Clients are encouraged to retry an invalid reference instead of correcting it.

## Steps to reproduce

1. With enforced Access and the same authenticated principal, create writable/readable Scopes A and B. Leave B empty.
2. POST `/v1/memory/remember` in A with `kind=fact` and arbitrary synthetic text.
3. POST `/v1/memory/entries/list` with A's `scope_id` and copy an entry's exact `citation`.
4. POST `/v1/memory/entries/get` with B's `scope_id` and that citation:

```json
{
  "scope_id": "<B>",
  "citation": {
    "memory_ref": {"family":"memory","artifact_id":"memory","revision":1},
    "entry_id": "<entry from A>",
    "entry_version_id": "<version from A>"
  }
}
```

Use the actual revision and IDs returned in step 3. Confirm the principal can read Scope B, and repeat the request.

## Expected behavior

Return a stable missing-resource/invalid-reference result within the authorized Scope. Preserve non-disclosure of inaccessible resources. Reserve temporary owner-pending/unavailable semantics for an existing resource whose ownership is actually pending.

## Actual behavior

SQLite, OceanBase, and seekdb each return the following on both attempts:

```http
HTTP/1.1 503 Service Unavailable
```

```json
{"error":{"code":"artifact_owner_pending","message":"Access Control is unavailable.","details":null}}
```

Other reads in the same Server work. No cross-Scope content leakage was observed; this report concerns incorrect error semantics and retry behavior.

The owner-pending decision is mapped to `AccessUnavailableError` at [src/powercontext/server/authz/service.py:747](https://github.com/oceanbase/powercontext/blob/fa581eeca400bf3116e1191fcf4add0dd702f5cf/src/powercontext/server/authz/service.py#L747). Distinguish a missing artifact in an authorized Scope from genuinely pending ownership. Cover both this case and the non-disclosure behavior for unauthorized Scopes.

## Environment

Teingi:

Closes #1668.

# Rationale for this change

Reading an exact Memory citation in an accessible but empty Scope returned `503 artifact_owner_pending`, encouraging clients to retry a reference that cannot resolve there.

When an exact Memory read is denied specifically because its owner is pending, verify that the caller can read the parent Scope and inspect the content-free logical catalog. Return `404 memory_not_found` if that Memory entry is absent. Existing entries without committed ownership still fail closed with 503, and unauthorized requests retain the same 403 response. Add regressions for both built-in and Casbin authorization, repeated wrong-Scope reads, missing entry IDs, and the existing-content/owner-pending control.

Missing exact Memory reads in readable Scopes return a stable 404. No API schema or permission grants change.

# How was this change tested?

- Both authorization providers reproduced the original 503 before the fix.
- Access HTTP and regression suites: 38 tests passed, plus the final owner-pending regression passed in a separate focused run (39 relevant tests total).
- Real HTTP with enforced Access on SQLite, OceanBase, and embedded seekdb: correct Scope 200; wrong readable Scope 404 on both attempts; inaccessible populated/empty/nonexistent Scopes all 403 with identical errors; deliberately failed owner persistence left existing content unreadable with 503. Dedicated resources were removed and cleanup verified.
- Full pre-commit hooks pass except the four pre-existing unresolved `scripts.*` imports in `tests/test_integration_guidance_evaluation.py`; `ty check --exclude tests/test_integration_guidance_evaluation.py` passes.

# AI usage statement


happy-v587 (commented):
LGTM

Zxf-xufeng (approved):
LGTM

PR #1681, opened 2026-09-20
feat(config): persist server settings and support explicit port changes

### LINKED ISSUE #1641: feat: configure the Web Dashboard / Server port through powercontext config
## Feature description

Allow users to choose and persist a custom Web Dashboard / Server port through `powercontext config`, including a fresh local setup and an existing configuration that still uses the default port `8000`.

Users may need a different port because `8000` is occupied or their local environment requires another port. They should be able to configure it through the CLI configuration tool without manually editing environment files.

The underlying Server already supports `POWERCONTEXT_SERVER_HTTP_PORT`. However, the current `powercontext config init` wizard only prompts for the local Server port when the stored port is invalid or is already different from `8000`:

```python
if state.scenario == "local" and (invalid_port or port != 8000):
    # Prompt for the Server port here.
    ...
```

Consequently, a fresh local setup skips the port selection, and users cannot change a valid default port through that part of the wizard. Remote custom-listener and reverse-proxy paths already expose a port prompt.

Please make the port editable in the configuration workflow for all relevant scenarios, with the existing value (or `8000` for a fresh setup) as the default. For example, users should be able to run `powercontext config init`, choose local setup, select `18000`, save, and start the Server with the saved configuration to access the Dashboard at `http://127.0.0.1:18000/`.

### Acceptance criteria

- A fresh local setup exposes an editable Web / Server port with `8000` as the default.
- Users can change an existing port from `8000` to a custom value, change between custom values, and restore `8000` through `powercontext config`.
- The chosen port is validated as an integer in the range `1–65535` and persisted as `POWERCONTEXT_SERVER_HTTP_PORT`.
- Starting or restarting the Server with that saved configuration uses the selected port.
- Generated local client URLs and displayed Dashboard / Server URLs reflect the selected port; explicitly configured remote or proxy URLs remain consistent with their access scenario.
- The UI explains that the Dashboard, HTTP API, and MCP share the Server listener and port. A separate Dashboard-only listener is not required.
- The workflow clearly indicates when an already-running Server needs to be restarted for the change to take effect.
- Add focused coverage for fresh setup and changing an existing default port, and document the configuration steps.

## Alternatives considered

- Manually edit `POWERCONTEXT_SERVER_HTTP_PORT` in the environment file. This is a workaround, but it does not meet the need to configure the port through `powercontext config`.
- Choose a remote access scenario just to expose its port prompt. Local users should not need to select a different deployment scenario to change a local port.

## Additional context

Source inspected: [`_network` in `config_wizard.py`](https://github.com/oceanbase/powercontext/blob/cd03cef045eba5ec6bf0d62b11ec1062c2d5862d/src/powercontext/cli/config_wizard.py#L515-L527).

Related issues reviewed:
- #1290 establishes the shared Dashboard / HTTP API / MCP listener.
- #949 concerns changing the legacy default port.
- #1160 requests legacy PowerMem management commands, including port selection.

These related issues are closed and do not specifically track making the default local Web / Server port editable through the current `powercontext config` workflow. Automatic port fallback is outside this request.

PsiACE:

Closes #1641.

# Rationale for this change

Keep personal server configuration independent of the working directory and support changing ports through the setup wizard.

- Default new installations to port `17429`, preserving explicitly configured ports and registered service data directories.
- Store `server.env` in the user configuration directory provided by `platformdirs`, with support for explicit files and existing `.env` files.
- Allow editing the default port in the wizard and update generated local connection URLs while preserving remote and proxy URLs.
- Add `setup <host> --configure-only` to update client connections using existing configuration write and recovery logic.
- Update English and Chinese documentation.

The Dashboard, HTTP API, and MCP share one port. Port conflicts fail startup. After changing the port, restart the server, update and reload client connections, and run `doctor` to check authentication and connectivity.

# How was this change tested?

- `uv run pytest --doctest-modules -q -ra --tb=short`: 2,940 passed, 116 skipped.
- DSH, Pi, OpenCode, and OpenClaw plugin suites: 513 passed.
- `make check`, `make contract-test`, plugin builds, and `make docs-test` passed.
- Verified occupied-port failures, existing configuration preservation, database access across working directories, and client reconnection.

# AI usage statement

Codex (GPT-6) assisted with implementation, testing, and review.


Teingi (approved):
LGTM

